PR #1, opened 2018-10-29
init for front-end

danielborowski:
:)


front-end/gatsby-node.js
```diff
@@ -0,0 +1,115 @@
+const path = require('path');
+const crypto = require('crypto');
+const uuidv1 = require('uuid/v1');
+const contentful = require('contentful');
+const mongoclient = require('mongodb').MongoClient;
+const { Stitch, AnonymousCredential } = require('mongodb-stitch-server-sdk');
+
+let LANGUAGES = [
```
i80and: Stick with `const` unless a binding definitely _must_ be non-constant.

front-end/src/components/Step.js
```diff
@@ -0,0 +1,55 @@
...
+    }
+    else if (sectionData.type === 'directive' && sectionData.name === 'tabs') {
+      return <Tabs tabsData={ sectionData } refDocMapping={ this.props.refDocMapping } key={ index } addLanguages={ this.props.addLanguages } activeLanguage={ this.props.activeLanguage } />
+    }
+    else if (sectionData.type === 'code') {
+      return <CodeBlock codeData={ sectionData } key={ index } />
+    }
+    else if (sectionData.type === 'heading') {
+      return (
+        <h3 key={ index }>{ sectionData.children[0].value }
```
i80and: Keep in mind that step titles may contain arbitrary inline formatting

front-end/src/components/Section.js
```diff
@@ -0,0 +1,97 @@
...
+      showAllSteps: !this.state.showAllSteps,
+      showAllStepsText: this.state.showAllSteps ? 'Expand All Steps' : 'Collapse All Steps'
+    });
+  }
+
+  render() {
+    return (
+      <div className="section" id={ this.props.sectionData.name }>
+        <h2> 
+          { this.nameMapping[this.props.sectionData.name] }
```
i80and: Not applicable for guides, but remember that section titles may have arbitrary inline formatting

front-end/src/components/Step.js
danielborowski: will fix in future

front-end/src/components/Section.js
danielborowski: will fix in future :)

PR #2, opened 2018-10-30
uri writer UI and admonition

danielborowski:


front-end/src/components/Step.js
```diff
@@ -3,13 +3,28 @@ import Paragraph from '../components/Paragraph';
 import CodeBlock from '../components/CodeBlock';
 import LiteralInclude from '../components/LiteralInclude';
 import Tabs from '../components/Tabs';
+import Admonition from '../components/Admonition';
 
 export default class Step extends Component {
 
+  constructor() {
+    super();
+    this.admonitions = [
```
i80and: This should probably be a global or a static member, since it's not going to change between instances.

front-end/src/components/Admonition.js
```diff
@@ -0,0 +1,44 @@
+import React, { Component} from 'react';
+import Paragraph from '../components/Paragraph';
+
+export default class Admonition extends Component {
+
+  // backwards compatible css clasnames
```
i80and: `classnames`

PR #5, opened 2018-11-09
factory to create components from AST nodes

danielborowski:


front-end/src/components/Paragraph.js
```diff
@@ -1,35 +1,24 @@
...
   render() {
     return (
       <p style={ { margin: this.props.admonition ? '0 auto' : '' } }>
-        { this.paragraphRendering() }
+        { 
+          this.props.nodeData.children.map((element, index) => {
+            if (element.type === 'text') {
+              return <span key={ index }>{ element.value }</span>
+            }
+            else if (element.type === 'emphasis') {
```
i80and: Emphasis can occur in non-paragraph contexts; for example, headings

front-end/src/components/Paragraph.js
```diff
@@ -1,35 +1,24 @@
...
+              return <i key={ index }>{ element.children[0].value }</i>
```
i80and: `<i>` hasn't been a standard tag since HTML4; `<em>` is preferred instead

PR #6, opened 2018-11-19
initial code for docker+k8 setup

danielborowski:


front-end/server.js
```diff
@@ -0,0 +1,65 @@
...
+      res.write('</html>');
+      res.end();
+    });
+  });
+};
+
+// https://github.com/gatsbyjs/gatsby/issues/3485
+router.get('/', (req, res) => { 
+  // pre-build config
+  if (!fs.existsSync('docs-tools/')) {
```
i80and: We should probably pull regardless, since otherwise this could end up stale

front-end/server.js
```diff
@@ -0,0 +1,65 @@
...
+  // get params
+  const STITCH_ID = req.query.stitch_id;
+  const NAMESPACE = req.query.namespace;
+  const PREFIX = req.query.prefix;
+  if (!STITCH_ID || !NAMESPACE || !PREFIX) {
+    res.write('<p>Error with query params</p>');
+    return;
+  }
+  // kick off build
+  const gatsbyBuild = exec(`STITCH_ID=${STITCH_ID} NAMESPACE=${NAMESPACE} PREFIX=${PREFIX} gatsby build --prefix-paths`);
```
i80and: This is dangerous *as hell*. We're passing unescaped query parameters into the shell!

Better:
```
exec(
    'gatsby build --prefix-paths',
    {
        env: {
            STITCH_ID,
            NAMESPACE,
            PREFIX
        }
    })
```

I additionally recommend validating the values of each query parameter with a regex.

front-end/server.js
```diff
@@ -0,0 +1,65 @@
...
+  gatsbyBuild.stdout.on('data', (data) => {
+    console.log(data); 
+    res.write('<p style="margin:0;font-size:13px;">' + data + '</p>');
```
i80and: We need to escape `data`.

front-end/package.json
```diff
@@ -10,14 +10,17 @@
...
   "dependencies": {
     "@babel/plugin-syntax-dynamic-import": "^7.0.0",
     "@babel/plugin-transform-react-jsx": "^7.0.0",
     "axios": "^0.18.0",
     "babel-plugin-transform-react-jsx": "^6.24.1",
-    "contentful": "^7.0.3",
+    "express": "^4.16.4",
     "gatsby": "^2.0.2",
-    "mongodb": "^3.1.6",
+    "http": "0.0.0",
```
i80and: What is this?

danielborowski: removed http and mongodb module is not needed anymore because of stitch sdk

PR #9, opened 2018-12-21
Guides site builds without errors: DOCSP-3895

danielborowski:
initial changes for incremental builds


front-end/src/components/Include.js
```diff
@@ -8,7 +8,20 @@ export default class Include extends Component {
...
     if (key.includes('.rst')) key = key.replace('.rst', '');
-    this.resolvedIncludeData = this.props.refDocMapping[key].ast.children;
+    this.resolvedIncludeData = [];
+    // TODO: use param from parent comp instead
+    // server vs. client side data fetching
+    if (Object.keys(this.props.refDocMapping).length > 0 ) {
+      this.props.refDocMapping[key].ast.children
+    } else {
+      this.props.stitchClient.callFunction('fetchDocuments', ['snooty/documents', { _id: 'guides/andrew/master' + '/' + key }]).then((response) => {
+        console.log('data for include', response);
```
skerschb: _id prefix will need to be dynamic based on URL input or posted parameters.

front-end/dist/index.html
```diff
@@ -0,0 +1,29 @@
...
+    <meta property="og:image" content="http://s3.amazonaws.com/info-mongodb-com/_com_assets/cms/mongodb-for-giant-ideas-bbab5c3cf8.png"/>
+    <meta property="og:image:secure_url" content="https://webassets.mongodb.com/_com_assets/cms/mongodb-for-giant-ideas-bbab5c3cf8.png"/>
+    <link rel="stylesheet" href="/static/guides.css" type="text/css" />
+    <link rel="stylesheet" href="/static/pygments.css" type="text/css" />
+    <link rel="stylesheet" href="/static/css/navbar.min.css" type="text/css" />
+    <link rel="search" type="application/opensearchdescription+xml" href="https://docs.mongodb.com/osd.xml" title="MongoDB Help"/>
+  </head>
+  <body data-project="guides" data-project-title="MongoDB Guides" data-branch="DOCSP-3279" data-enable-marian="1">
+    <div id="navbar" data-navprops='{"links": [{"url": "https://docs.mongodb.com/manual/","text": "Server"},{"url": "https://docs.mongodb.com/ecosystem/drivers/","text": "Drivers"},{"url": "https://docs.mongodb.com/cloud/","text": "Cloud"},{"url": "https://docs.mongodb.com/tools/","text": "Tools"},{"url": "https://docs.mongodb.com/guides/","text": "Guides","active": true}]}'></div>
```
skerschb: hardcoded branch

front-end/src/templates/guide.js
```diff
@@ -47,11 +81,27 @@ export default class Guide extends Component {
...
+      }
+    });
+  }
+
   setupStitch() {
-    const appName = this.props.pageContext.__stitchID;
+    const appName = this.stitchId;
+    if (!appName) return;
     this.stitchClient = Stitch.hasAppClient(appName) ? Stitch.defaultAppClient : Stitch.initializeDefaultAppClient(appName);
     this.stitchClient.auth.loginWithCredential(new AnonymousCredential()).then((user) => {
```
skerschb: this.stitchClient = Stitch.hasAppClient(appName)
        ? Stitch.getAppClient(appName)
        : Stitch.initializeAppClient(appName);

//we can't assume that the default is the appName

front-end/dist/index.html
```diff
@@ -0,0 +1,29 @@
+<html>
```
skerschb: Typically we name branches after tickets, so once this is merged in any future changes should have branch names that look like DOCSP-xxxx and then something more descriptive in the commit message.

front-end/src/templates/guide.js
```diff
@@ -1,16 +1,51 @@
 import React, { Component } from 'react';
+import ReactDOM from 'react-dom';
 import TOC from '../components/TOC';
 import GuideSection from '../components/GuideSection';
 import GuideHeading from '../components/GuideHeading';
 import Modal from '../components/Modal';
 import { Stitch, AnonymousCredential } from 'mongodb-stitch-browser-sdk';
```
skerschb: Can we abstract this stitch connectivity in order to avoid binding to the StitchClient for all of the content templates?

front-end/src/components/Include.js
```diff
@@ -8,9 +8,25 @@ export default class Include extends Component {
     let key = this.props.nodeData.argument[0].value;
     if (key.startsWith('/')) key = key.substr(1);
     if (key.includes('.rst')) key = key.replace('.rst', '');
```
i80and: Is there a reason this is includes rather than endsWith?

front-end/src/components/Role.js
```diff
@@ -31,6 +31,22 @@ export default class Role extends Component {
         </a>
       )
     }
+    else if (this.props.nodeData.name === 'binary') {
```
i80and: This should probably be a set membership test rather than a comparison, since I think this will apply to other roles than just binary

front-end/src/templates/guide.js
```diff
@@ -125,9 +129,9 @@ export default class Guide extends Component {
         };
       } else {
         const parsed = this.DOMParser.parseFromString(response, 'text/html');
-        const mainContainer = parsed.getElementById(findHashPart).nextElementSibling;
+        const mainContainer = parsed.getElementById(findHashPart) ? parsed.getElementById(findHashPart).nextElementSibling : false;
```
i80and: `null`, not `false`. Otherwise this type signature gets weird

front-end/src/templates/guide.js
```diff
@@ -125,9 +129,9 @@ export default class Guide extends Component {
...
         contentObj = {
-          text: mainContainer.getElementsByTagName('p')[0].textContent.trim(),
+          text: mainContainer.getElementsByTagName('p')[0] ? mainContainer.getElementsByTagName('p')[0].textContent.trim() : 'FIX: no content found',
```
i80and: We should log this rather than injecting content. Makes it easier to find issues.

front-end/src/components/ComponentFactory.js
```diff
@@ -43,9 +45,13 @@ export default class ComponentFactory extends Component {
     let ComponentType = this.componentMap[lookup];
     // the different admonition types are all under the Admonition component
     // see 'this.admonitions' in 'guide.js' for the list
-    if (!ComponentType && this.props.admonitions.includes(name)) {
+    if (!ComponentType && this.props.admonitions && this.props.admonitions.includes(name)) {
       ComponentType = this.componentMap['admonition'];
     }
+    // component with this type not implemented
+    if (!ComponentType) {
+      return <span>==Not implemented: { type }, { name } ==</span>
```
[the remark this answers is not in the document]
danielborowski: I think eventually yes but for now I think this works better because it explicitly shows what components are missing that I need to create for the Guides site. In the future though, logging it would be better practice as to not break the page etc.

front-end/src/components/Include.js
```diff
@@ -8,9 +8,25 @@ export default class Include extends Component {
...
-    this.resolvedIncludeData = this.props.refDocMapping[key].ast.children;
-    this.props.updateTotalStepCount(this.resolvedIncludeData.length);
-    console.log(99, this.props.refDocMapping[key]);
+    this.resolvedIncludeData = [];
+    // TODO: use param from parent comp instead
+    // server vs. client side data fetching
+    if (this.props.refDocMapping && Object.keys(this.props.refDocMapping).length > 0 ) {
```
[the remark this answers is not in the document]
danielborowski: `refDocMapping` is originally passed in from gatsby-node.js which is created in a Stitch function and it doesn't support Maps yet :( So I *could* create an object in stitch and then convert it to a map before sending it to all components, but I figured I'd leave it as an object.

front-end/src/components/LiteralInclude.js
```diff
@@ -5,13 +5,19 @@ export default class LiteralInclude extends Component {
 
   constructor(props) {
     super(props);
-    let key = this.props.nodeData.argument[0].children[0].value;
+    let key = this.props.nodeData.argument[0].value;
+    // fix for some includes
+    if (key && key[0] === '/') {
+      key = key.substr(1);
```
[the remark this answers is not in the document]
danielborowski: I remove the first slash here because all the `include` files are stored in the `refDocMapping` object as `include/a/b/c` whether or not they began with a slash or not. So to find the correct one I disregard whether its relative to root or current document.

front-end/src/components/Include.js
danielborowski: good point

front-end/src/templates/guide.js
danielborowski: fixed, but I'll need to eventually rewrite this whole function to be more robust and account for all different role types.

front-end/src/templates/guide.js
danielborowski: Similar to previous comment, once I implement all the roles I'll make this function more robust to account for errors and no-content.

front-end/src/components/Include.js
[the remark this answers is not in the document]
i80and: Bah humbug :+1:

front-end/src/components/LiteralInclude.js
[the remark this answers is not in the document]
danielborowski: TODO: fix in new branch to account for relative/absolute paths

PR #11, opened 2019-02-01
DOCSP-4173: init commit for jest tests

danielborowski:


front-end/babel.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  presets: [
+    [
+      '@babel/preset-env',
+      {
+        targets: {
+          node: 'current',
```
i80and: Will anything be running in the client?

front-end/babel.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
```
i80and: Should the .babelrc be merged into this?

PR #13, opened 2019-02-08
Bugfix: properly fetch guides for landing page

sophstad:
A restructuring of the AST caused guides to not be properly parsed for the landing page. This PR properly renders the homepage, and finds guides by searching through `children` arrays of the AST rather than querying a specific index.


danielborowski:
LGTM I tested it and it works on develop and build

PR #14, opened 2019-02-11
DOCSP-4188: URI Writer

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-4188)] This PR adds two components associated with the URI Writer:
- `URIWriter.js`: a form that allows user to input their own data, from which we generate connection strings
- `URIText.js`: replaces placeholders with the URI string. For now, we are only examining text that appears as a `Code` component, but it would be easy to replace text in any component with a URI string.


front-end/tests/URIWriter.test.js
```diff
@@ -0,0 +1,267 @@
+import React from 'react';
+import { mount } from 'enzyme';
```
skerschb: We were thinking about using React testing library for the react component tests but I have to admit I like the way these tests look...the mocking is elegant.

sophstad: Ah sorry! I should have checked in first 😣 I am pretty familiar with Enzyme so I went with that—but am happy to update these if we want to go with the other library.

front-end/src/components/GuideSection.js
```diff
@@ -62,9 +73,12 @@ export default class GuideSection extends Component {
             return <ComponentFactory { ...this.props } 
                                       nodeData={ child } 
                                       key={ index } 
+                                      handleUpdateURIWriter={this.handleUpdateURIWriter.bind(this)}
```
i80and: Nit: Spaces around interior of curly braces

i80and: Optionally, use public class fields to avoid the need for `.bind()`

front-end/src/components/URIText.js
```diff
@@ -0,0 +1,123 @@
+import React from 'react';
+
+const TEMPLATE_TYPE_SELF_MANAGED = 'local MongoDB';
+const TEMPLATE_TYPE_REPLICA_SET = 'local MongoDB with replica set';
+const TEMPLATE_TYPE_ATLAS_36 = 'Atlas (Cloud) v. 3.6';
+const TEMPLATE_TYPE_ATLAS_34 = 'Atlas (Cloud) v. 3.4';
+const TEMPLATE_TYPE_ATLAS = 'Atlas (Cloud)';
+
+const URI_PLACEHOLDER = '<URISTRING>';
```
i80and: Hm. We should probably single-source these constants between the two components

front-end/src/components/URIText.js
```diff
@@ -0,0 +1,123 @@
...
+      value: 'true'
+    }
+  ],
+}
+
+const URIText = ({ value, templateType, uri }) => {
+  return replacePlaceholderWithURI(value, templateType, uri);
+}
+
+const replacePlaceholderWithURI = (value, templateType, uri) => {
```
i80and: I'm inclined to think that these should be regular function declarations

front-end/tests/URIText.test.js
```diff
@@ -0,0 +1,204 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import renderer from 'react-test-renderer';
+import URIText from '../src/components/URIText';
+
+describe('local MongoDB', () => {
+  const templateType = 'local MongoDB'
```
i80and: Missing semicolon

front-end/tests/URIText.test.js
```diff
@@ -0,0 +1,204 @@
...
+      host2: 'cluster0-shard-00-02-igkvv.mongodb.net:27017',
+    },
+    replicaSet: '',
+    ssl: '',
+    username: 'myUsername',
+  };
+
+  describe('when called without a placeholder', () => {
+    it('returns the original string', () => {
+      const value = `This is the body text`;
```
i80and: Use single-quotes unless you're actually formatting text

sophstad:
@i80and thanks! I've addressed your comments, ready for another look.

front-end/src/components/GuideSection.js
danielborowski: I like spaces around interior of curly braces as well

front-end/tests/URIWriter.test.js
skerschb: No this is great, we can keep it.

PR #15, opened 2019-02-13
DOCSP-4317: Add linter to next-gen React builds

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-4317)] This PR:
- Implements [Prettier](http://prettier.io) and [ESLint](http://eslint.org) to standardize code formatting
- Adds a precommit hook to format code using [Husky](https://www.npmjs.com/package/husky)
- Runs lint tests on Drone
- Properly formats all existing code
- Implements type checking using [prop-types](https://www.npmjs.com/package/prop-types). Existing components have been updated, and future components will require type validation
- Uses the same formatting rules as [Atlas](https://github.com/10gen/mms) and [Stitch](https://github.com/10gen/stitch-ui) to increase standardization

The changes in this PR were largely automated by Prettier. Some linter errors and prop type errors were fixed manually, and other minor improvements were made to components when convenient.


.drone.yml
```diff
@@ -13,6 +13,15 @@ pipeline:
       - cd front-end/
       - npm test
 
+  lint:
+    image: node:10.10.0
```
i80and: Is there a reason this is `node:10.10.0` and not just `node:10`? This is already out of date versus the current `10.15.1`

front-end/src/html.js
```diff
@@ -1,52 +1,72 @@
...
 HTML.propTypes = {
-  htmlAttributes: PropTypes.object,
-  headComponents: PropTypes.array,
-  bodyAttributes: PropTypes.object,
-  preBodyComponents: PropTypes.array,
-  body: PropTypes.string,
-  postBodyComponents: PropTypes.array,
-}
+  /* eslint-disable react/forbid-prop-types */
+  htmlAttributes: PropTypes.object.isRequired,
```
i80and: Is there any way we can be more specific types here? I suspect not, but I might as well ask

front-end/src/templates/guide.js
```diff
@@ -38,11 +31,13 @@ export default class Guide extends Component {
...
-        example: null
-      }
-    }; 
-    console.log(2222, this.props.pageContext.__refDocMapping);
-    console.log(4544, this.sections);
+        example: null,
+      },
+    };
+
+    this.addLanguages = this.addLanguages.bind(this);
```
i80and: Any chance we can use public class fields to avoid the need for these `.bind()`'s?

.drone.yml
sophstad: `node:10.10.0` is the exact version [referenced in the Dockerfile](https://github.com/mongodb/snooty/blob/master/Dockerfile#L5), so I opted to use the same. But I'd definitely feel comfortable changing both of them to `node:10`.

front-end/src/html.js
sophstad: I did some research into how Gatsby populates these fields as they are Gatsby-specific, but (shocker) couldn't find much documented. I will have another look but this may be the best we can do.

.drone.yml
i80and: I'd prefer that. Doesn't need to be part of this PR though, since it's kind of unrelated

front-end/src/html.js
i80and: :+1:

front-end/src/templates/guide.js
sophstad: I wasn't familiar with public class fields! Just pushed a change implementing them.

sophstad:
@i80and RFAL! 😸

i80and:
@sophstad I'll merge if you resolve the Admonition conflict!

PR #16, opened 2019-02-20
fix master from breaking on P elements

danielborowski:


i80and:
I like the revised version a lot more ;)

PR #18, opened 2019-02-22
Bugfix: Allow enzyme shallow rendered components to match snapshots

sophstad:
This PR introduces a plugin that automatically converts Enzyme shallow mounted components to JSON, allowing them to be compared to snapshot files using Jest's `.toMatchSnapshot()`.


danielborowski:
lGtm

PR #20, opened 2019-02-26
DOCSP 4189: Tabbed content

danielborowski:


front-end/src/templates/guide.js
```diff
@@ -52,23 +55,42 @@ export default class Guide extends Component {
...
-      activeLanguage: setLanguages[0],
+      [opts.type]: setTabs,
+      [opts.active]: setTabs[0],
     });
   };
 
+  // this function gets an array of objects that compose some tabset
+  // currently only supports: language pills, OS tabs
+  // TODO: cloud/local
+  addTabset = languages => {
```
sophstad: Because we now have the "name" of a tabset (e.g. `cloud`, `drivers`), this function should be reworked. The `Tabs` component (which calls this function) passes in `nodeData.children`, but `nodeData.options.tabset` specifies the tabset type. We no longer need to check any of the object values. As a bonus, it should be pretty easy to make the local/cloud tabs work right away.

front-end/src/templates/guide.js
```diff
@@ -52,23 +55,42 @@ export default class Guide extends Component {
...
+    return setTabs;
+  };
+
   changeActiveLanguage = language => {
     this.setState({
       activeLanguage: language,
     });
   };
 
+  changeActiveOSTab = language => {
```
sophstad: You should be able to generalize this handler to avoid creating separate functions for specific tab types. This would require passing the tabset type in to the function. For example, something like:

```js
setActiveTab = (tabset, value) => {
  this.setState({ [tabset]: value });
}
```

This will also simplify the props we have to pass to child components. I did something similar for `URIWriter` and used [this](https://reactjs.org/docs/forms.html#handling-multiple-inputs) resource.

front-end/gatsby-node.js
```diff
@@ -21,6 +21,12 @@ const LANGUAGES = [
   ['scala', 'Scala']
 ];
 
+const OSTABS = [
```
sophstad: What was your reasoning behind including `OSTABS` in this Gatsby file? As far as I can tell, we could define the constants in `GuideSection.js` and simply pass the active tab to its parent component. I think all of the other child components only need access to the _active_ OS, not the entire set of options. Ditto for languages. But maybe there is a good reason to define them here!

front-end/src/components/GuideSection.js
```diff
@@ -56,9 +56,7 @@ export default class GuideSection extends Component {
   };
 
   render() {
-    const {
-      guideSectionData: { children, name },
-    } = this.props;
+    const { guideSectionData, OSTabs, activeOSTab, changeActiveOSTab } = this.props;
```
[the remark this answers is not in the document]
danielborowski: you're right it works now idk what was going on earlier

front-end/src/components/GuideSection.js
```diff
@@ -86,7 +84,29 @@ export default class GuideSection extends Component {
...
+                    changeActiveOSTab(langOpts);
+                  }}
+                >
+                  {langOpts[1]}
+                </li>
+              );
+            })}
+          </ul>
+        ) : (
+          ''
```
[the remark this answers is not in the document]
danielborowski: ah yes clever

front-end/src/constants.js
```diff
@@ -0,0 +1,22 @@
+const __languages = [
```
sophstad: JavaScript best practice is to use uppercase for constants—could we make this `LANGUAGES`?

front-end/src/components/GuideHeading.js
```diff
@@ -19,26 +19,43 @@ export default class GuideHeading extends Component {
...
+              Deployment Type:
+              <span className="show-current-deployment"> {activeDeployment[1]}</span>
+            </div>
+            <ul className="guide__pills pillstrip-declaration" role="tablist" data-tab-preference="cloud">
+              {deployments.map((langOpts, index) => (
+                <li
+                  className={
+                    activeDeployment[0] === langOpts[0]
+                      ? 'guide__pill guide__deploymentpill guide__deploymentpill--active'
+                      : 'guide__pill guide__deploymentpill'
```
sophstad: This could be simplified, but not necessary. Your call!
```js
className={`guide__pill guide__deploymentpill ${activeDeployment[0] === langOpts[0] && 'guide__deploymentpill--active'}`}
```

front-end/src/constants.js
```diff
@@ -0,0 +1,22 @@
...
+  ['java-async', 'Java (Async)'],
+  ['c', 'C'],
+  ['cpp', 'C++11'],
+  ['csharp', 'C#'],
+  ['perl', 'Perl'],
+  ['ruby', 'Ruby'],
+  ['scala', 'Scala'],
+];
+
+const __ostabs = [['windows', 'Windows'], ['macOS', 'macOS'], ['linux', 'Linux']];
```
sophstad: Same as above

front-end/src/constants.js
```diff
@@ -0,0 +1,22 @@
...
+
+const __deployments = [['cloud', 'Cloud'], ['local', 'Local']];
```
sophstad: Same as above

PR #25, opened 2019-03-04
Bugfix: templateType, pills

sophstad:
Removes all warnings when running `npm test` (mostly replacing `templateType` with `activeDeployment`) and makes general improvements to pill implementation. This includes updating the constants file to be more readable.


danielborowski:
LGTM thumbs up emoji

PR #26, opened 2019-03-04
DOCSP-4611: Implement roles/fix admonitions

danielborowski:


front-end/src/components/Role.js
```diff
@@ -10,42 +10,67 @@ export default class Role extends Component {
...
+      const label = nodeData.label && nodeData.label.value ? nodeData.label.value : nodeData.label;
       return (
-        <a
-          href={href}
-          onMouseEnter={e => {
-            modal(e, href);
-          }}
-        >
-          {nodeData.label}
+        <a href={`https://docs.mongodb.com/compass/current/#${nodeData.target}`} className="reference external">
```
sophstad: Do nodes named `ref` always linked to Compass? Sorry just not familiar with them! It seems like they are often not appearing on the pages, apparently due to issues with `List` and nested `Include` components :/

front-end/src/components/Role.js
```diff
@@ -10,42 +10,67 @@ export default class Role extends Component {
...
+          <span className="xref std std-ref">{label}</span>
         </a>
       );
     }
-    // binary case is unique (maybe others will be as well)
+    // basic roles with interaction
+    if (this.roleDataTypes[nodeData.name]) {
```
sophstad: Looks like this role type isn't working exactly right. When running the `server/read_operators` page locally, the `$and` ref doesn't appear the same as the [source](https://docs.mongodb.com/guides/server/read_operators/#write-an-implied-and-query), and the link is dead as well.

PR #27, opened 2019-03-08
Improve testing

sophstad:
This PR:
- Adds all unit tests to `tests/unit/` directory
- Adds a new command to run only unit tests
- Allows users to run tests by keyword/filename
- Updates the README with instructions

In the future, new directories can be added within `tests/` for different types of tests. Commands comparable to `test-unit` can be added to `package.json` to allow running tests in certain directories.


danielborowski:
mostly file path changes and some config, so LGTM merge away! but I think there's some merge conflicts :(

PR #28, opened 2019-03-20
DOCSP-4798: role files

danielborowski:


front-end/src/components/ComponentFactory.js
```diff
@@ -52,11 +72,21 @@ export default class ComponentFactory extends Component {
...
-    if (type === 'target' || type === 'class' || type === 'cssclass' || name === 'cssclass') {
+    if (type === 'target' || type === 'class' || type === 'cssclass' || name === 'cssclass' || name === 'class') {
       return null;
     }
     const lookup = type === 'directive' ? name : type;
     let ComponentType = this.componentMap[lookup];
+    // roles are each in separate file
+    if (type === 'role') {
+      // remove namespace
+      let modName = name;
```
sophstad: If you want a one-liner 😛:
```js
const roleName = name.includes(':') ? name.split(':')[1] : name;
```

front-end/src/components/Roles/Ref.js
```diff
@@ -0,0 +1,40 @@
...
+  const {
+    nodeData,
+    refDocMapping: { REF_TARGETS },
+  } = props;
+  const label = nodeData.label && nodeData.label.value ? nodeData.label.value : nodeData.label;
+  // make sure target is hardcoded in list for now
+  if (!REF_TARGETS[nodeData.target]) {
+    return (
+      <span>
+        ==Role TARGET does not exist:
```
sophstad: Do we still want this here? It may be better to console log if there is an error instead of rendering this.

front-end/src/components/Roles/Ref.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const RoleRef = props => {
+  const {
+    nodeData,
+    refDocMapping: { REF_TARGETS },
+  } = props;
+  const label = nodeData.label && nodeData.label.value ? nodeData.label.value : nodeData.label;
+  // make sure target is hardcoded in list for now
```
sophstad: Should there be a `TODO` here for a future way to update this and avoid hardcoding?

front-end/tests/Role.test.js
```diff
@@ -1,6 +1,9 @@
 import React from 'react';
 import { shallow } from 'enzyme';
-import Role from '../src/components/Role';
+
+// the different roles
+import RoleLink from '../src/components/Roles/Link';
+import RoleRef from '../src/components/Roles/Ref';
```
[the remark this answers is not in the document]
danielborowski: yes :(

front-end/src/components/Roles/GUILabel.js
```diff
@@ -0,0 +1,17 @@
...
+const RoleGUILabel = props => {
+  const {
+    nodeData: { label },
+  } = props;
+  return <span className="guilabel">{label}</span>;
+};
+
+RoleGUILabel.propTypes = {
+  nodeData: PropTypes.shape({
+    label: PropTypes.oneOfType([PropTypes.string, PropTypes.object]).isRequired,
```
sophstad: Since `label` is rendered directly in a `<span>`, how could it be an object? I think that would throw an error.

front-end/src/components/Roles/Program.js
```diff
@@ -0,0 +1,17 @@
...
+const RoleProgram = props => {
+  const {
+    nodeData: { label },
+  } = props;
+  return <strong className="program">{label}</strong>;
+};
+
+RoleProgram.propTypes = {
+  nodeData: PropTypes.shape({
+    label: PropTypes.oneOfType([PropTypes.string, PropTypes.object]).isRequired,
```
sophstad: Same comment as in `GUILabel.js`

front-end/src/components/Roles/Ref.js
```diff
@@ -0,0 +1,33 @@
...
+  // TODO: chat w andrew about how to move forward with sphinx inv files
```
sophstad: Should this be `env`?

front-end/src/components/Roles/Ref.js
```diff
@@ -0,0 +1,33 @@
...
+  if (!REF_TARGETS[nodeData.target]) {
+    console.log(`ERROR: ROLE TARGET DOES NOT EXIST => ${nodeData.target}`);
```
sophstad: You can use [`console.error()`](https://developer.mozilla.org/en-US/docs/Web/API/Console/error) to make this have error highlighting in the console 😄

front-end/src/components/Roles/Ref.js
danielborowski: nah it's some weird inventory files generated by Sphinx (Andrew can explain in more detail)

front-end/src/components/Roles/Ref.js
danielborowski: wow awesome

PR #29, opened 2019-03-22
DOCSP-4921: Refactor tabs

sophstad:
This PR refactors tabs.
- Adds support for anonymous tabs of any kind (not just OS)
- Names anonymous tabsets by looping through tab names and joining with a slash `/`, e.g. `windows/macos/linux` (see `Tabs.generateAnonymousTabsetName`)
- Updates URI Writer components, which were affected by the new way we handle `activeDeployment`
- Will make it easier to support local storage of tab states


front-end/src/components/GuideHeading.js
```diff
@@ -23,27 +24,27 @@ export default class GuideHeading extends Component {
           <div className="guide-prefs">
             <div className="guide-prefs__caption">
               Deployment Type:
-              <span className="show-current-deployment"> {activeDeployment}</span>
+              <span className="show-current-deployment"> {stringifyTab(activeTabs.deployments)}</span>
             </div>
-            <ul className="guide__pills pillstrip-declaration" role="tablist" data-tab-preference="cloud">
+            <ul className="guide__pills pillstrip-declaration" role="tablist" data-tab-preference="deployments">
```
skerschb: wondering if it's possible to genericize these style names as the pills are used for the manual as well. Perhaps a question for @i80and

front-end/src/templates/guide.js
```diff
@@ -176,7 +167,7 @@ export default class Guide extends Component {
 
   createSections() {
     const { pageContext } = this.props;
-    const { activeDeployment, activeLanguage, activeOSTab, OSTabs, languages, deployments } = this.state;
+    const { activeTabs } = this.state;
```
skerschb: wow that's a lot cleaner, nice!

front-end/src/constants.js
```diff
@@ -1,87 +1,47 @@
...
-  {
-    name: 'cloud',
-    value: 'Cloud',
-  },
-  {
-    name: 'local',
-    value: 'Local',
-  },
-];
+const SLUG_TO_STRING = {
```
skerschb: I like this -- nice abstraction.

PR #30, opened 2019-03-26
DOCSP-4689: Local Storage

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-4689)] This PR adds local storage support for tabs and the URI writer. Values are saved using helper functions in `localStorage.js` and used to fill URI Writer `input` fields or select the proper tab when the page is next loaded.

Some variables have also been renamed so that they are more clear in the context of local storage and anonymous tablets.


front-end/src/components/Tabs.js
```diff
@@ -13,8 +13,19 @@ export default class Tabs extends Component {
...
   }
 
+  /*
+   * For anonymous tabsets, create a tabset name that alphabetizes the tab names, formats them in lowercase,
+   * and joins them with a forward slash (/)
+   */
   generateAnonymousTabsetName = nodeData => {
-    return nodeData.children.flatMap(child => child.argument[0].value).join('/');
+    return nodeData.children
+      .flatMap(child => child.argument[0].value.toLowerCase())
```
i80and: flatMap isn't supported on Edge

PR #31, opened 2019-03-29
DOCSP-5018: Reorganize Directory

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-5018)] This PR:
- Removes source code from `front-end/` into the root directory and fixes tests accordingly
- Removes the `trigger`step from our `.drone.yml` file so that all Drone tests should be passing consistently
- Modifies `package.json` to remove warnings (installs Husky as a dependency, removes vulnerabilities)


danielborowski:
LGTM

PR #32, opened 2019-04-02
DOCSP-4175: Unit tests for interaction

danielborowski:


tests/unit/Tabs.test.js
```diff
@@ -0,0 +1,40 @@
...
+
+    beforeAll(() => {
+      wrapper = mountTabs({ mockCallback, activeTabs: { platforms: PLATFORMS[0] } });
+    });
+
+    it('tabs container exists', () => {
+      expect(wrapper.find('.tab-strip__element').exists()).toEqual(true);
+    });
+
+    it('active tab is set', () => {
```
sophstad: I think it could be good to include a test that the active tab is what we think it is (`PLATFORMS[0]`). You can do this by checking the `activeTabs` prop using [`.props()`](https://airbnb.io/enzyme/docs/api/ReactWrapper/props.html) in Enzyme.

tests/unit/Tabs.test.js
```diff
@@ -0,0 +1,40 @@
...
+// data for this component
+import mockData from './data/Tabs.test.json';
+
+const mountTabs = ({ activeTabs, mockCallback }) =>
+  mount(<Tabs nodeData={mockData} addTabset={mockCallback} setActiveTab={mockCallback} activeTabs={activeTabs} />);
+
+describe('Tabs testing', () => {
+  describe('Tabs selection', () => {
+    let wrapper;
+    const mockCallback = jest.fn();
```
sophstad: I think this test will be more clear by renaming this to something like `mockSetActiveTab`.

tests/unit/Tabs.test.js
```diff
@@ -0,0 +1,40 @@
...
+
+    beforeAll(() => {
+      wrapper = mountTabs({ mockCallback, activeTabs: { platforms: PLATFORMS[0] } });
+    });
+
+    it('tabs container exists', () => {
+      expect(wrapper.find('.tab-strip__element').exists()).toEqual(true);
+    });
+
```
sophstad: A test checking that the tab container has the correct number of tabs may be good.

tests/unit/Tabs.test.js
```diff
@@ -6,35 +6,41 @@ import Tabs from '../../src/components/Tabs';
 // data for this component
 import mockData from './data/Tabs.test.json';
 
-const mountTabs = ({ activeTabs, mockCallback }) =>
-  mount(<Tabs nodeData={mockData} addTabset={mockCallback} setActiveTab={mockCallback} activeTabs={activeTabs} />);
+const mountTabs = ({ mockSetActiveTab, activeTabs }) =>
+  mount(<Tabs nodeData={mockData} addTabset={mockSetActiveTab} setActiveTab={mockSetActiveTab} activeTabs={activeTabs} />);
```
sophstad: Whoops, didn't catch this before—probably a good idea to set up two different mock functions here so we can predict the behavior more easily. Since this isn't a pill set, we _woudn't_ expect `this.props.addTabset` to get called, and we want to make sure `mockSetActiveTab` is only getting called in the context of clicking on a tab. This also may mean that you don't have to call `.mockClear()` later. Let me know if that doesn't make sense!

sophstad: Just thought of this—you can also check that `mockAddTabset` is getting called once when the component is constructed.

tests/unit/__snapshots__/Admonition.test.js.snap
```diff
@@ -1,6 +1,40 @@
...
+exports[`admonition one-off (name=admonition) renders correctly 1`] = `
+<div
+  className="admonition admonition-mlab-is-now-a-part-of-mongodb,-inc. note"
+>
+  <p
+    className="first admonition-title"
+  >
+    mLab is now a part of MongoDB, Inc.
+  </p>
+  <ComponentFactory
```
sophstad: This snapshot test seems weird. We want to check that the admonition gets fully rendered, not that it outputs a ComponentFactory, right?

danielborowski: Resolved in person not to deeply render components because all tests would fail if one thing changes. Will keep as is.

PR #33, opened 2019-04-04
DOCSP-5086: Fix Build

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-5086)] This PR fixes Gatsby builds by doing the following:
- Removes modals, which were the only components relying on the `mongodb-stitch-browser-sdk` package (the cause of failure)
- Updates localStorage functions to first check if the runtime environment is a browser

### DIFF
--- a/src/templates/guide.js
+++ b/src/templates/guide.js
@@ -20,16 +20,11 @@ export default class Guide extends Component {
       pageContext.__refDocMapping[
         this.props['*'] // eslint-disable-line react/destructuring-assignment
       ].ast.children[0].children;
-    this.DOMParser = undefined;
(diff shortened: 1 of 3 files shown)


src/templates/guide.js
```diff
@@ -22,37 +20,16 @@ export default class Guide extends Component {
...
-      modalVisible: false,
-      modalContent: {
-        text: null,
-        example: null,
-      },
     };
   }
 
   componentDidMount() {
     this.DOMParser = new DOMParser();
```
danielborowski: You can also remove domparser it was only used for the modal.

PR #34, opened 2019-04-04
Fix: improve conditional render of tabsets

sophstad:
Small improvement so that we don't show an empty `<ul>` in the place where a pillset would appear.


danielborowski:
LGTM

PR #35, opened 2019-04-11
DOCSP-5133: Sort named tabsets

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-5133)] This PR correctly sorts named tab/pillsets to match the order defined in `src/constants.js` (this order is based roughly on popularity).

There is a common case where an _anonymous_ tabset is used to create tabs for Windows/macOS/Linux rather than a `platforms` tabset. We will now sort these tabs to match the order defined by the `platforms` tabset.

This PR also adds Go as a language option.

### DIFF
--- a/src/components/Tabs.js
+++ b/src/components/Tabs.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import ComponentFactory from './ComponentFactory';
-import { stringifyTab } from '../constants';
+import { PLATFORMS, stringifyTab } from '../constants';
 
 export default class Tabs extends Component {
   constructor(props) {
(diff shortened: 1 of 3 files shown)


src/components/Tabs.js
```diff
@@ -28,15 +28,28 @@ export default class Tabs extends Component {
...
+        referenceArray.indexOf(b.argument[0].value.toLowerCase())
+    );
+  };
+
   render() {
     const { tabsetName } = this.state;
     const { nodeData, activeTabs, setActiveTab } = this.props;
     const isHeaderTabset = tabsetName === 'drivers' || tabsetName === 'cloud';
+    const tabs =
+      tabsetName === 'platforms' || tabsetName === 'linux/macos/windows'
```
danielborowski: Might need to account for tabsetName = "macos/windows" or any combination of platforms in any order?

danielborowski:
LGTM

PR #36, opened 2019-04-12
DOCSP-5132: Add lightbox and tests

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-5132)] This PR adds a Lightbox to images that either:
- Specify `:lightbox:` as one of their options
- Are 10% larger than their container ([as is done currently](https://github.com/mongodb/docs-tools/blob/f6f48fb3876f4ed7b5131d00119755a25a09de7d/themes/mongodb/src/js/componentLightbox.js#L57))

Screen Shot 2019-04-12 at 4 36 16 PM
image

### DIFF
--- a/src/components/Figure.js
+++ b/src/components/Figure.js
@@ -1,13 +1,42 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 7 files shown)


danielborowski:
tested and works. LGTM

PR #38, opened 2019-04-16
DOCSP-4967: Fix Gatsby Build Relative Paths

danielborowski:


gatsby-config.js
```diff
@@ -1,5 +1,11 @@
-// https://www.gatsbyjs.org/docs/gatsby-config/
+const runningEnv = process.env.NODE_ENV ? process.env.NODE_ENV : 'production';
```
sophstad: I think we want the default to be `development` (that's what the [docs show](https://www.gatsbyjs.org/docs/environment-variables/#example-1)—correct me if I'm wrong!). You can simplify the syntax as well:
```js
const runningEnv = process.env.NODE_ENV || 'development';
```

gatsby-config.js
```diff
@@ -1,5 +1,11 @@
...
+
+require('dotenv').config({
+  path: `.env.${runningEnv}`,
+});
+
+console.log(`=== NODE_ENV IS SET TO ${runningEnv} ===`);
```
sophstad: Remove log

gatsby-node.js
```diff
@@ -41,6 +40,31 @@ const setupStitch = () => {
...
+const validateEnvVariables = () => {
+  // make sure necessary env vars exist
+  if (!process.env.NAMESPACE || !process.env.STITCH_ID || !process.env.DOCUMENTS || process.env.GATSBY_PREFIX === undefined) {
+    return { 
+      error: true, 
+      message: 'ERROR with .env.* file: parameters required are GATSBY_PREFIX, DOCUMENTS, NAMESPACE, and STITCH_ID' 
+    };
+  }
+  // make sure formats are correct
+  if (process.env.NODE_ENV === 'production' && !process.env.GATSBY_PREFIX.startsWith('/')) {
```
sophstad: Should we include a similar check for the `DOCUMENTS` field? I just tried running `develop` without a preceding slash and we didn't throw an error, but the site doesn't work.

Nice use of `.startsWith()` btw!

gatsby-node.js
```diff
@@ -41,6 +40,31 @@ const setupStitch = () => {
...
+    return { 
+      error: true, 
+      message: 'ERROR with .env.* file: GATSBY_PREFIX must be in format /<site>/<user>/<branch>' 
+    };
+  }
+  // create split prefix for use in stitch function
+  DOCUMENTS = process.env.DOCUMENTS.substr(1).split('/');
+  NAMESPACE_ASSETS = process.env.NAMESPACE.split('/')[0] + '/' + 'assets';
```
sophstad: [nit] May as well use a template literal:
```js
NAMESPACE_ASSETS = `${process.env.NAMESPACE.split('/')[0]}/assets`;
```

src/components/Figure.js
```diff
@@ -1,13 +1,16 @@
...
 
-const Figure = ({ nodeData }) => (
-  <img
-    src={nodeData.argument[0].value}
-    alt={nodeData.options.alt ? nodeData.options.alt : nodeData.argument[0].value}
-    width="50%"
-  />
-);
+const Figure = ({ nodeData }) => {
+  const fullImagePath = (process.env.GATSBY_PREFIX ? process.env.GATSBY_PREFIX : '') + nodeData.argument[0].value;
```
sophstad: In JavaScript, instead of using a ternary `if x ? x : y`, you can use `x || y`. So this would be
```js
const fullImagePath = process.env.GATSBY_PREFIX || ''  + nodeData.argument[0].value;
```

src/html.js
```diff
@@ -21,17 +21,17 @@ const HTML = ({ body, bodyAttributes, headComponents, htmlAttributes, preBodyCom
...
+      <link rel="stylesheet" href={`${process.env.GATSBY_PREFIX}/static/css/navbar.min.css`} type="text/css" />
       <link
         rel="search"
         type="application/opensearchdescription+xml"
         href="https://docs.mongodb.com/osd.xml"
         title="MongoDB Help"
       />
-      {headComponents}
     </head>
+    {headComponents}
```
sophstad: Why does `{headComponents}` go outside of the `<head>` tags?

gatsby-config.js
danielborowski: This is because for some reason, running gatsby serve does not set NODE_ENV = production like it [says it should](https://www.gatsbyjs.org/docs/environment-variables/). So we explicitly specify what NODE_ENV to use in the .env.* files now, but if it's not specified (like when running gatsby serve) we actually want to use production data with the correct prefix and not build for localhost.

src/html.js
danielborowski: fixed

gatsby-node.js
```diff
@@ -51,15 +51,15 @@ const validateEnvVariables = () => {
     };
   }
   // make sure formats are correct
-  if (process.env.NODE_ENV === 'production' && !process.env.GATSBY_PREFIX.startsWith('/')) {
+  if (process.env.NODE_ENV === 'production' && (!process.env.GATSBY_PREFIX.startsWith('/') || !process.env.DOCUMENTS.startsWith('/'))) {
```
sophstad: I think it's worth splitting this into two separate checks so we can give the user a more specific error.

After that is done and merge conflicts are resolved, LGTM!

PR #39, opened 2019-04-16
DOCSP-5259: Fix roles

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-5259)] This PR:
- Separates `RoleLink` into `RoleDoc`, `RoleManual`, and `RoleTerm`
- Adds support for the `update` role
- Improves clarity in the `Code` role via destructuring props
- Adds tests to support all Role components
- Introduces util function `findKeyValuePair` to find objects at any depth and replaces instances where a similar function was being used


src/components/LandingPageCards.js
```diff
@@ -43,7 +27,7 @@ const Category = ({ cards, category, refDocMapping }) =>
           let completionTime;
           if (card.name === 'card') {
             const cardSlug = card.argument[0].value;
-            completionTime = findGuideProperty(refDocMapping[cardSlug].ast.children, 'time').argument[0].value;
+            completionTime = findKeyValuePair(refDocMapping[cardSlug].ast.children, 'name', 'time').argument[0].value;
```
danielborowski: Do you think we should set some default `completionTime` because if it doesn't exist or if it's incorrectly entered as empty string then the Card component will display undefined or empty string right? Or do we just assume it will exist and always be correct?

src/components/Roles/Doc.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { findKeyValuePair } from '../../util';
+
+const RoleDoc = ({ nodeData: { label, target }, refDocMapping }) => {
+  const labelDisplay =
```
danielborowski: propTypes at bottom says label can be of type string or object, so what if label = "hello" then this might not be doing the correct thing?

src/templates/guide.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import TOC from '../components/TOC';
 import GuideSection from '../components/GuideSection';
 import GuideHeading from '../components/GuideHeading';
-import { LANGUAGES, DEPLOYMENTS, REF_TARGETS } from '../constants';
+import { LANGUAGES, DEPLOYMENTS } from '../constants';
```
danielborowski: good idea to remove `REF_TARGETS` from `__refDocMapping`

src/components/LandingPageCards.js
sophstad: Yeah good idea! I will set a default of 15 minutes since that seems to be a common time estimate for our guides.

src/components/Roles/Doc.js
sophstad: Good point—I forgot to update the PropTypes check, but it can only be a string

danielborowski:
LGTM

PR #41, opened 2019-04-24
DOCSP-4305: Log build errors

danielborowski:


sophstad:
Based on what you told me about your conversation with Sue, it sounds like she has set up logging errors with Docker based on what is output to stderr via the console, e.g.
```js
console.error('Error message here');
```

Many of the errors in this file are returned via `throw Error()` rather than the console. Would we still expect these to work with the Docker setup?

PR #43, opened 2019-04-29
DOCSP-5257: Show language pills on landing page

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-5257)]
- Show pills on landing page and save to local storage on click (new component: `src/components/LandingPage/CardPills.js`)
- Reorganize landing page components into a new directory: `src/components/LandingPage/`
- Move pills (in header and on landing page) into their own `Pills` component (and add tests)
- Fix broken landing page introduced by an AST update


src/components/GuideHeading.js
```diff
@@ -1,100 +1,79 @@
...
-    const { setActiveTab, cloud, drivers, activeTabs } = this.props;
-    return (
-      <div className="section" id="SOMETHING_HERE">
-        <h1>
-          {this.getSection(['type', 'heading']).children[0].value}
-          <a className="headerlink" href="#read-data-from-mongodb" title="Permalink to this headline">
-            ¶
-          </a>
-        </h1>
+  const setActiveCloudPill = pill => {
```
danielborowski: Can this and function below potentially be one a single function with second param specifying type (cloud/drivers/etc.)?

src/components/Pills.js
```diff
@@ -0,0 +1,50 @@
...
+  pills: PropTypes.arrayOf(PropTypes.string).isRequired,
+  ulClass: PropTypes.string,
+};
+
+Pills.defaultProps = {
+  activeClass: 'guide__pill--active',
+  activePill: undefined,
+  dataTabPreference: undefined,
+  isTruncated: false,
+  liClass: undefined,
```
danielborowski: Should these be default `undefined` or empty strings? How do the `undefined` values work when used above on line 10?

src/components/LandingPage/CardPills.js
```diff
@@ -0,0 +1,64 @@
...
+    let isTruncated = false;
+    const pills = [];
+    node.children[0].children.forEach(pillObj => {
+      const pill = getPillTitle(pillObj);
+      if (totalLength > 50) {
+        isTruncated = true;
+      } else {
+        pills.push(pill);
+      }
+      totalLength += stringifyTab(pill).length + 4;
```
danielborowski: Maybe just add a comment here to say that the 4 (set it to some variable even called paddingLength potentially?) is to take into account for padding in the string between pills.

src/components/Pills.js
sophstad: Good point! They should be empty strings.

danielborowski:
LGTM

PR #46, opened 2019-05-01
Fix: Env Vars

sophstad:
- Modify env vars so that we can use `SITE`, `USER`, and `BRANCH` variables independently within the front end.
- `STITCH_ID` renamed to `GATSBY_STITCH_ID` to expose it to the front end.
- `GATSBY_PREFIX` no longer required.
  - It will be used if specified.
  - If not, `production` environments will default to `/SITE/USER/BRANCH` and `development` environments will default to `/`.
- No longer set `NODE_ENV` in .env files, as this is a [reserved environment variable](https://www.gatsbyjs.org/docs/environment-variables/#reserved-environment-variables).
- Remove some deprecated GitHub fetch code.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -16,7 +16,6 @@ export const findKeyValuePair = (nodes, key, value) => {
   return result;
 };
 
-// export const getPrefix = () => process.env.GATSBY_PREFIX || ''
 export const getPrefix = () =>
   process.env.GATSBY_PREFIX ||
   (process.env.NODE_ENV === 'production'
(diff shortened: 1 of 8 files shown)


src/util.js
```diff
@@ -16,4 +16,9 @@ const findKeyValuePair = (nodes, key, value) => {
   return result;
 };
 
-export { findKeyValuePair }; // eslint-disable-line import/prefer-default-export
+// export const getPrefix = () => process.env.GATSBY_PREFIX || ''
```
danielborowski: whats this comment?

danielborowski:
tested it and works great, LGTM

src/util.js
sophstad: Good catch, removed!

PR #47, opened 2019-05-03
DOCSP-5425: Add missing roles

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-5425)] Add 3 missing roles:
- `java-sync-api`
- `csharp-api`
- `py:class`

Bonus fix: catches nodes of type `comment` and returns nothing.

### DIFF
--- a/src/components/Roles/Api.js
+++ b/src/components/Roles/Api.js
@@ -1,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const RoleJavaSyncApi = ({
+  nodeData: {
+    label: { value },
+    name,
+    target,
+  },
+}) => {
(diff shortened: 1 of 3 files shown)


src/components/Roles/Api.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const RoleJavaSyncApi = ({
```
danielborowski: Maybe a more generic name can be used instead of Java?

sophstad: Whoops yeah! Renamed the file but not the component

danielborowski:
LGTM

PR #48, opened 2019-05-08
DOCSP-5460: syntax highlighting

danielborowski:


package.json
```diff
@@ -39,6 +39,9 @@
   },
   "jest": {
     "verbose": true,
+    "moduleNameMapper": {
```
sophstad: Is this bit still used in your implementation? `__mockStyles.js` looks empty.

package.json
```diff
@@ -68,12 +71,14 @@
     "eslint-plugin-jest": "^22.3.0",
     "express": "^4.16.4",
     "gatsby": "^2.3.4",
+    "highlight.js": "^9.15.6",
```
sophstad: We don't need to include `highlight.js` because it is a dependency of `react-highlight` ([source](https://github.com/akiran/react-highlight/blob/master/package.json#L28)). I tested after removing this line and reinstalling packages and it still works (you ought to test as well to verify).

src/html.js
```diff
@@ -25,6 +25,17 @@ const HTML = ({ body, bodyAttributes, headComponents, htmlAttributes, preBodyCom
       <link rel="stylesheet" href={`${getPrefix()}/static/guides.css`} type="text/css" />
       <link rel="stylesheet" href={`${getPrefix()}/static/pygments.css`} type="text/css" />
       <link rel="stylesheet" href={`${getPrefix()}/static/css/navbar.min.css`} type="text/css" />
+      <style
+        dangerouslySetInnerHTML={{
+          __html: `
+          .hljs {
+            background: none !important;
```
sophstad: I would assume these `!important`s were required to override the package styles?

danielborowski: Needed to overwrite the global .hljs styles added.

PR #49, opened 2019-05-09
Fix: Update roles to match new AST spec

sophstad:
Update roles to address changes made to the parser in [DOCSP-5504](https://jira.mongodb.org/browse/DOCSP-5504). Most notably, the label structure of `Role` nodes has changed:
- If it exists, `label` will always be an object of the following form:
```js
label: {
  value: 'Title string is here'
}
```
- If no label has been provided in the rST, the object **will not** contain a `label` property
- Labels will **never** be strings: they are either objects, or the property does not exist

Also adds two `PropTypes` fixes to `Admonition` and `ListTable` components.


danielborowski:
LGTM

PR #50, opened 2019-05-09
DOCSP-5429: copy button

danielborowski:


src/components/Code.js
```diff
@@ -14,28 +15,82 @@ const URI_PLACEHOLDERS = [
...
-          copy
-          <div className="code-button__tooltip code-button__tooltip--inactive">copied</div>
-        </button>
-      </div>
-      <div className="copyable-code-block highlight-python notranslate">
-        <div className="highlight">
-          <pre>{code}</pre>
+  copyCodeButton = code => {
+    if (!document) return;
+    const tempElement = document.createElement('textarea');
```
sophstad: I think this is the approach docs-tools is using, but we can do it much more efficiently using React (since we're not trying to find an element and inject a copy button on a static page). [This StackOverflow answer](https://stackoverflow.com/a/42844911) looks like a good approach—I like the use of a `ref`. You could even try using hooks! 😄

src/components/Code.js
```diff
@@ -14,28 +15,82 @@ const URI_PLACEHOLDERS = [
...
+      });
+      // hide after some time passes
+      setTimeout(() => {
+        this.setState({
+          copied: false,
+        });
+      }, 1500);
+    } catch (err) {
+      console.error(err);
+    }
```
sophstad: We probably want a `finally` clause here to remove the textarea:
```js
document.body.removeChild(tempElement);
```

src/components/Code.js
skerschb: I agree that the hook version looks much lighter. Let's add a //TODO for now in the interest of getting guides into QA, and talk generally about where we can leverage hooks as part of the CET migration. Unless you've already started to rework this Daniel.

sophstad: The approach I linked to above won't actually work because we need to spoof a `textarea`, but we could still use hooks (and have used them in a few components). I support focusing on implementing hooks with CET.

PR #51, opened 2019-05-13
Fix: Remove required value property

sophstad:
The `Class` role (updated in DOCSP-5425) does not require a `label.value` property, which was previously marked as required and lead to page render failure. If `label.value` is not specified, use the role's `target` property as the link text.


danielborowski:
LGTM

PR #53, opened 2019-05-15
DOCSP-5547: URI Writer fixes

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-5547)] [[Staging](https://docs-mongodborg-staging.corp.mongodb.com/fix-uriwriter/sophstad/DOCSP-5547/server/drivers)]
- Move URI components into their own directory (`components/URIWriter/`)
  - Separate `LocalForm` and `CloudForm` components
- Properly display URI strings on load when fetched from local storage
- Populate forms on load from local storage
- Prevent invalid hostnames from appearing in the populated URI string
- Separate storage of cloud/local URI parameters


src/components/URIWriter/CloudForm.js
```diff
@@ -0,0 +1,249 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { TEMPLATE_TYPE_ATLAS_34, TEMPLATE_TYPE_ATLAS_36 } from './constants';
+import { getLocalValue, setLocalValue } from '../../localStorage';
```
skerschb: Could we add a translation layer in the localStorage component to check for the existence of the old style of localStorage and copy it to the new? We can create a new ticket for this.

sophstad: [DOCSP-5622](https://jira.mongodb.org/browse/DOCSP-5622)

PR #54, opened 2019-05-16
DOCSP-5575: Fix TOC

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-5575)] [[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/fix-toc/sophstad/DOCSP-5575/server/install)] Lots of fixes:
- Fix TOC: display only sections that appear in the guide, not all sections statically
- Dynamically highlight table of contents
- Remove jQuery, Bootstrap, and `controller.js` dependencies
- Refactor `GuideHeading` component so that we don't need to pass all section data as a prop
- Move several objects into `src/constants.js` (e.g. admonitions, section slug -> title mapping)
- Fix IDs to allow for internal anchor linking (using the same formatting as V1 so that internal page links remain valid)
- Fix tabs: use page offsets so that when a tab is selected, the page will not "jump"


src/components/GuideHeading.js
```diff
@@ -3,21 +3,19 @@ import PropTypes from 'prop-types';
 import ComponentFactory from './ComponentFactory';
 import Pills from './Pills';
 import { stringifyTab } from '../constants';
+import { slugifyTitle } from '../util';
```
skerschb: we will need to add a //TODO item for all (@i80and @danielborowski @sophstad )-- 1. semantic tagging for templated content at the AST level. It's the only way we will be able to offer true value/benefit to separation of concerns. 2. validation of the template content at the render level -- we will need the templated content to be validated against content template versions.

src/components/GuideHeading.js
```diff
@@ -3,21 +3,19 @@ import PropTypes from 'prop-types';
...
-  const getSection = params => {
-    return sections.filter(s => s[params[0]] === params[1])[0];
-  };
-
+const GuideHeading = ({ activeTabs, author, cloud, description, drivers, setActiveTab, time, title, ...rest }) => {
   const setActivePill = pillsetName => pill => {
     setActiveTab(pill, pillsetName);
   };
 
+  const displayTitle = title.children[0].value;
```
skerschb: we will need to add some validation here. //TODO item

src/components/TOC.js
```diff
@@ -1,45 +1,34 @@
 import React from 'react';
+import PropTypes from 'prop-types';
+import { SECTION_NAME_MAPPING } from '../constants';
+import { slugifyTitle } from '../util';
 
-const TOC = () => (
+const TOC = ({ activeSection, sectionKeys }) => (
```
skerschb: nice abstraction and this component is now reusable. did highlight the need for validation of the templated content types ("guide" is the only one so far but we should probably tackle this in a sprint or two).

src/util.js
```diff
@@ -30,3 +30,43 @@ export const getPrefix = () =>
   (process.env.NODE_ENV === 'production'
     ? `/${process.env.GATSBY_SITE}/${process.env.GATSBY_USER}/${process.env.GATSBY_BRANCH}`
     : '');
+
+export const throttle = (func, wait) => {
```
skerschb: What does this do? We should put a comment here.

sophstad: I too would like to know what it does... [here](https://github.com/mongodb/docs-tools/blob/master/themes/mongodb/src/js/util.js#L30) is the source from `docs-tools`. @i80and any hints?

src/components/GuideHeading.js
sophstad: We do validation using `PropTypes` at the bottom of the file... do you think this is adequate? I've been wondering myself.

skerschb: Yes this is good. I'm thinking we will need something a little more specific -- field lengths, etc. It's a tricky problem because presumably we might get content for render that we have not put through our AST parser (for example automated scripts populating the db with AST generated from source code). The only place really to put this extra validation and make sure we catch any issues would be in the render layer.

skerschb: But...we need to design this. So for now LGTM!

PR #57, opened 2019-05-21
Fix Links

sophstad:
[[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/guides/sophstad/fix-links/)] Roles of type `Doc` are internal links and require prefixing to work in staging.
- In a build environment, prefix them with `GATSBY_PREFIX`, if specified, or `GATSBY_SITE/GATSBY_USER/GATSBY_BRANCH` if not (all of which are defined in the `.env.production` file).
- In a local development environment, do not include a prefix.

This is a stopgap until relative links are implemented.


danielborowski:
LGTM

PR #60, opened 2019-05-28
DOCSP-5689: Fix lightbox component

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-5689)] [[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/guides/sophstad/DOCSP-5689/)] Fix bug where an `img`'s `onLoad` property was not being called (unless run in `develop` environment). As a result, large images were never being converted into Lightbox components.

Also moves the Feedback widget to the proper place in the DOM (where it is currently found in V1).


danielborowski:
LGTM

PR #62, opened 2019-05-28
DOCSP-5640: segment + stitch fix so works on IE11

danielborowski:


src/components/GuideSection.js
```diff
@@ -102,7 +102,9 @@ export default class GuideSection extends Component {
 }
 
 GuideSection.propTypes = {
-  headingRef: PropTypes.shape({ current: PropTypes.element }).isRequired,
+  headingRef: PropTypes.shape({
+    current: PropTypes.oneOfType([PropTypes.element, PropTypes.object]),
```
sophstad: I included a fix for this `PropTypes` error in #63 ([here](https://github.com/mongodb/snooty/pull/63/files#diff-882df33072db4648118937dbb9147411R101)). I think that fix may better enforce prop types in this instance, so perhaps remove this change from your PR.

src/components/Widgets/Deluge/Deluge.js
```diff
@@ -37,11 +37,15 @@ class Deluge extends Component {
...
-    this.stitchClient.auth.loginWithCredential(new AnonymousCredential()).catch(err => {
-      console.error(err);
-    });
+    try {
+      const appId = 'feedback-ibcyy';
+      this.stitchClient = Stitch.hasAppClient(appId) ? Stitch.getAppClient(appId) : Stitch.initializeAppClient(appId);
+      this.stitchClient.auth.loginWithCredential(new AnonymousCredential()).catch(err => {
+        console.error(err);
+      });
+    } catch (error) {
```
sophstad: Let's make sure to hide the Feedback Widget if this fails.

PR #66, opened 2019-05-30
DOCSP-5644: no abnormal focus

danielborowski:


.eslintrc.js
```diff
@@ -27,6 +27,8 @@ module.exports = {
     'no-underscore-dangle': 0,
     'prefer-destructuring': 0,
     'import/no-cycle': 0,
+    'jsx-a11y/anchor-is-valid': 0,
```
sophstad: Instead of ignoring these rules for the entire codebase, let's only ignore the lines that throw errors. You can find lots of examples of `eslint-disable-line` in our repo.

src/components/Code.js
```diff
@@ -75,9 +75,9 @@ export default class Code extends Component {
     return (
       <div className="button-code-block">
         <div className="button-row">
-          <button
+          <a
```
sophstad: Let's include a TODO for changing this link to a button when we are working on CSS more actively.

.eslintrc.js
skerschb: What are the exceptions we are trying to avoid with the rules changes?

sophstad: Because of the way our CSS is written, it looks like we'll have to keep the "copy" button as a `<a>` tag rather than a `<button>`, which is more correct and accessible. The a11y warnings are because we are using an `<a>` for something other than navigation, so we need to suppress them.

PR #71, opened 2019-06-06
DOCSP-5719: Resolve staging config + much more!

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-5719)] [[Staging](https://docs-mongodborg-staging.corp.mongodb.com/guides/sophstad/DOCSP-5719/)] Lots of changes:
- Reduce complexity of environment variables
  - Compute `user` and `branch` values in `gatsby-config.js`
  - Make these variables available to the front-end using Gatsby's [`siteMetadata`](https://www.gatsbyjs.org/docs/gatsby-config/#sitemetadata) object
- Reconfigure `utils`: convert from one `util.js` file to a directory with separate files per function
  - This was required to easily test components that make use of the `siteMetadata` above.
- Modify arguments passed to Stitch functions
- Begin using [React Helmet](https://github.com/nfl/react-helmet) to dynamically set `<body>` attributes (previously done in `html.js`, but it had to be moved in order to make use of `siteMetadata`)
- Mock relevant util functions


gatsby-config.js
```diff
@@ -1,15 +1,32 @@
...
+const getPathPrefix = () => {
+  const user = userInfo().username;
+  const gitBranch = getGitBranch();
+
+  return runningEnv === 'production' ? `/${process.env.SITE}/${user}/${gitBranch}` : '/';
+};
 
 module.exports = {
   pathPrefix: getPathPrefix(),
+  plugins: [`gatsby-plugin-react-helmet`],
```
danielborowski: just use normal string rather than backticks if no vars are used within

PR #73, opened 2019-06-07
Fix path prefix

sophstad:
[[Staging](https://docs-mongodborg-staging.corp.mongodb.com/guides/sophstad/fix-get-path-prefix/)]
- Replace custom `getPathPrefix()` with Gatsby's (poorly documented) [`withPrefix()`](https://www.gatsbyjs.org/docs/gatsby-link/#add-the-path-prefix-to-paths-using-withprefix)
  - Remove relevant mocks
  - Update image src attributes and `<head>` links using this function
- Replace our one instance of an internal link with a Gatsby [`Link`](https://www.gatsbyjs.org/docs/gatsby-link/) component
- Fix tests

### DIFF
--- a/__mocks__/gatsby.js
+++ b/__mocks__/gatsby.js
@@ -1,8 +1,9 @@
-const gatsby = jest.requireActual("gatsby")
+const gatsby = jest.requireActual('gatsby');
 
 module.exports = {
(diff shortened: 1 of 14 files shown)


danielborowski:
LGTM

PR #77, opened 2019-06-13
DOCSP-5843: Fix nested object access

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-5843)] [[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/guides/sophstad/DOCSP-5843/)]
- Add new util function `getNestedValue()`
- Replace all instances of accessing nested values in objects with this function
- Ignore `default-domain` directive (per Andrew's recommendation)

### DIFF
--- a/src/components/Tabs.js
+++ b/src/components/Tabs.js
@@ -3,12 +3,13 @@ import PropTypes from 'prop-types';
 import ComponentFactory from './ComponentFactory';
 import { PLATFORMS, stringifyTab } from '../constants';
 import { reportAnalytics } from '../utils/report-analytics';
+import { getNestedValue } from '../utils/get-nested-value';
 
 export default class Tabs extends Component {
   constructor(props) {
     super(props);
     const { nodeData, addTabset } = this.props;
-    const tabsetName = nodeData.options ? nodeData.options.tabset : this.generateAnonymousTabsetName(nodeData);
+    const tabsetName = getNestedValue(['options', 'tabset'], nodeData) || this.generateAnonymousTabsetName(nodeData);
     this.state = { tabsetName };
 
     addTabset(tabsetName, [...nodeData.children]);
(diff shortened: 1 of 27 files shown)


src/templates/guide.js
```diff
@@ -151,7 +164,7 @@ export default class Guide extends Component {
                 cloud={cloud}
                 description={findKeyValuePair(this.sections, 'name', 'result_description')}
                 drivers={drivers}
-                refDocMapping={pageContext ? pageContext.__refDocMapping : {}}
+                refDocMapping={getNestedValue(['__refDocMapping'], pageContext) || {}}
```
danielborowski: Should we ever actually assign refDocMapping to be an empty object? Wouldn't the entire site just not render then?

sophstad: There are a decent number of components that can be rendered without ever referencing `refDocMapping`. This instance came up mostly having to do with that mLab orphan page we were discussing yesterday. Even if the `refDocMapping` cannot be found (for whatever reason), we can still render other page elements for orphan pages. Returning `null` instead of an empty object could potentially lead to some `PropTypes` warnings which is why I kept it as an empty object.

danielborowski: ok perfect

src/components/Tabs.js
```diff
@@ -52,7 +53,7 @@ export default class Tabs extends Component {
         {isHeaderTabset || isHidden || (
           <ul className="tab-strip tab-strip--singleton" role="tablist">
             {tabs.map((tab, index) => {
-              const tabName = tab.argument[0].value.toLowerCase();
+              const tabName = getNestedValue(['argument', 0, 'value'], tab).toLowerCase();
```
danielborowski: If this returns null because it doesn't exist, it'll error trying to call `toLowerCase` on null right?

danielborowski:
LGTM

src/components/Tabs.js
sophstad: Yeah I was thinking about this.... I guess I need to check if the variable exists before calling `toLowerCase()`

PR #87, opened 2019-06-25
DOCSP-5772: Implement replace functionality

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-5772)] [[Staging](https://docs-mongodborg-staging.corp.mongodb.com/spark-connector/sophstad/DOCSP-5772/)] Improvements are visible on the [`/java/aggregation`](https://docs-mongodbcom-staging.corp.mongodb.com/spark-connector/sophstad/DOCSP-5772/java/aggregation) page.
- Parse page tree to find all `substitution_definition` nodes
- Parse all includes to find `substitution_definition` nodes
- Save these definitions in a `substitutions` map, which is passed as a prop to the `ComponentFactory`
- When the component factory encounters a `substitution_reference` node, replace it using the child nodes saved in the `substitutions` mapping
- Ignore `substitution_definition` nodes so they are not rendered on the page

### DIFF
--- a/src/components/ComponentFactory.js
+++ b/src/components/ComponentFactory.js
@@ -34,6 +34,9 @@ import RoleProgram from './Roles/Program';
 import RoleRef from './Roles/Ref';
 import RoleTerm from './Roles/Term';
 
+const IGNORED_NAMES = ['class', 'cssclass', 'default-domain'];
+const IGNORED_TYPES = ['class', 'comment', 'cssclass', 'substitution_definition', 'target'];
+
 export default class ComponentFactory extends Component {
   constructor() {
     super();
@@ -98,16 +101,7 @@ export default class ComponentFactory extends Component {
     }
 
     // do nothing with these nodes for now (cc. Andrew)
-    if (
-      type === 'target' ||
-      type === 'class' ||
-      type === 'cssclass' ||
-      name === 'cssclass' ||
-      name === 'class' ||
-      type === 'comment' ||
-      name === 'default-domain' ||
-      type === 'substitution_definition'
-    ) {
+    if (IGNORED_TYPES[type] !== undefined || IGNORED_NAMES[name] !== undefined) {
       return null;
     }
 
--- a/src/components/ComponentFactory.js
+++ b/src/components/ComponentFactory.js
@@ -101,7 +101,7 @@ export default class ComponentFactory extends Component {
     }
 
     // do nothing with these nodes for now (cc. Andrew)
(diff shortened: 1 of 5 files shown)


src/utils/get-include-file.js
```diff
@@ -0,0 +1,9 @@
+import { getNestedValue } from './get-nested-value';
+
+export const getIncludeFile = (refDocMapping, filename) => {
+  let key = filename;
+  if (key.startsWith('/')) key = key.substr(1);
```
i80and: Not sure if it matters right now, but if a path doesn't start with a slash, it's relative to the file requesting the path. It's possible we want the parser to make all paths absolute, but I don't think that happens right now. Might be worth logging a warning if it doesn't start with /

src/utils/get-include-file.js
```diff
@@ -0,0 +1,9 @@
...
+  if (key.endsWith('.rst')) key = key.replace('.rst', '');
```
i80and: Should we always strip the suffix?

src/templates/document.js
```diff
@@ -3,14 +3,33 @@ import PropTypes from 'prop-types';
...
 
 const Document = props => {
   const {
     '*': pageSlug,
     pageContext: { __refDocMapping },
   } = props;
   const pageNodes = getNestedValue([pageSlug || 'index', 'ast', 'children'], __refDocMapping) || [];
+
+  const getSubstitutions = () => {
+    const fileSubstitutions = findAllKeyValuePairs(pageNodes, 'type', 'substitution_definition');
```
i80and: This function is a little complex; could we have a couple of comments here just briefly running through things in prose?

src/utils/get-include-file.js
sophstad: I can't find any documents that include file suffixes in their `_id` field. Is it possible that Snooty would include that in an id?

src/components/ComponentFactory.js
```diff
@@ -34,6 +34,9 @@ import RoleProgram from './Roles/Program';
 import RoleRef from './Roles/Ref';
 import RoleTerm from './Roles/Term';
 
+const IGNORED_NAMES = ['class', 'cssclass', 'default-domain'];
```
i80and: These are arrays, so I don't think line 104 will work. You can do either of the following:
```
const IGNORED_NAMES = {'class': true, 'cssclass': true...
const IGNORED_NAMES = new Set(['class', 'cssclass', ...
```

src/utils/get-include-file.js
i80and: Snooty currently strips suffixes

PR #89, opened 2019-07-02
DOCSP-5936: Add escape key support to image lightbox

rayangler:
Ticket: [https://jira.mongodb.org/browse/DOCSP-5936](https://jira.mongodb.org/browse/DOCSP-5936)
Staging: [https://docs-mongodbcom-staging.corp.mongodb.com/guides/raymundrodriguez/DOCSP-5936/cloud/migrate-from-aws-to-atlas](https://docs-mongodbcom-staging.corp.mongodb.com/guides/raymundrodriguez/DOCSP-5936/cloud/migrate-from-aws-to-atlas)

### DIFF
--- a/src/components/Lightbox.js
+++ b/src/components/Lightbox.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useRef, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { withPrefix } from 'gatsby';
 import { getNestedValue } from '../utils/get-nested-value';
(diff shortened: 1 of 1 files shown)


src/components/Lightbox.js
```diff
@@ -10,11 +10,26 @@ const Lightbox = ({ nodeData }) => {
...
   };
 
+  const handleOnKeyDown = e => {
+    // Escape key
+    if (e.keyCode === 27) {
+      toggleShowModal();
+    }
+  };
+
+  // Hook to take effect with ever re-render
```
sophstad: [nit] Typo I think! Did you mean "every"?

src/components/Lightbox.js
```diff
@@ -10,11 +10,26 @@ const Lightbox = ({ nodeData }) => {
...
+  React.useEffect(() => {
```
sophstad: You could import `useEffect` at the top if you like so that you don't need the `React.` prefix here.

src/components/Lightbox.js
```diff
@@ -24,7 +39,15 @@ const Lightbox = ({ nodeData }) => {
...
       {showModal && (
-        <div className="lightbox__modal" title="click to close" onClick={toggleShowModal} role="button" tabIndex="-1">
+        <div
+          className="lightbox__modal"
+          title="click to close"
+          onClick={toggleShowModal}
+          ref={modal}
+          onKeyDown={handleOnKeyDown}
+          role="button"
+          tabIndex="0"
```
sophstad: Does your implementation still work if we keep `tabIndex="-1"`? If so, let's avoid changing it since I don't think we need the Lightbox to be in a screen reader flow.

rayangler: Still works with -1

PR #90, opened 2019-07-03
DOCSP-5996: Improve build performance

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-5996)] Improve performance of builds
- Minimize `refDocMapping`: only provide each page with its own data in `gatsby-node.js`
  - Update usage of `refDocMapping` throughout the site
- Introduce `pageMetadata` object, which includes each page's slug, title, time estimate, and category. This is used to generate the Guides landing page as well as internal links (as seen in `Roles/Doc.js`).
- Fetch all image assets in one call to Stitch, rather than making a separate request for each image
  - Generate intermediate directories if the image filename requires it


gatsby-node.js
```diff
@@ -58,8 +61,12 @@ const validateEnvVariables = () => {
 
 const saveAssetFile = async (name, objData) => {
   return new Promise((resolve, reject) => {
-    fs.writeFile(path.join('static', name), objData.data.buffer, 'binary', err => {
-      if (err) reject(err);
+    // Create nested directories as specified by the asset filenames if they do not exist
+    mkdirp(path.join('static', path.dirname(name)), err => {
```
skerschb: Do we know what permissions level these files have by default? Should not be higher than 755

sophstad: Permissions are 644 by default 👍

PR #91, opened 2019-07-08
DOCSP-5934: Clicking on left hand nav item scrolls correctly

rayangler:
Staging link: [https://docs-mongodbcom-staging.corp.mongodb.com/guides/raymundrodriguez/DOCSP-5934/](https://docs-mongodbcom-staging.corp.mongodb.com/guides/raymundrodriguez/DOCSP-5934/)

Ticket link: [https://jira.mongodb.org/browse/DOCSP-5934](https://jira.mongodb.org/browse/DOCSP-5934)

### DIFF
--- a/src/components/TOC.js
+++ b/src/components/TOC.js
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { SECTION_NAME_MAPPING } from '../constants';
 
-const TOC = ({ activeSection, sectionKeys }) => (
+const TOC = ({ activeSection, sectionKeys, disableScrollable }) => (
   <aside className="left-toc hide-first-toc-level">
     <div className="left-toc__title">Overview:</div>
     <ul>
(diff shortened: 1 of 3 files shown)


src/templates/guide.js
```diff
@@ -120,6 +127,14 @@ export default class Guide extends Component {
     setLocalValue(tabsetName, activeTab);
   };
 
+  // Temporarily disable scrolling listener by
```
sophstad: Looks like you were in the middle of a thought here :)

src/components/TOC.js
```diff
@@ -24,10 +30,12 @@ const TOC = ({ activeSection, sectionKeys }) => (
 TOC.propTypes = {
   activeSection: PropTypes.string,
   sectionKeys: PropTypes.arrayOf(PropTypes.string).isRequired,
+  disableScrollable: PropTypes.func,
 };
 
 TOC.defaultProps = {
   activeSection: undefined,
+  disableScrollable: undefined,
```
sophstad: Since `disableScrollable` is a function, it is a good idea to set its default prop as a no-op function so that we don't get an error when trying to call `undefined`:
```js
disableScrollable: () => {}
```

src/templates/guide.js
```diff
@@ -120,6 +127,14 @@ export default class Guide extends Component {
...
+  disableScrollable = newSection => {
```
sophstad: [nit] Perhaps `clickedSection` would make more sense as the parameter here.

PR #98, opened 2019-07-16
DOCSP-5935: Code blocks show when javascript is disabled

rayangler:
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/guides/raymundrodriguez/DOCSP-5935/server/install/) [Ticket](https://jira.mongodb.org/browse/DOCSP-5935)

Testing with javascript disabled results in a blank screen on localhost. Results do show up in the staging link though.


src/components/Code.js
```diff
@@ -102,17 +103,20 @@ export default class Code extends Component {
...
+      if (isBrowser()) code = this.htmlDecode(code);
+      isJsEnabled = Object.keys(activeTabs).length !== 0;
     }
+
     // TODO: when we move off docs-tools CSS, change the copy button from <a> to <button>
     return (
       <div className="button-code-block">
         <div className="button-row">
-          {showCopyButton && (
+          {showCopyButton && isJsEnabled && (
```
sophstad: I'm not sure it's a good idea to semantically couple whether JS is enabled with whether `activeTabs` is set (even thought it adds up logically). At the very least, this variable could be named `hasActiveTabs`. But...

I think it would be a good idea to move this logic up to `isCopyButtonEnabled()`. That function is responsible for the same thing as this variable, so we ought to have the logic around `activeTabs` as an `else if` block there.

src/components/Tabs.js
```diff
@@ -112,15 +124,17 @@ export default class Tabs extends Component {
...
-              </React.Fragment>
-            )
-          );
+
+          // If there are no activeTabs, js would typically be disabled
+          let tabContent = this.createFragment(tab, index);
+          if (activeTabs) {
+            tabContent =
+              Object.keys(activeTabs).length === 0
+                ? this.createFragment(tab, index)
```
sophstad: Can we avoid calling `createFragment` with the same arguments twice?
```js
const tabContent =
  !activeTabs || Object.keys(activeTabs).length === 0
    ? this.createFragment(tab, index)
    : activeTabs[tabsetName] === tabName && this.createFragment(tab, index);
```

PR #102, opened 2019-07-31
DOCSP-6332: Add misc. charts directives

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-6332)] [[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/charts/sophstad/DOCSP-6332/)] Regression tested on guides and charts.
- Change Stitch ID to begin using new Stitch app after failed migration
- Refactor admonition component to more easily add `seealso` and `example` admonitions
- Make CSSWrapper generic so that it is easy to add a class to any unknown DOM element
- Add horizontal list (`hlist`) component
- Add line block component
- Fix PropTypes warnings in ListTable component


src/components/CSSWrapper.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { findDOMNode } from 'react-dom';
+import PropTypes from 'prop-types';
+
+export default class CSSWrapper extends React.Component {
+  componentDidMount() {
+    const { className } = this.props;
+
+    /*
+     * Using findDOMNode() is discouraged in favor of passing refs to children, but we use it here
```
rayangler: I thought you're able to use ref hooks for stateless components. Or am I thinking of something else? Just wondering

sophstad: We could use a ref hook to attach a ref to a _DOM element_ within a functional component, but we can't attach a ref to the component itself ([source](https://reactjs.org/docs/refs-and-the-dom.html#refs-and-function-components)). Good question!

PR #103, opened 2019-08-08
DOCSP-6379: Page contents

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-6379)] [[Charts Staging](https://docs-mongodbcom-staging.corp.mongodb.com/charts/sophstad/DOCSP-6379/)]


src/components/Tabs.js
```diff
@@ -21,7 +21,7 @@ export default class Tabs extends Component {
     if (addTabset !== undefined) {
       addTabset(tabsetName, [...nodeData.children]);
     } else {
-      setActiveTab(nodeData.children[0].argument[0].value, tabsetName);
+      setActiveTab(tabsetName, nodeData.children[0].argument[0].value);
```
sophstad: FYI this is a bugfix; function arguments were being passed in the wrong order.

src/components/Contents.js
```diff
@@ -0,0 +1,103 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { getNestedValue } from '../utils/get-nested-value';
+
+const ContentsListItem = ({ id, listChildren, title }) => (
+  <li>
+    <a href={`#${id}`}>{title}</a>
```
i80and: Technically the title can contain arbitrary formatting! I think it's probably okay to ignore that possibility since it's dumb

src/components/Contents.js
```diff
@@ -0,0 +1,103 @@
...
+    PropTypes.shape({
+      children: PropTypes.arrayOf(PropTypes.object),
+      id: PropTypes.string.isRequired,
+      title: PropTypes.string.isRequired,
+    })
+  ).isRequired,
+};
+
+ContentsList.defaultProps = {
+  className: '',
```
i80and: What if the depth isn't specified? (According to the spec the default should be ∞)

src/components/Contents.js
```diff
@@ -0,0 +1,103 @@
...
+const ContentsListItem = ({ id, listChildren, title }) => (
+  <li>
+    <a href={`#${id}`}>{title}</a>
+    {listChildren.length > 0 && <ContentsList listItems={listChildren} />}
+  </li>
+);
+
+ContentsListItem.propTypes = {
+  id: PropTypes.string.isRequired,
+  listChildren: PropTypes.arrayOf(PropTypes.object).isRequired,
```
i80and: Are the children nodes? If so, is it possible to define their shape a little more precisely as a constant somewhere?

i80and: Are the children nodes? If so, is it possible to define their shape a little more precisely as a constant somewhere?

PR #105, opened 2019-08-15
DOCSP-6503: Refactor tabs

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-6503)] [[Compass Staging](https://docs-mongodbcom-staging.corp.mongodb.com/compass/sophstad/DOCSP-6503/)]
- Refactor tabs to use new `options.tabid` property
- Support `tabs-pillstrip` directive
- Update tests


src/components/Tabs.js
```diff
@@ -101,35 +108,36 @@ export default class Tabs extends Component {
...
                 </li>
               );
             })}
           </ul>
         )}
         {tabs.map((tab, index) => {
-          let tabName = getNestedValue(['argument', 0, 'value'], tab);
-          if (tabName) tabName = tabName.toLowerCase();
+          const tabId = getNestedValue(['options', 'tabid'], tab);
+          // let tabName = getNestedValue(['argument', 0, 'value'], tab);
```
i80and: Why are these commented out? It's generally better to just not include commented out lines; you can always get them back in the git history.

src/components/Tabs.js
```diff
@@ -101,35 +108,36 @@ export default class Tabs extends Component {
...
+                  {tabTitle}
                 </li>
               );
             })}
           </ul>
         )}
         {tabs.map((tab, index) => {
-          let tabName = getNestedValue(['argument', 0, 'value'], tab);
-          if (tabName) tabName = tabName.toLowerCase();
+          const tabId = getNestedValue(['options', 'tabid'], tab);
```
i80and: If tabId isn't defined, does the propTypes checking log a useful message?

src/templates/guide.js
```diff
@@ -78,31 +77,33 @@ export default class Guide extends Component {
   };
 
   addGuidesTabset = (tabsetName, tabData) => {
-    const { setActiveTab } = this.context;
-    let tabs = tabData.map(tab => tab.argument[0].value);
+    let tabs = tabData.map(tab => getNestedValue(['options', 'tabid'], tab));
```
i80and: should this be just `let tabs`, and then the default assignment in an `else` clause of the following block?

src/components/Tabs.js
sophstad: Oops my mistake! Thanks for catching.

src/components/Tabs.js
sophstad: Yes, it would say
```
Warning: Failed prop type: The prop `nodeData.children[0].tabid` is marked as required in `Tabs`, but its value is `undefined`.
```

src/templates/guide.js
sophstad: No, `tabs` is filtered in each clause to make sure that only valid tabs are included in non-anonymous tabsets. Then `tabs` is reassigned to equal the filtered value. Would it perhaps be more clear if `tabs` were renamed to be a different value (e.g. `tabsFiltered`) on lines 82 and 85? Like so:
```js
const tabsFiltered = DEPLOYMENTS.filter(tab => tabs.includes(tab));
```

i80and: Oh my bad, misread the logic. Yup, I think I might prefer that; that would also allow `tabs` to be const, which makes me happy.

PR #106, opened 2019-08-16
DOCSP-6575: Add support for image caption and legend

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-6575)] [[Compass Staging](https://docs-mongodbcom-staging.corp.mongodb.com/compass/sophstad/DOCSP-6575/export-pipeline-to-language#select-your-desired-export-language)]

### DIFF
--- a/src/components/CaptionLegend.js
+++ b/src/components/CaptionLegend.js
@@ -4,7 +4,13 @@ import ComponentFactory from './ComponentFactory';
 
 const CaptionLegend = ({ nodeData: { children }, ...rest }) => (
   <React.Fragment>
-    {children.length > 0 && <ComponentFactory {...rest} nodeData={children[0]} />}
+    {children.length > 0 && (
+      <p className="caption">
+        <span className="caption-text">
+          <ComponentFactory {...rest} nodeData={children[0]} parentNode="caption" />
+        </span>
+      </p>
+    )}
     {children.length > 1 && (
       <div className="legend">
         {children.slice(1).map((child, index) => (
--- a/src/components/CaptionLegend.js
+++ b/src/components/CaptionLegend.js
@@ -1,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ComponentFactory from './ComponentFactory';
+
+const CaptionLegend = ({ nodeData: { children }, ...rest }) => (
+  <React.Fragment>
+    {children.length > 0 && <ComponentFactory {...rest} nodeData={children[0]} />}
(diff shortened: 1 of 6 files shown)


src/components/CaptionLegend.js
```diff
@@ -0,0 +1,30 @@
...
+    {children.length > 0 && (
+      <p className="caption">
+        <span className="caption-text">
+          <ComponentFactory {...rest} nodeData={children[0]} parentNode="caption" />
+        </span>
+      </p>
+    )}
+    {children.length > 1 && (
+      <div className="legend">
+        {children.slice(1).map((child, index) => (
```
rayangler: For clarification: does this mean that the first child of a Figure is a caption and the second, and onwards, are the arbitrary legend contents?

src/components/Figure.js
```diff
@@ -50,6 +51,7 @@ export default class Figure extends Component {
           onLoad={this.handleImageLoaded}
           ref={this.imgRef}
         />
+        <CaptionLegend {...rest} nodeData={nodeData} />
```
rayangler: Question about how `...rest` works: Based on my understanding, `...rest` is used as a constant/reference to the rest of the props that aren't explicitly being used as constants. But if the props of a component consist of `nodeData`, which we are using as a constant, what is `...rest` referring to exactly?

src/components/CaptionLegend.js
sophstad: Yes! From the [spec](http://docutils.sourceforge.net/docs/ref/rst/directives.html#figure):
> A "figure" consists of image data (including image options), an optional caption (a _**single**_ paragraph), and an optional legend (arbitrary body elements).

src/components/Figure.js
sophstad: Good question! That is exactly what `...rest` refers to. However, the props of a component don't _exclusively_ consist of `nodeData`. There are a few props that are used in a more "global" way—information that can be used by any component that is not explicitly tied to the data being rendered. For example:

- `pageMetadata`: created in `gatsby-node.js`, maps page slugs to their titles (and more!), used for internal link generation
- [`substitutions`](https://github.com/mongodb/snooty/blob/master/src/components/layout.js#L47-L57): writers can define a substitution in one file and use it throughout that page (in its includes as well), so we need to pass the mapping of substitutions to all components

These are just a few examples. We also may need to create more of these global-style props in the future, so using `...rest` ensures that they will be passed to all nodes without us needing to manually add them to a component's list of props.

PR #108, opened 2019-08-20
DOCSP-6189: Single page can be rendered locally

rayangler:
[Jira ticket](https://jira.mongodb.org/browse/DOCSP-6189)
Staging links to make sure implementing local preview doesn't ruin anything:
[Guides staging](https://docs-mongodbcom-staging.corp.mongodb.com/guides/raymundrodriguez/DOCSP-6189/)
[Bi-connector staging](https://docs-mongodbcom-staging.corp.mongodb.com/bi-connector/raymundrodriguez/DOCSP-6189/)

**For testing locally, go to your command line and execute:**
`npm run preview -- --env.PREVIEW_PAGE="<page-you-want-to-render>"`
Note: Your `.env.development` file should be set up the way you would when using `gatsby develop`.

**Example to get to the landing page:**
`npm run preview -- --env.PREVIEW_PAGE="index"`
**Example to get to https://docs.mongodb.com/guides/server/drivers/:**
`npm run preview -- --env.PREVIEW_PAGE="server/drivers"`

This allows the user to render a single page without having to build the entire website. Since it only builds one page, navigating to a separate page does not work properly. To view a different page, you must exit the process and run the above command with the page of your choice.

Other notes:
* Breadcrumbs, which use `gatsby` to work, do NOT work properly in preview. By default, the breadcrumb will be returned as "Breadcrumb" rather than its actual name.


preview-start.js
```diff
@@ -0,0 +1,62 @@
...
+            pageData: null,
+            template: null
+        }
+    }
+
+    componentDidMount() {
+        getPageData(process.env.PREVIEW_PAGE).then((pageData) => {
+            let templateComponent;
+            if (pageData.template === 'guide') {
+                templateComponent = <Guide pageContext={pageData.context} path={pageData.path}/>
```
sophstad: This looks like a great opportunity to use [dynamic component names](https://medium.com/@Carmichaelize/dynamic-tag-names-in-react-and-jsx-17e366a684e9) to make things a little cleaner. You can see an example in Snooty in our [List component](https://github.com/mongodb/snooty/blob/master/src/components/List.js#L7).

preview-start.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    render() {
+        const { pageData, template } = this.state;
+
+        return(
+            <React.Fragment>
+                <Helmet>
+                    {process.env.GATSBY_SITE === 'guides' ? (
+                        <link rel="stylesheet" href='./public/docs-tools/guides.css' type="text/css" />
```
sophstad: I think this ternary violates the DRY principle a little too strongly. If we need to introduce more complex logic for CSS, we'll have to update that in two places. What if we moved this `Helmet` section (and the same section in html.js) into our `DefaultLayout` component? It's possible that this causes a [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content), but worth a shot!

preview-start.js
```diff
@@ -0,0 +1,62 @@
...
+                    ) : (
+                        <link rel="stylesheet" href='./public/docs-tools/mongodb-docs.css' type="text/css" />
+                    )}
+                </Helmet>
+                {pageData && 
+                <DefaultLayout pageContext={pageData.context} path={pageData.path}>
```
sophstad: Since we pass in `pageContext` and `path` here, I don't think we need to pass it in to each individual template component above. DRY FTW!

src/components/Figure.js
```diff
@@ -4,21 +4,48 @@ import PropTypes from 'prop-types';
 import CaptionLegend from './CaptionLegend';
 import Lightbox from './Lightbox';
 import { getNestedValue } from '../utils/get-nested-value';
+// import { getAssetData } from '../../preview/preview-setup';
 
 export default class Figure extends Component {
   constructor(props) {
     super(props);
     this.imgRef = React.createRef();
+    this._isMounted = false; // Can't use this.isMounted
```
sophstad: Why not link to the React docs' statement on this in your comment? (assuming you were referring to this: https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html)

src/components/Figure.js
```diff
@@ -4,21 +4,48 @@ import PropTypes from 'prop-types';
 import CaptionLegend from './CaptionLegend';
 import Lightbox from './Lightbox';
 import { getNestedValue } from '../utils/get-nested-value';
+// import { getAssetData } from '../../preview/preview-setup';
```
sophstad: This comment should probably be deleted!

preview/preview-setup.js
```diff
@@ -0,0 +1,167 @@
...
+const INCLUDE_FILES = {};
+const PAGE_TITLE_MAP = {};
+
+// in-memory object with key/value = filename/document
+let RESOLVED_REF_DOC_MAPPING = {};
+
+// stich client connection
+let stitchClient;
+
+const setupStitch = () => {
```
sophstad: I know this is a lot of work, but it appears that all of these functions (except the base 64 stuff) are identical to those in `gatsby-node.js`. To avoid repeating ourselves and simplify making changes downstream, could we isolate these functions and import them for use in each file? I'm thinking something like `utils/setup/setup-stitch.js`. Luckily, neither `gatsby-node.js` nor `preview-setup.js` accept ES6 imports so we won't have to worry about that annoying syntax issue!

preview/get-include-file.js
```diff
@@ -0,0 +1,15 @@
+const { getNestedValue } = require('../src/utils/get-nested-value');
+
+// TODO: Figure out a way to use the same getIncludeFile() in both gatsby and preview
```
sophstad: Can you mention in this comment that the issue has to do with usage of ES6 import/export vs Node.js require?

preview/preview-setup.js
```diff
@@ -0,0 +1,167 @@
...
+        }
+      };
+    }
+    return null;
+}
+
+// Use checksum from a Figure component to return base64 data of image
+export const getBase64Uri = async (checksum) => {
+  const query = {_id: {$eq: checksum}}
+  const assets = await stitchClient.callFunction('fetchDocuments', [DB, ASSETS_COLLECTION, query]);
```
sophstad: Some neat syntactic sugar:
```js
const [assetData,] = await stitchClient.callFunction('fetchDocuments', [DB, ASSETS_COLLECTION, query]);
```

src/components/Navbar.js
```diff
@@ -26,13 +26,11 @@ export default class Navbar extends Component {
 
   componentDidMount() {
     // Add script to give navbar functionality and css
-    if (document !== undefined) {
```
sophstad: Why do we no longer need this check? Just wondering!

preview/noop.js
```diff
@@ -0,0 +1,15 @@
+// Placeholder functions for preview mode to work without crashing from gatsby functions
+function useSiteMetadata() {
+    return {
+        title: "Breadcrumb"
+    };
+}
+
+function withPrefix(str) {
+    return "#";
```
sophstad: Are you sure that this `#` doesn't cause any problems? I'm wondering if internal links ([implemented here](https://github.com/mongodb/snooty/blob/master/src/components/Roles/Doc.js)) would be affected. What about returning `''` instead?

src/components/Figure.js
rayangler: I actually didn't know about this blog post! I was just trying to see if I can get rid of the potential memory leak using a boolean and ran into the issue with naming it. Good find!

preview/preview-setup.js
rayangler: Pretty cool, thanks for letting me know I can do this!

src/components/Navbar.js
rayangler: The if statement may be unnecessary because of the way `componentDidMount()` works. `componentDidMount()` is called when the component is mounted into the DOM, meaning access to `document` should already be defined. From the React documentation:

> componentDidMount() is invoked immediately after a component is mounted (inserted into the tree). Initialization that requires DOM nodes should go here.

Reference: https://reactjs.org/docs/react-component.html#componentdidmount

preview/noop.js
rayangler: Returning `''` causes the same page to re-render upon clicking the link. Returning `#` causes the browser to return to the top of the page. 

Since the pages these internal links refer to aren't necessarily available in preview, perhaps a good compromise for the user would be to return `#!`? Clicking on the internal links will have no effect and will prevent from moving/refreshing the page unnecessarily

sophstad: I like that! 👍

PR #109, opened 2019-09-06
DOCSP-6528: Add deprecated, versionchanged, versionadded handling

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-6528)] [[BI Connector Staging](https://docs-mongodbcom-staging.corp.mongodb.com/bi-connector/sophstad/DOCSP-6528/reference/mongosqld)]

### DIFF
--- a/src/components/VersionModified.js
+++ b/src/components/VersionModified.js
@@ -1,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ComponentFactory from './ComponentFactory';
+import { getNestedValue } from '../utils/get-nested-value';
+
+const VersionModified = ({ introText, nodeData, ...rest }) => {
+  const version = getNestedValue(['argument', 0, 'value'], nodeData);
+  return (
+    <div className={nodeData.name}>
+      <p>
+        <span className="versionmodified">
+          {introText} {version}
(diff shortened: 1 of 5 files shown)


src/components/VersionModified.js
```diff
@@ -0,0 +1,38 @@
...
+const VersionModified = ({ introText, nodeData, ...rest }) => {
+  const version = getNestedValue(['argument', 0, 'value'], nodeData);
+  return (
+    <div className={nodeData.name}>
+      <p>
+        <span className="versionmodified">
+          {introText} {version}
+          {nodeData.children.length > 0 || nodeData.argument.length > 1 ? ': ' : '.'}
+        </span>
+        {nodeData.argument.length > 1 &&
```
rayangler: This might have more to do with how the parser works but I'm curious about an example of when the number of arguments are > 1

sophstad: These directives are unique because they have 1 required argument and one optional argument (as per the [documentation](https://www.sphinx-doc.org/en/1.5/markup/para.html)). They definitely did require special handling—here is the PR where these were implemented in the parser: https://github.com/mongodb/snooty-parser/pull/24.

PR #115, opened 2019-09-25
DOCSP-6902: Fix image includes

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-6902)] [[Manual Staging](https://docs-mongodbcom-staging.corp.mongodb.com/manual/sophstad/DOCSP-6902/core/index-multikey)]
- Fix images added by the `include` directive
- Clean up miscellaneous code in `gatsby-node.js`

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -1,11 +1,10 @@
 const path = require('path');
 const fs = require('fs').promises;
 const mkdirp = require('mkdirp');
-const { validateEnvVariables } = require('./src/utils/setup/validate-env-variables');
 const { Stitch, AnonymousCredential } = require('mongodb-stitch-server-sdk');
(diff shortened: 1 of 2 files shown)


gatsby-node.js
```diff
@@ -87,6 +92,17 @@ const populateIncludeNodes = nodes => {
   return nodes.map(replaceInclude);
 };
 
+// Get various metadata for a given page
+const getPageMetadata = pageNode => {
+  const children = getNestedValue(['ast', 'children'], pageNode);
```
danielborowski: You think it'd be better if `getIncludeFile` took in an object instead of 2 params. Then within that object you had `filename` property and `includeObj` property or something similar?

danielborowski:
otherwise LGTM

PR #118, opened 2019-10-14
DOCSP-7186: Update preview files to support all templates

rayangler:
[JIRA Ticket](https://jira.mongodb.org/browse/DOCSP-7186)
[Compass Staging](https://docs-mongodborg-staging.corp.mongodb.com/compass/raymundrodriguez/DOCSP-7186/)

This PR introduces changes to the preview files on the frontend. These changes will allow the VS Code extension to preview pages that use templates other than just `Guide`.

`process.env.GATSBY_SITE` is reimplemented into the webpack config but instead of obtaining it from `.env.development`, it is passed as a variable from within the Snooty extension.

Note: Since the VS Code extension requires these frontend changes to present in the submodule, this PR needs to be pushed before changes to the extension are pushed. (Extension's branch can be found [here](https://github.com/mongodb/snooty-vscode/tree/DOCSP-7186))

### DIFF
--- a/preview/preview-setup.js
+++ b/preview/preview-setup.js
@@ -3,13 +3,20 @@ const pageAST = require('./page-ast.json'); // File created by Snooty extension 
 
 // Returns bare minimum data needed by a single page
 export const getPageData = async () => {
+  const fileId = process.env.PREVIEW_PAGE;
   const pageNodes = {
     ast: pageAST,
   };
 
+  // Pick which template to use
+  let template = 'document';
+  if (process.env.GATSBY_SITE === 'guides') {
+    template = fileId === 'index' ? 'guides-index' : 'guide';
+  }
+
   return {
     path: '',
-    template: 'guide',
+    template,
     context: {
       snootyStitchId: '',
       __refDocMapping: pageNodes,
--- a/preview/preview-setup.js
+++ b/preview/preview-setup.js
@@ -1,5 +1,6 @@
 // eslint-disable-next-line import/no-unresolved
 const pageAST = require('./page-ast.json'); // File created by Snooty extension on VS Code
+const { getTemplate } = require('../src/utils/get-template');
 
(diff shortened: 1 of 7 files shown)


webpack.config.js
```diff
@@ -55,6 +55,7 @@ module.exports = env => {
       // TODO: Pass in an environment variable for process.env.GATSBY_SITE
```
sophstad: We probably want to get rid of this TODO!
```suggestion
```

preview/preview-setup.js
```diff
@@ -3,13 +3,20 @@ const pageAST = require('./page-ast.json'); // File created by Snooty extension
...
+  const fileId = process.env.PREVIEW_PAGE;
   const pageNodes = {
     ast: pageAST,
   };
 
+  // Pick which template to use
+  let template = 'document';
+  if (process.env.GATSBY_SITE === 'guides') {
+    template = fileId === 'index' ? 'guides-index' : 'guide';
+  }
```
sophstad: I notice the [same code](https://github.com/mongodb/snooty/blob/master/gatsby-node.js#L176-L179) is used in `gatsby-node.js`. Any chance we can break this out into a util function? I worry that as we introduce more complex logic (e.g. around landing pages), we will have trouble keeping things straight.

src/utils/get-template.js
```diff
@@ -0,0 +1,10 @@
+// Returns the name of the template to be used based on the site and page name.
+const getTemplate = page => {
+  let template = 'document';
+  if (process.env.GATSBY_SITE === 'guides') {
```
sophstad: As a best practice, we should keep these util functions self-contained (i.e. not environment-dependent). Can we pass in `site` as a parameter along with `page`? Sorry to nit! 😄

rayangler: Nothing to be sorry about; thanks for bringing it to my attention!

PR #125, opened 2019-11-13
DOCSP-7498: Implement breadcrumbs component

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-7498)] [[Ecosystem Staging](https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/sophstad/DOCSP-7498/drivers/pymongo)]
- Fetch breadcrumb data from Atlas and renders React component
- Design Change: No longer show the current page as the final breadcrumb element
image


src/components/Breadcrumbs.js
```diff
@@ -0,0 +1,32 @@
...
+import { Link } from 'gatsby';
+import { getNestedValue } from '../utils/get-nested-value';
+
+const Breadcrumbs = ({ parentPaths, slugTitleMapping }) => (
+  <div className="bc">
+    {parentPaths && (
+      <ul>
+        {parentPaths.map((path, index) => (
+          <li key={path}>
+            <Link to={`/${path}`}>{getNestedValue([path], slugTitleMapping)}</Link>
```
rayangler: I'm having trouble rendering pages with more than one slug on `gatsby develop` (`drivers/pymongo` or `drivers/c` won't render, but `drivers/` works fine). The following error received:
```
Objects are not valid as a React child (found: object with keys {type, position, value}). If you meant to render a collection of children, use an array instead.
```

However, it works when using:
```
<Link to={`/${path}`}>{getNestedValue([path, 0, 'value'], slugTitleMapping)}</Link>
```
I was wondering if you had this same issue on `develop` as well or if it's just me

PR #135, opened 2019-12-04
Bugfix: Only render .txt files as pages

sophstad:
Only files with the extension `.txt` should be rendered as pages. Fix this oversight.

Additionally, improve safety of a string modification utility function.


danielborowski:
lgtm

PR #139, opened 2020-01-02
DOCSP-8134: Fix tabset rendering in ecosystem

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-8134)] [[Ecosystem Staging](https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/ubuntu/master/use-cases/client-side-field-level-encryption-guide)] [[Guides Staging](https://docs-mongodbcom-staging.corp.mongodb.com/guides/ubuntu/master/)] Fixed bug where `drivers` tabsets were not rendering in Ecosystem docs.

In Guides, certain tabsets are rendered as pills at the top of the page, not inline. This PR adds a check of which site is being built to determine whether to render tabs or pills.

Eventually, we will add better directives for defining pillsets.

### DIFF
--- a/src/components/Tabs.js
+++ b/src/components/Tabs.js
@@ -82,8 +82,10 @@ export default class Tabs extends Component {
     const { tabsetName } = this.state;
     const { nodeData, pillstrips } = this.props;
     const { activeTabs, setActiveTab } = this.context;
+    // Certain tabsets are rendered at the top of the page, rather than inline, in Guides
     const isHeaderTabset =
-      tabsetName === 'drivers' || tabsetName === 'cloud' || Object.keys(pillstrips).includes(tabsetName);
+      process.env.GATSBY_SITE === 'guides' &&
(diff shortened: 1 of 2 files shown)


src/components/Tabs.js
```diff
@@ -82,8 +82,10 @@ export default class Tabs extends Component {
     const { tabsetName } = this.state;
     const { nodeData, pillstrips } = this.props;
     const { activeTabs, setActiveTab } = this.context;
+    // Certain tabsets are rendered at the top of the page, rather than inline, in Guides
```
i80and: Either of the following causes tabsets to be rendered inline:

* A preceding `.. tabs-top` directive
* An anonymous tabset

Otherwise, it is rendered at the top of the page.

I know this is a Quick Fix for guides, but we should mention that the logic is more complex than this I think.

sophstad: To clarify, is that only the case for guides?

i80and: That's the case for all properties, I believe

PR #141, opened 2020-01-14
Fix preview mode

sophstad:
Fix preview mode bugs:
- Fix Babel
  - Install missing dependencies
  - Reintroduce config in `webpack.config.js`
- Don't render components that require sitewide information in non-guides sites (TOC, Breadcrumbs, internal page nav)
  - `DocumentBody` is now its own component


danielborowski:
lgtm

PR #148, opened 2020-02-03
Add text components, update storybook styling

jestapinski:
This PR:

- Adds text components (H1, etc) and adds fonts from mockup to global styles in layouts
- Uses ^ on existing layout pages
- Updates storybook styling and adds text components to it

Screen Shot 2020-02-03 at 10 54 20 AM
Screen Shot 2020-02-03 at 10 54 41 AM

### DIFF
--- a/src/components/dev-hub/layout.js
+++ b/src/components/dev-hub/layout.js
@@ -8,6 +8,15 @@ const globalStyles = css`
   body {
     background: ${colorMap.devBlack};
     color: ${colorMap.devWhite};
+    font-family: 'Fira Mono Bold', 'Space Mono Bold', 'Source Code Pro';
+  }
+  /* Need overrides for guides.css styles */
+  h1,
+  h2,
+  h3,
+  h4 {
+    border: none;
+    font-family: 'Fira Mono Bold', 'Space Mono Bold', 'Source Code Pro';
   }
 `;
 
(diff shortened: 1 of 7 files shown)


src/components/dev-hub/text/text.js
```diff
@@ -0,0 +1,45 @@
...
+import styled from '@emotion/styled';
+import { fontSize, size } from '../theme';
+
+/**
+ *
+ * @param {Object.string} param
+ *  @param {string} italic should be italic style
+ *  @param {string} bold should be bold weight
+ */
+const variants = ({ bold, collapse }) =>
```
jestapinski: We might also want to pull this out for the time being, what do you think?

src/components/dev-hub/layout.js
```diff
@@ -8,6 +8,15 @@ const globalStyles = css`
   body {
     background: ${colorMap.devBlack};
     color: ${colorMap.devWhite};
+    font-family: 'Fira Mono Bold', 'Space Mono Bold', 'Source Code Pro';
```
jestapinski: I'm not sure what all of the rules for when these are used vs Akzidenz, so I stuck with just these for now and when the mockups are finalized we can update this.

src/components/dev-hub/layout.js
```diff
@@ -8,6 +8,15 @@ const globalStyles = css`
...
+  }
+  /* Need overrides for guides.css styles */
```
greggb: We'll need to update the GATSBY_SITE name to get rid of that stylesheet. Might also need to work with Sophie to fix the ternary that will load an alternate stylesheet in `html.js`

src/components/dev-hub/text/text.js
greggb: fine with me

src/components/dev-hub/layout.js
jestapinski: Noted

PR #149, opened 2020-02-03
Add content label

jestapinski:
This PR adds the "content label" (badge). I went with the top variation while the design team finishes deciding what they want to do.

I wanted this to be pretty generic to support multiple types of content labels. These color/word associations don't seem finalized yet either.

Also added to storybook.

Screen Shot 2020-02-03 at 11 30 16 AM


src/components/dev-hub/contentLabel/content-label.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import styled from '@emotion/styled';
+import { P } from '../text';
+import { colorMap } from '../theme';
+
```
greggb: I think we're going to need a map in here so we can just do something like `<ContentLabel>Quick Start</ContentLabel>` and it uses the right colors automatically.

src/components/dev-hub/contentLabel/content-label.js
```diff
@@ -0,0 +1,22 @@
...
+/* TODO: Update font once decided */
+const ContentLabelText = styled(P)`
```
greggb: Could we call them `Badge`?

src/components/dev-hub/contentLabel/content-label.js
```diff
@@ -0,0 +1,22 @@
...
+  background: ${colorMap.greyDarkThree};
+  font-size: 10px;
```
greggb: should we adjust the font and size maps in theme for the devhub to match common values?

src/components/dev-hub/contentLabel/content-label.js
```diff
@@ -0,0 +1,22 @@
...
+  font-weight: bold;
+  letter-spacing: 1px;
+  padding: 5px 10px;
+  ${({ textColor }) => textColor && `color: ${textColor};`}
+  text-transform: uppercase;
+  width: fit-content;
+`;
+
+const ContentLabel = ({ text, textColor }) => {
```
greggb: Maybe just `color` for the prop? Since it'll be black and some color, but could also be the border or BG

greggb: Let's pass the text as children

src/components/dev-hub/theme.js
```diff
@@ -3,7 +3,7 @@
  * @property {string} fontsize returns px value
  */
 const fontSize = {
-  xsmall: '11px',
+  xsmall: '10px',
```
jestapinski: I have a feeling these values are still in flux, so we can revisit these down the road a bit more as well.

src/components/dev-hub/badge/badge.js
```diff
@@ -0,0 +1,29 @@
...
+/* TODO: Update font once decided */
+const BadgeText = styled(P)`
+  background: ${colorMap.greyDarkThree};
+  font-size: ${fontSize.xsmall};
+  font-weight: bold;
+  letter-spacing: 1px;
+  padding: ${size.tiny} ${size.small};
+  ${({ textColor }) => textColor && `color: ${textColor};`}
+  text-transform: uppercase;
+  width: fit-content;
```
greggb: This is cool. Does it need the value arg? What happens in the Edge browser?

src/components/dev-hub/badge/badge.js
```diff
@@ -0,0 +1,29 @@
...
+const badgeTypeColorMap = {
+  Article: colorMap.salmon,
+  Community: colorMap.magenta,
+  'Deep Dive': colorMap.teal,
+  Event: colorMap.orange,
+  'How-To': colorMap.yellow,
+  'Quick Start': colorMap.lightGreen,
+};
+
+const Badge = ({ children }) => <BadgeText textColor={badgeTypeColorMap[children]}>{children}</BadgeText>;
```
greggb: Nice!

src/components/dev-hub/badge/badge.js
jestapinski: Looking fine on edge, it uses the `max-content` size of the child unless there is less than that amount of space. This is fine here, because no reasonable width would result in these being cut down (Quick Start is ~100px)

Screen Shot 2020-02-04 at 10 59 32 AM

jestapinski: (the value arg is optional)

jestapinski:
Thanks!

PR #151, opened 2020-02-03
Add link component (tertiary button)

jestapinski:
I brought the `Link` over from MITX and made some simplifications since this looks to have just one style for now. Colors could be in flux while design decides, but the functionality should be there.

The `link` seems like the tertiary button as well, but I figured a `button tertiary` would just use `link` in the future.

Normal state:
Screen Shot 2020-02-03 at 4 17 53 PM

Hover state:
Screen Shot 2020-02-03 at 4 17 48 PM


jestapinski:
Thanks!

PR #155, opened 2020-02-12
DOCSP-8882: Add release workflow

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOCSP-8882)] General release workflow:
1. Call `npm version <major/minor/patch>` to kick off [`npm-version`](https://docs.npmjs.com/cli/version). Three scripts are then called:
    1. `preversion`: ensures master branch, runs standard lint/format/unit tests
    2. `version`: auto-generates changelog using git history and [`auto-changelog`](https://www.npmjs.com/package/auto-changelog). `npm-version` then adds a commit here.
    3. `postversion`: push with tags
2. GitHub Action invoked: release is created, then an artifact is built and attached to said release.

I also did some simplification of our git submodule. Right now `snooty.zip` is around 58M. So close! Also, do let me know if you know of any approach for testing the GitHub Action. Thanks!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,10 @@ jobs:
         run: |
           npm install --production
           zip -9r snooty.zip . -x "*.git*" "*.cache*" public/\* docs-tools/\*
+      - name: Get environment
+        id: environment
+        run: |
+          echo "::set-output name=date::$(date +%Y-%m-%d)"
       - name: Create Release
         id: create_release
         uses: actions/create-release@v1
@@ -23,7 +27,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           tag_name: ${{ github.ref }}
-          release_name: Release ${{ github.ref }}
+          release_name: "Release: [${{ github.ref }}] - ${{ steps.environment.outputs.date }}"
           draft: true
           prerelease: true
       - name: Upload Release Asset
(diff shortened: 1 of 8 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,42 @@
...
+  build:
+    name: Upload Release Asset
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Build project
+        run: |
+          npm install --production
+          zip -9r snooty.zip . -x "*.git*" "*.cache*" public/\* docs-tools/\*
```
i80and: I think it would be better to list the files and directories specifically we want to include: `zip` supports reading the files from stdin, so we can cat a list of paths into it.

But that doesn't need to be done right this moment

i80and: Also it would be nice to have this be a script so we can easily run a command to build a package for testing --- ideally, a `package` target in the Makefile for symmetry with the parser.

sophstad: Should things like tests and lint configuration be left out of the zip?

sophstad: Also, should the zip be constructed specifically with preview mode in mind? E.g. all `gatsby-*.js` files should be ignored?

i80and: Yup. This is a prod release, so as little as possible

sophstad: Gotcha, in that case, I think I'll also rename the zipfile to `snooty-preview.zip` to help clarify!

PR #160, opened 2020-02-28
DOCSPLAT-141: Add navigation tab functionality

mango-db:
~[Staged `ecosystem` docs](https://docs-mongodbcom-staging.corp.mongodb.com/master/ecosystem/ubuntu/DOCSPLAT-141/).~


mango-db:
[[JIRA](https://jira.mongodb.org/browse/DOCSPLAT-141)] [[Ecosystem staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/ecosystem/claire/DOCSPLAT-141/)] [[Tools staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/atlas-open-service-broker/claire/DOCSPLAT-141/)]

- Add navigation tab functionality to open and close the sidebar
- Use initial window size to determine state of navigation tab (open or closed)

### Testing Notes

- Tested on Ecosystem because it has a sidebar
- Tested on Tools because it has a sidebar

src/templates/document.js
```diff
@@ -22,20 +24,40 @@ const Document = props => {
...
   return (
     <React.Fragment>
       <Helmet>
         <title>{title}</title>
       </Helmet>
       <Navbar />
       <div className="content">
-        <div id="left-column">
-          <Sidebar slug={slug} publishedBranches={publishedBranches} toctreeData={toctree} />
+        <div>
```
sophstad: Though it solves our problem, this `<div>` has no semantic meaning. We could avoid adding an extra DOM node by using a [`<React.Fragment>`](https://reactjs.org/docs/fragments.html)—as long as it doesn't re-introduce the bug!

src/hooks/get-window-size.js
```diff
@@ -0,0 +1,30 @@
+import React, { useState, useEffect } from 'react';
+
+export const GetWindowSize = () => {
```
sophstad: [nit] Capitalization in React is used to indicate a component. Since this is a hook, let's name it `getWindowSize()`.

src/hooks/get-window-size.js
```diff
@@ -0,0 +1,30 @@
...
+  const isClient = typeof window === 'object';
```
sophstad: We have an [`isBrowser()`](https://github.com/mongodb/snooty/blob/master/src/utils/is-browser.js) util function that does the same thing! You can use that instead.

(Also `isBrowser()` should probably be converted to a variable at some point...)

src/hooks/get-window-size.js
```diff
@@ -0,0 +1,30 @@
...
+  const getSize = React.useCallback(() => {
+    return {
+      width: isClient ? window.innerWidth : undefined,
+      height: isClient ? window.innerHeight : undefined,
+    };
+  }, [isClient]);
+
+  const [windowSize, setWindowSize] = useState(getSize);
+
+  useEffect(() => {
```
sophstad: Did you try using [`useLayoutEffect`](https://reactjs.org/docs/hooks-reference.html#uselayouteffect) instead here? Wondering if could help prevent the brief moment when the sidebar is still hidden upon load.

src/hooks/get-window-size.js
```diff
@@ -0,0 +1,30 @@
...
+    if (!isClient) {
+      return false;
+    }
+
+    function handleResize() {
```
sophstad: [nit] Arrow function preferred!

src/components/Sidebar.js
```diff
@@ -3,13 +3,17 @@ import Link from './Link';
...
-const Sidebar = ({ slug, publishedBranches, toctreeData }) => {
+const Sidebar = ({ slug, publishedBranches, toctreeData, toggleLeftColumn }) => {
   const { title } = toctreeData;
   return (
-    <aside className="sidebar" id="sidebar">
-      <div className="sphinxsidebar" id="sphinxsidebar">
+    <aside className={`sidebar ${style.sidebar}`} id="sidebar">
+      <div className={`sphinxsidebar ${style.sphinxsidebar}`} id="sphinxsidebar">
         <div id="sphinxsidebarwrapper" className="sphinxsidebarwrapper">
+          <span className={`closeNav`} id="closeNav" onClick={toggleLeftColumn} style={{ cursor: 'pointer' }}>
```
sophstad: [nit] Tiny change for the sake of consistency!
```suggestion
          <span className="closeNav" id="closeNav" onClick={toggleLeftColumn} style={{ cursor: 'pointer' }}>
```

src/templates/document.js
mango-db: Unfortunately, wrapping in `<React.Fragment>` instead of `<div>` doesn't solve the bug!

PR #170, opened 2020-03-31
DOP-906: Update VersionDropdown React component to support EOL versions implementation

tahiya-mongo:
Jira: https://jira.mongodb.org/browse/DOP-906

Staging: https://docs-mongodborg-staging.corp.mongodb.com/master/manual/tahiya.chowdhury/DOP-906/


src/components/VersionDropdown.js
```diff
@@ -56,6 +56,22 @@ const VersionDropdown = ({
   const wrapperRef = useRef(null);
   useOutsideHandler(wrapperRef);
 
+  // Handle cases where EOL versions exist
+  const legacyDocsURL = normalizePath(
+    `${generatePathPrefix({ ...siteMetadata, parserBranch: 'legacy-docs' })}/?site=${process.env.GATSBY_SITE}/${slug}`
+  );
```
sophstad: The legacy URL that we're using is actually quite static. It's simply `https://docs.mongodb.com/legacy` with the query string appended. No need to calculate based on path prefix or anything complicated!

tests/unit/VersionDropdown.test.js
```diff
@@ -87,3 +112,30 @@ describe('when rendered', () => {
     });
   });
 });
+
+describe('when rendered', () => {
```
sophstad: [nit] This is a good place to explain that we're rendering a property without legacy docs:
```js
describe('when rendering a property without legacy docs', () => {
```

Above (line 48ish), you can mention that you're testing a property _with_ legacy docs.

PR #176, opened 2020-04-07
Dop 785: Use LeafyGreen Button component for version dropdown

tahiya-mongo:
Jira: https://jira.mongodb.org/browse/DOP-785
Stagining: https://docs-mongodbcom-staging.corp.mongodb.com/master/manual/tahiya.chowdhury/DOP-785/

### DIFF
--- a/src/components/VersionDropdown.js
+++ b/src/components/VersionDropdown.js
@@ -4,6 +4,8 @@ import { useSiteMetadata } from '../hooks/use-site-metadata';
 import { generatePathPrefix } from '../utils/generate-path-prefix';
 import { normalizePath } from '../utils/normalize-path';
 import dropdownStyles from '../styles/version-dropdown.module.css';
+import Button from '@leafygreen-ui/button';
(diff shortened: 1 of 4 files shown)


src/components/VersionDropdown.js
```diff
@@ -72,17 +74,16 @@ const VersionDropdown = ({
...
+        variant="default"
         className="version-button dropdown-toggle"
-        onClick={() => setHidden(!hidden)}
         title="Select version"
+        glyph={<Icon glyph="CaretDown" />}
+        onClick={() => setHidden(!hidden)}
+        size="large"
       >
         {prefixVersion(currentBranch)}
-        <span className="caret"></span>
```
sophstad, line 84: Good thinking on using the `glyph` property supported by LeafyGreen. However, I think we may want to keep this caret implementation so that we can have the icon at the right side of the button. This will help with stylistic consistency between next-gen and legacy builds as we begin the migration. But again, thanks for taking initiative to implement it the "right" way!

PR #185, opened 2020-04-23
FeedbackCard should display behind tooltips

nlarew:
- Removes the unnecessary `z-index` on `<FeedbackCard />` that made `<StarRating />` tooltips display behind the card

[Staged](https://docs-mongodbcom-staging.corp.mongodb.com/master/node/nlarew/feedback-fix/)


sophstad:
Thanks for catching this bug, Nick! However, we added this `z-index` property because LeafyGreen code components were hiding the feedback widget. I'm still noticing this on the staging link:
image

Looks like some more `z-index` massaging may be required to fix this 😓

sophstad:
Addressed in #191

PR #200, opened 2020-06-05
DOP-877: Properly render literal includes

mango-db:
Staging:
- [Docs-realm](https://docs-mongodbcom-staging.corp.mongodb.com/master/realm/claire/DOP-877/)
- [Docs-tutorials](https://docs-mongodbcom-staging.corp.mongodb.com/master/guides/claire/DOP-877/)
- [Docs-node](https://docs-mongodbcom-staging.corp.mongodb.com/master/node/claire/DOP-877/)


src/components/LiteralInclude.js
```diff
@@ -33,7 +22,7 @@ export default class LiteralInclude extends Component {
         {...this.props}
         nodeData={{
           type: 'code',
-          value: this.codeExample.substring(0, this.codeExample.lastIndexOf('\n')).trim(),
+          value: this.codeExample,
```
sophstad, line 25: It looks like the approach here is to:

1. Extract the contents of the node's first child
2. Create a new child node that uses these same contents

How can we simplify this? The component factory pattern we use throughout Snooty can do most of the work for you.

src/components/LiteralInclude.js
```diff
@@ -42,15 +31,7 @@ export default class LiteralInclude extends Component {
...
-      })
-    ),
-    options: PropTypes.shape({
-      'start-after': PropTypes.string.isRequired,
-      'end-before': PropTypes.string.isRequired,
-    }),
+    children: PropTypes.arrayOf(PropTypes.node).isRequired,
   }).isRequired,
-  refDocMapping: PropTypes.objectOf(PropTypes.object).isRequired,
+  codeExample: PropTypes.string.isRequired,
```
sophstad, line 36: The `prop-types` package is only used to validate React props, not class variables, so this doesn't apply.

src/components/LiteralInclude.js
```diff
@@ -1,30 +1,19 @@
 import React, { Component } from 'react';
-import PropTypes from 'prop-types';
+import PropTypes, { node } from 'prop-types';
 import ComponentFactory from './ComponentFactory';
 import { getNestedValue } from '../utils/get-nested-value';
 
 export default class LiteralInclude extends Component {
```
sophstad, line 6: [nit] I think we can convert this to a function component; `LiteralInclude` isn't doing any state management.

src/components/LiteralInclude.js
```diff
@@ -1,30 +1,19 @@
...
   constructor(props) {
     super(props);
 
-    const { nodeData, refDocMapping } = this.props;
+    const { nodeData, _ } = this.props;
```
sophstad, line 10: The underscore for unnamed variables is a strictly Pythonic convention. Plus, [destructuring variables](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) doesn't require enumerating every property present in an object, it handily allows us a way of naming only those that are used.

src/components/LiteralInclude.js
```diff
@@ -1,56 +1,18 @@
...
-
-    this.codeExample =
-      typeof this.resolvedIncludeData === 'string'
-        ? this.resolvedIncludeData.substring(
-            this.resolvedIncludeData.indexOf(startText) + startText.length,
-            this.resolvedIncludeData.indexOf(endText)
-          )
-        : '';
+const LiteralInclude = ({ nodeData: { children }, ...rest }) => {
+  if (children && children.length) {
```
sophstad, line 6: I think the simplest way to handle a literal include would be to iterate through all members of `children` and render them. You may have noticed in the AST that `children` is always present, even when it is empty, so we don't have to worry about this approach throwing an error. The benefits of this are:
- No assumptions made about what kind/how many nodes are present in `children`
- Handle cases where `literalinclude` directives have multiple children (future-proof)
- No conditionals (cleaner component)

PR #202, opened 2020-06-09
upgrade html-screen-capture-js to 1.0.43

urikalish:
Since html-screen-capture-js@1.0.31 (the version currently used by Snooty), we added support for more types of inputs (e.g. checkboxes, textareas, etc.), handled resized images, and fixed some minor defects. I don't have Snooty working locally, so I'm unable to test this, but html-screen-capture-js doesn't break compatibility between versions, so hopefully, this should go smoothly. It would be great if some Mongo Ninja can test and merge this upgrade.


urikalish:
newer version exists

PR #204, opened 2020-06-09
DOP-889: Render glossary directive

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOP-889)] [[Docs Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/manual/sophstad/DOP-889/reference/glossary)] Note that there is still a discrepancy in the glossary's target IDs output by the parser ([JIRA](https://jira.mongodb.org/browse/DOP-1171)), so linking to a next-gen glossary from legacy tooling would not be expected to work.


src/components/Glossary.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ComponentFactory from './ComponentFactory';
+
+const Glossary = ({ nodeData: { children }, ...rest }) => (
+  <>
```
i80and, line 6: I'm just curious: what does this do?

PR #221, opened 2020-07-07
DOP-1219 Add leafygreen text input

jestapinski:
This PR simply adds the [Leafygreen Text Input](https://github.com/mongodb/leafygreen-ui/tree/master/packages/text-input) package as a dependency. It is not in use yet.


sophstad:
Dependency added in #223

PR #223, opened 2020-07-10
[DOP-1219] Searchbar Part 1: Desktop Searchbar UI

jestapinski:
[Staging Link](https://docs-mongodborg-staging.corp.mongodb.com/master/guides/jordanstapinski/searchbar/)

(The target branch is the feature branch for the project, also the addition to package/lock.json has been approved)

This PR adds in a basic searchbar (not mobile-responsive, desktop only) which uses LeafyGreen components. Locally I disabled the navbar in docs-tools, but this will show up even without doing so.

I updated the `tabindex` but also noticed this is a bit odd on Safari so will continue to look into that

Screen Shot 2020-07-10 at 4 39 33 PM

Curious how the CSS style is compared to what the team traditionally does, so feedback there would be helpful!

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "@leafygreen-ui/leafygreen-provider": "^1.0.0",
     "@leafygreen-ui/modal": "^3.0.1",
     "@leafygreen-ui/palette": "^2.0.0",
+    "@leafygreen-ui/text-input": "^2.0.0",
     "@leafygreen-ui/tooltip": "^3.0.0",
     "@loadable/component": "^5.12.0",
     "babel-loader": "^8.0.6",
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "@leafygreen-ui/leafygreen-provider": "^1.0.0",
     "@leafygreen-ui/modal": "^3.0.1",
     "@leafygreen-ui/palette": "^2.0.0",
-    "@leafygreen-ui/text-input": "^2.0.0",
+    "@leafygreen-ui/text-input": "^3.0.0",
     "@leafygreen-ui/tooltip": "^3.0.0",
     "@loadable/component": "^5.12.0",
     "babel-loader": "^8.0.6",
--- a/src/components/Navbar.js
+++ b/src/components/Navbar.js
@@ -78,13 +78,7 @@ export default class Navbar extends Component {
     this.navprops = this.modifyActiveLink();
     return (
       <>
-        <div
-          tabIndex="0"
-          id="navbar"
-          className="navbar"
-          data-navprops={this.navprops}
-          style={{ position: 'absolute' }}
-        />
+        <div id="navbar" className="navbar" data-navprops={this.navprops} tabIndex="0" />
         <Searchbar />
       </>
     );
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -67,6 +67,7 @@
     "@leafygreen-ui/leafygreen-provider": "^1.0.0",
     "@leafygreen-ui/modal": "^3.0.1",
     "@leafygreen-ui/palette": "^2.0.0",
+    "@leafygreen-ui/text-input": "^2.0.0",
```
sophstad, line 70: Version 3.0.0 of this package was [just released](https://github.com/mongodb/leafygreen-ui/blob/master/packages/text-input/CHANGELOG.md), so let's use that!

src/components/Navbar.js
```diff
@@ -75,6 +76,17 @@ export default class Navbar extends Component {
...
-    return <div id="navbar" className="navbar" data-navprops={this.navprops} />;
+    return (
+      <>
+        <div
+          tabIndex="0"
+          id="navbar"
+          className="navbar"
+          data-navprops={this.navprops}
+          style={{ position: 'absolute' }}
+        ></div>
```
sophstad, line 87: [nit] Any reason why this wouldn't just be a `<div />`?

src/components/Searchbar/Searchbar.js
```diff
@@ -0,0 +1,118 @@
+import React, { useCallback, useMemo, useState } from 'react';
+import { css, keyframes } from '@emotion/core';
+import styled from '@emotion/styled';
+import Button from '@leafygreen-ui/button';
+import TextInput from '@leafygreen-ui/text-input';
+import Icon from '@leafygreen-ui/icon';
+import SearchDropdown from './SearchDropdown';
+
+const GO_BUTTON_COLOR = '#E4F4F4';
```
sophstad, line 9: Looks like this is `green.light3` in the LeafyGreen palette package! ([source](https://github.com/mongodb/leafygreen-ui/blob/master/packages/palette/src/uiColors.ts#L19))

package.json
jestapinski: Oh wow literally today haha

src/components/Navbar.js
jestapinski: Nope, I was playing around with the implementation using children but never switched back.

jestapinski:
> This looks good! Are you able to generate a staging link? Should be doable just by running `make stage` in the snooty root after building.
> 
> With regard to CSS, this is pretty similar to the direction we're heading towards. You can check out the `components/landing/` directory for reference, as these are some newer components we worked on. For that work, we started using a theme (`theme/docsTheme.js`) that is largely much the same as UP's theme, so feel free to use that as well. I'd encourage you to add any scaffolding you think would be helpful!
> 
> Also, I think adding unit tests for these components would be helpful. No worries if you were planning to to add that later on!

I think for unit testing I was hoping to add to the existing `Navbar` test now but worry about a new `Searchbar` one once the mobile functionality is added since that might change the rendered content structure. What do you think?

I'll also generate a staging link once these changes are made and update the description accordingly, thanks!

jestapinski:
Got the staging link to work! I didn't update docs-tools for it but will work on locally doing so to remove the existing searchbar from the nav. Not a real issue for now though

jestapinski:
thanks!

PR #227, opened 2020-07-14
Upgrade Leafy Text Input to 3.0.1

jestapinski:
[Changelog](https://github.com/mongodb/leafygreen-ui/blob/master/packages/text-input/CHANGELOG.md)

Fixes the aria-labelledby casing on the TextInput (which I also was noticing in the console). Didn't think this one needed to be staged :)


jestapinski:
💯

PR #228, opened 2020-07-15
[DOP-1219] Searchbar Part 2: Responsiveness

jestapinski:
[Ecosystem Staging Link](https://docs-mongodbcom-staging.corp.mongodb.com/master/ecosystem/jordanstapinski/searchbar-mobile/) (also sent to Elle/Allison for review)
[Guides Staging Link](https://docs-mongodbcom-staging.corp.mongodb.com/master/guides/jordanstapinski/searchbar-mobile/)

(This PR was large enough, so will add unit test updates in a separate PR)

This PR adds in responsive support for the new Searchbar component. Specifically, on smaller screen sizes, we will hide the Searchbar until a magnifying glass icon is clicked. I also cleaned up some of the existing animations that don't need `keyframes` to reduce some clutter.

One note:
- `expanded` refers to the functionality at the lower breakpoints where we have the magnifying glass which, on click, opens the full navbar. I wanted to control this in `Navbar` to not only abstract control out of the `Searchbar` but also provide the opacity effect.


src/hooks/useScreenSize.js
```diff
@@ -3,13 +3,17 @@ import { useMediaQuery } from 'react-responsive';
 export default function useScreenSize() {
   const isSmallScreen = useMediaQuery({ query: '(max-width: 750px)' });
   const isTabletOrMobile = useMediaQuery({ query: '(max-width: 1224px)' });
+  const isMediumScreen = useMediaQuery({ query: 'only screen and (max-width: 767px)' });
```
jestapinski, line 6: I saw this file for accessing media queries in JS and was wondering if this was the best place to add new queries (based on the new theme) to access via a hook. Alternatively we could just `useMediaQuery` where needed. What do you think?

sophstad: Yeah good catch! I like the idea of `useScreenSize` (and using the theme variables), so if you think it makes sense then I'm all for it!

jestapinski:
> When testing for responsiveness in Safari, the search bar appears cut off. Are you able to reproduce?
> 
> image

Oh interesting, at that breakpoint I get the mobile searchbar we are going for. Did you do anything specific and then resize? Curious if maybe it just had not rerendered for some reason.

This is what I see:
Screen Shot 2020-07-15 at 5 24 40 PM

jestapinski:
Here is the screenshot without the dropdown on:

Screen Shot 2020-07-15 at 5 26 42 PM

sophstad:
Huh, weird, I refreshed another few times and it went away. False alarm I suppose!

jestapinski:
> Huh, weird, I refreshed another few times and it went away. False alarm I suppose!
> 
> Edit: wait actually, why do your screenshots not show the search box outline? On my render I can see a box with rounded left and right sides!

I believe this is what we should be seeing at this width? The style changes to just occupy the entire top nav area (no rounded corners). I wonder if there is some odd caching or something going on.

I also made some changes based on Allison's feedback and re-updated the staging link so hopefully this discrepancy is resolved. I switched to ecosystem from guides because I was noticing some odd behavior on guides (will come back to this separately, this PR is getting large enough as is 😂 )

src/hooks/useScreenSize.js
jestapinski: Allison changed the logic for when to expand/collapse by default, so I cleaned up this function to only add a new hook from the theme specifically and just used `useMediaQuery` where more complex logic was needed.

jestapinski:
Looking into the responsiveness SSR issue we found here, will unmark WIP when this is ready for a look on Monday.

src/hooks/use-media.js
```diff
@@ -0,0 +1,34 @@
+import { useEffect, useState } from 'react';
+
+// from https://github.com/streamich/react-use/blob/master/src/useMedia.ts
+
+const isClient = typeof window === 'object';
+
+const useMedia = (query, defaultState = false) => {
+  const [state, setState] = useState(defaultState);
+
+  useEffect(() => {
```
jestapinski, line 10: The SSR problem we saw with the Navbar is something we had also seen when building out the DevHub. The issue stems from Gatsby not having a `window` object at build time, causing media queries/hooks to fail silently during the SSR process and not properly be re-hydrated on mount. This is also an [existing issue the react-responsive library has](https://github.com/contra/react-responsive/issues/162) and so using this library as-is isn't enough to get another check of the window during a production build.

The solution we had for DevHub (and also referenced in this issue) is embodied in this hook we took from DevHub:
- We use a `useEffect` to run on mount
- *We explicitly check for a `window`*
- We add a listener to the window for the media query so resizing works as expected.

The `react-responsive` `useMediaQuery` suggests hardcoding options for a specific screen size, but does not check on the mount, so we were stuck with one state. I could not get the intended behavior with the library so I decided to bring in this hook.

What do you think about this? Have you seen other issues when using the `useMediaQuery` hook?

jestapinski:
I've addressed the SSR issue and added an additional staging link, so this is again ready for review.

src/hooks/useScreenSize.js
```diff
@@ -1,15 +1,18 @@
 import { useMediaQuery } from 'react-responsive';
+import useMedia from './use-media';
 
 export default function useScreenSize() {
   const isSmallScreen = useMediaQuery({ query: '(max-width: 750px)' });
   const isTabletOrMobile = useMediaQuery({ query: '(max-width: 1224px)' });
+  const isMobile = useMedia('only screen and (max-width: 374px)');
```
sophstad, line 7: Could we import this value from the theme? The same query is found [here](https://github.com/mongodb/snooty/blob/master/src/theme/docsTheme.js#L39).

src/hooks/use-media.js
sophstad: To be honest, we don't use `useMediaQuery` widely throughout the site at the moment, so I don't have any experience with it. But now I'm inclined to do some investigating... 🕵️‍♀️ 

I have, however, run into this "window is not defined" issue in other ways and like this approach. I have no allegiance to `react-responsive` so am happy with this!

src/hooks/use-media.js
```diff
@@ -0,0 +1,34 @@
+import { useEffect, useState } from 'react';
+
+// from https://github.com/streamich/react-use/blob/master/src/useMedia.ts
+
+const isClient = typeof window === 'object';
```
sophstad, line 5: [nit] We already have a util for this called [`isBrowser`](https://github.com/mongodb/snooty/blob/master/src/utils/is-browser.js). `isBrowser` uses a slightly different comparison, but it's taken directly from the [Gatsby docs](https://www.gatsbyjs.org/docs/debugging-html-builds/#how-to-check-if-window-is-defined).

src/hooks/use-media.js
sophstad: To be honest, we don't use useMediaQuery widely throughout the site at the moment, so I don't have any experience with it. But now I'm inclined to do some investigating... 🕵️‍♀️

I have, however, run into this "window is not defined" issue in other ways and like this approach. I have no allegiance to react-responsive so am happy with this!

src/hooks/useScreenSize.js
jestapinski: Ah great point! Thanks!

src/hooks/use-media.js
jestapinski: Seems to functionally work as we need!

jestapinski:
Thanks! Glad it made sense :)

PR #230, opened 2020-07-20
[DOP-1219] Searchbar Part 3: Testing

jestapinski:
This PR adds some simple unit snapshot tests for the Searchbar (closed vs open state) and augment the existing Navbar snapshot based on where #228 is going. For the `useMedia` hook I had to create a simple mock for it in `testSetup`.

### DIFF
--- a/tests/testSetup.js
+++ b/tests/testSetup.js
@@ -30,4 +30,5 @@ Object.defineProperty(global.self, 'crypto', {
   },
 });
 
+window.matchMedia = () => ({ addListener: () => {}, removeListener: () => {} });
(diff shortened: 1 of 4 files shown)


jestapinski:
> Great! 👏
> 
> I'd be remiss if I didn't point out the [`withScreenSize()`](https://github.com/mongodb/snooty/blob/master/tests/utils/index.js#L28-L41) utility we use for testing. Perhaps it will come in handy when we add some interaction tests!

Thanks for sharing! Definitely will come in handy later to check mobile styling specifically (since the Navbar controls when this is expanded). Thanks!

PR #231, opened 2020-07-21
DOP-1220 Search Results Part 1: Generic Pagination

jestapinski:
[Staging Guides](https://docs-mongodbcom-staging.corp.mongodb.com/master/guides/jordanstapinski/pagination-component/)
[Staging Ecosystem](https://docs-mongodbcom-staging.corp.mongodb.com/master/ecosystem/jordanstapinski/pagination-component/)
^ For these I just left the dropdown open for easier inspection. That is not a change made in this PR.

This PR adds a generic `Pagination` component to be used with search results. I also added interaction and snapshot testing for it. This component lets a parent component control the current page and enables/disables actions accordingly.

Screen Shot 2020-07-21 at 1 43 07 PM

### DIFF
--- a/src/components/Searchbar/Pagination.js
+++ b/src/components/Searchbar/Pagination.js
@@ -5,27 +5,15 @@ import Icon from '@leafygreen-ui/icon';
 import { uiColors } from '@leafygreen-ui/palette';
 import { theme } from '../../theme/docsTheme';
 
-const BUTTON_HEIGHT = theme.size.medium;
 const BUTTON_WIDTH = '14px';
 const ENABLED_COLOR = uiColors.gray.dark2;
 const DISABLED_COLOR = uiColors.gray.light1;
 
 const PaginationButton = styled(IconButton)`
   background-color: #fff;
-  height: ${BUTTON_HEIGHT};
   padding: 0;
   width: ${BUTTON_WIDTH};
   z-index: 1;
-  /* Below removes default hover effects from button */
-  background-image: none;
-  border: none;
-  box-shadow: none;
-  :before {
-    display: none;
-  }
-  :after {
-    display: none;
-  }
 `;
 
 const PaginationContainer = styled('div')`
--- a/src/components/Searchbar/Pagination.js
+++ b/src/components/Searchbar/Pagination.js
@@ -1,6 +1,6 @@
 import React, { useCallback, useMemo } from 'react';
 import styled from '@emotion/styled';
-import Button from '@leafygreen-ui/button';
+import IconButton from '@leafygreen-ui/icon-button';
 import Icon from '@leafygreen-ui/icon';
 import { uiColors } from '@leafygreen-ui/palette';
 import { theme } from '../../theme/docsTheme';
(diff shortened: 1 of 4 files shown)


src/components/Searchbar/Pagination.js
```diff
@@ -0,0 +1,79 @@
...
+`;
+
+const Pagination = ({ currentPage, setCurrentPage, totalPages }) => {
+  const decrementPage = useCallback(() => setCurrentPage(currentPage - 1), [currentPage, setCurrentPage]);
+  const incrementPage = useCallback(() => setCurrentPage(currentPage + 1), [currentPage, setCurrentPage]);
+  const canDecrementPage = useMemo(() => currentPage !== 1, [currentPage]);
+  const canIncrementPage = useMemo(() => currentPage < totalPages, [currentPage, totalPages]);
+  return (
+    <PaginationContainer>
+      <PaginationButton
```
sophstad, line 56: Any reason you went with this approach rather than [LeafyGreen's IconButton](https://github.com/mongodb/leafygreen-ui/tree/master/packages/icon-button)?

jestapinski: Ah, didn't even know that was a thing! I swapped over to it, thanks!

sophstad: Thanks for the quick fix! I'm looking at the staging link now and am not seeing the hover state for the icon buttons ([demo here](https://mongodb.github.io/leafygreen-ui/?path=/story/iconbutton--default)). Any idea why?

jestapinski: Ah they are `disabled` right now (mainly because there is no data so no pagination), but specifically:
- We are at page 1, so decrementing is disabled
- We are beyond the last page, so incrementing is disabled

When they are re-enabled once we have data, I'll double check for that hover state

src/components/Searchbar/Pagination.js
```diff
@@ -0,0 +1,72 @@
...
+  /* Below removes default hover effects from button */
+  background-image: none;
+  border: none;
+  box-shadow: none;
+  :before {
+    display: none;
+  }
+  :after {
+    display: none;
+  }
```
sophstad, line 28: We probably don't need this anymore

src/components/Searchbar/Pagination.js
```diff
@@ -0,0 +1,72 @@
...
+import { theme } from '../../theme/docsTheme';
+
+const BUTTON_HEIGHT = theme.size.medium;
+const BUTTON_WIDTH = '14px';
+const ENABLED_COLOR = uiColors.gray.dark2;
+const DISABLED_COLOR = uiColors.gray.light1;
+
+const PaginationButton = styled(IconButton)`
+  background-color: #fff;
+  height: ${BUTTON_HEIGHT};
```
sophstad, line 15: [nit] Perhaps it would be better to use `IconButton`'s `size` prop, but either way!

jestapinski: Ah, well it turns out the default size is the same as this!

jestapinski:
Thanks Sophie!

PR #233, opened 2020-07-22
DOP-1220 Search Results Part 2: Desktop Search Results

jestapinski:
[Ecosystem Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/ecosystem/jordanstapinski/fetch-results/)
[Guides Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/guides/jordanstapinski/fetch-results/)

This PR adds the ability to pull results from Marian and display (specifically for Desktop) in the searchbar. One idea to note is Deepak wants us to abstract Marian-specific details out of the component if possible for future re-use. As such I decided to move Marian references into utils (such as `searchParamsToUtil`) and will continue to expand on that as I continue to work on this ticket.

Highlighting relevant text, mobile responsiveness, and abstracting the marian result/removing property will come in separate PRs, this one is large enough :)

### DIFF
--- a/src/components/Searchbar/SearchResult.js
+++ b/src/components/Searchbar/SearchResult.js
@@ -15,6 +15,7 @@ const truncate = maxLines => css`
 `;
 
 const SearchResultContainer = styled('div')`
+  height: 100%;
   :hover,
   :focus {
     background-color: ${RESULT_HOVER_COLOR};
--- a/src/components/Searchbar/SearchResult.js
+++ b/src/components/Searchbar/SearchResult.js
@@ -1,0 +1,68 @@
+import React from 'react';
+import { css } from '@emotion/core';
+import styled from '@emotion/styled';
+import { theme } from '../../theme/docsTheme';
+
+const LINK_COLOR = '#494747';
+const RESULT_HOVER_COLOR = '#d8d8d8';
+
+// Truncates text to a maximum number of lines
+const truncate = maxLines => css`
+  display: -webkit-box;
+  -webkit-line-clamp: ${maxLines}; /* supported cross browser */
+  -webkit-box-orient: vertical;
+  overflow: hidden;
+`;
+
+const SearchResultContainer = styled('div')`
+  :hover,
+  :focus {
+    background-color: ${RESULT_HOVER_COLOR};
+    transition: background-color 150ms ease-in;
+  }
+`;
+
+const SearchResultLink = styled('a')`
+  color: ${LINK_COLOR};
+  height: 100%;
+  text-decoration: none;
+  :hover,
+  :focus {
+    color: ${LINK_COLOR};
+    text-decoration: none;
+  }
+`;
+
+const StyledPreviewText = styled('p')`
+  font-family: Akzidenz;
+  font-size: 14px;
+  letter-spacing: 0.5px;
+  line-height: 20px;
+  margin-bottom: 0;
+  margin-top: 0;
+  ${({ maxLines }) => truncate(maxLines)};
+`;
+
+const StyledResultTitle = styled('p')`
+  font-family: Akzidenz;
+  font-size: 14px;
+  line-height: ${theme.size.medium};
+  letter-spacing: 0.5px;
+  height: ${theme.size.medium};
+  margin-bottom: 6px;
+  margin-top: 0;
+  ${truncate(1)};
+`;
+
+const SearchResult = React.memo(({ maxLines = 2, preview, title, url, ...props }) => (
+  <SearchResultLink href={url}>
+    <SearchResultContainer {...props}>
+      <StyledResultTitle>
+        <strong>{title}</strong>
+      </StyledResultTitle>
+      <StyledPreviewText maxLines={maxLines}>{preview}</StyledPreviewText>
+    </SearchResultContainer>
+  </SearchResultLink>
+));
+
+export default SearchResult;
--- a/src/components/Searchbar/SearchResult.js
+++ b/src/components/Searchbar/SearchResult.js
@@ -4,7 +4,6 @@ import styled from '@emotion/styled';
 import { theme } from '../../theme/docsTheme';
 
 const LINK_COLOR = '#494747';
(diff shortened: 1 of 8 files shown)


src/components/Searchbar/Searchbar.js
```diff
@@ -211,15 +213,32 @@ const Searchbar = ({ isExpanded, setIsExpanded }) => {
...
+  const onSearchChange = useCallback(
+    e => {
+      const value = e.target.value;
+      setValue(e.target.value);
+      // Debounce any queued search event since the query has changed
+      clearTimeout(searchEvent);
+      if (value) {
+        // Set a timeout to trigger the search to avoid over-requesting
+        setSearchEvent(
+          setTimeout(async () => {
```
jestapinski, line 233: I tried using the `debounce` function Nick added here a while ago, but was having issues getting the proper search term to bind to the timeout function, so I decided just to manage this with hooks :)

src/components/Searchbar/Searchbar.js
```diff
@@ -211,15 +213,32 @@ const Searchbar = ({ isExpanded, setIsExpanded }) => {
...
+            const result = await fetch(searchParamsToURL(value, {}));
+            const resultJson = await result.json();
+            setSearchResults(resultJson.results);
```
jestapinski, line 236: This will be abstracted in a next PR.

jestapinski:
> I'm unable to navigate through the paginated results—clicking on the arrow, or any part of the dropdown, closes the dropdown. Are you able to reproduce? If not, I'll take a screencast!

Ah I see this on Firefox now (was only looking at chrome before). Let me fix this, thanks for catching!

src/hooks/use-click-outside.js
```diff
@@ -0,0 +1,31 @@
+import { useEffect } from 'react';
+
+/**
+ * This hook fires an onClickOutside handler if the given node ref is clicked
+ * outside of or the escape key is pressed
+ * @param {*} ref a node which we will fire the onClickOutside handler if clicked
+ * outside of
+ * @param {*} onClickOutside a callback handler
+ */
+export const useClickOutside = (ref, onClickOutside) => {
```
jestapinski, line 10: Thanks for finding the clicking issue! Turns out the `onBlur` event handler in react is nice, but it was getting pretty complicated with children. This approach is a bit more standard for cases like this and just so much cleaner overall.

This hook binds an event to check if something outside of the searchbar is clicked and runs the `onBlur` handler then. It is much more reliable than the old way and (I checked this time) works on Chrome, FF, and Safari! You should no longer have that issue.

src/components/Searchbar/SearchResult.js
```diff
@@ -0,0 +1,66 @@
...
+  text-decoration: none;
+  :hover,
+  :focus {
+    color: ${LINK_COLOR};
+    text-decoration: none;
+  }
+`;
+
+const StyledPreviewText = styled('p')`
+  font-family: 'Akzidenz Grotesk BQ Light';
```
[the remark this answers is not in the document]
jestapinski: I checked this out but it doesn't quite give the font we were getting. Not sure if the `200` is being interpreted correctly, as no value really changed the look (except `bolder`).

Here is what it should be (with the BQ Light):
Screen Shot 2020-07-23 at 4 58 52 PM

Here is the 200:
Screen Shot 2020-07-23 at 4 59 01 PM

src/hooks/use-click-outside.js
```diff
@@ -0,0 +1,31 @@
...
+      }
+    };
+    const handleEscape = e => {
+      e = e || window.event;
+      if (e.key === 'Escape') {
+        onClickOutside();
+      }
+    };
+    document.addEventListener('mousedown', handleClickOutside);
+    // Cannot handle ESC on FF https://bugzilla.mozilla.org/show_bug.cgi?id=1443758
```
jestapinski, line 24: Wanted to use [Leafy's hook](https://github.com/mongodb/leafygreen-ui/blob/master/packages/hooks/src/useEscapeKey.ts) but it had some SSR issues (since it seems to use `document` outside of a `useEffect`).

sophstad: Good find! Have you let them know about this bug? I believe SSR is supposed to be supported!

jestapinski: I will soon! Wanted to investigate this a bit more independently before posting in leafygreen-ui

src/hooks/use-click-outside.js
sophstad: Thanks for figuring that out! It's working great 💯

src/components/Searchbar/SearchResult.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import { css } from '@emotion/core';
+import styled from '@emotion/styled';
+import { theme } from '../../theme/docsTheme';
+
+const LINK_COLOR = '#494747';
+const RESULT_HOVER_COLOR = '#d8d8d8';
```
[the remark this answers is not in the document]
jestapinski: Ah I think Allison agrees with you. We have been keeping design feedback [in this doc](https://docs.google.com/document/d/1x_g96vDj1QZx7S1XilpAEGQNRhjRyaoN9ctmvLPggc0/edit#) for reference.

jestapinski:
> Amazing work, can't believe how quickly this has come together! Just some thoughts, nothing blocking.
> 
> Now that hover is enabled, I noticed this smushed circle looks slightly weird (due to [this line](https://github.com/mongodb/snooty/pull/231/files#diff-3602c781439bcc7fed64a741b0c06916R15), I imagine). If Allison is okay with it then fine by me, just thought it was worth pointing out.
> 
> image

Ah good point. Yeah I wanted to keep the chevron icon close but this is one negative side effect. @allisonmui9 any ideas here?

jestapinski:
I'm going to go ahead and merge this to keep the code going, but have made notes of the pagination button for us to continue discussing separately.

Thanks Sophie!

PR #234, opened 2020-07-24
DOP-1220 Part 3: Abstract Marian Result Parsing, Limit Results

jestapinski:
[Ecosystem Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/ecosystem/jordanstapinski/abstract-marian/)
[Guides Staging](https://docs-mongodborg-staging.corp.mongodb.com/master/guides/jordanstapinski/abstract-marian/)

This PR pulls the property out of the search title and abstracts out data postprocessing after fetching the search results to a helper function called `getSearchbarResultsFromJSON`. This function adds the ability to pull the property title out as well as optionally limit the number of returned results. We also will use this on the dedicated search page.

Allison would like us to limit dropdown results to at most 9 (3 pages), so this helper function gives us that functionality (with tests!).

I also updated the relevant snapshots since I did not do so in the previous PR.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -25,5 +25,12 @@ module.exports = {
         '^.+\\.jsx?$': `<rootDir>/jest-preprocess.js`,
       },
     },
+    {
+      displayName: 'utils',
+      testMatch: ['<rootDir>/tests/utils/*.test.js'],
+      transform: {
+        '^.+\\.jsx?$': `<rootDir>/jest-preprocess.js`,
+      },
+    },
   ],
 };
--- a/src/components/Navbar.js
+++ b/src/components/Navbar.js
@@ -94,7 +94,7 @@ const Navbar = () => {
         data-navprops={navprops}
       />
       <Searchbar
-        getResultsFromJson={getSearchbarResultsFromJSON}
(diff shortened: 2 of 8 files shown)


jest.config.js
```diff
@@ -25,5 +25,12 @@ module.exports = {
         '^.+\\.jsx?$': `<rootDir>/jest-preprocess.js`,
       },
     },
+    {
+      displayName: 'utils',
```
jestapinski, line 29: I noticed my test in `utils` was throwing issues so I added this into the config so it would run properly.

src/components/Navbar.js
```diff
@@ -93,6 +94,7 @@ const Navbar = () => {
         data-navprops={navprops}
       />
       <Searchbar
+        getResultsFromJson={getSearchbarResultsFromJSON}
```
sophstad, line 97: [nit] Noticed `Json` vs. `JSON` in these variables—may be nice to standardize.

jestapinski: ah good point! I chose JSON since that is more of the convention.

jestapinski:
Thanks!

PR #236, opened 2020-07-27
DOP-1220 Part 5: Mobile Search Results

jestapinski:
[Ecosystem Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/ecosystem/jordanstapinski/mobile-results/)
[Guides Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/guides/jordanstapinski/mobile-results/)

This PR is the last for displaying search results (aside from 2 design comments I will tackle separately) and adds in the mobile design. I did see a console error since here we add an `<a>` within an `<a>` so I want to ask @allisonmui9 if the card is still fully a link on mobile.

I also re-updated the snapshots to include the new context provider from before.

Screen Shot 2020-07-27 at 1 10 11 PM

### DIFF
--- a/src/components/Searchbar/SearchResult.js
+++ b/src/components/Searchbar/SearchResult.js
@@ -33,7 +33,7 @@ const LearnMoreLink = styled('a')`
 
 const SearchResultContainer = styled('div')`
   height: 100%;
-  @media ${theme.screenSize.upToXSmall} {
+  @media ${theme.screenSize.upToSmall} {
     display: flex;
     flex-direction: column;
     height: 100%;
@@ -74,7 +74,7 @@ const StyledResultTitle = styled('p')`
   margin-bottom: 6px;
   margin-top: 0;
   ${truncate(1)};
-  @media ${theme.screenSize.upToXSmall} {
+  @media ${theme.screenSize.upToSmall} {
     font-size: ${theme.size.default};
     line-height: ${theme.size.medium};
   }
--- a/src/components/Searchbar/SearchResult.js
+++ b/src/components/Searchbar/SearchResult.js
@@ -1,9 +1,14 @@
-import React from 'react';
+import React, { useContext } from 'react';
+import sanitizeHtml from 'sanitize-html';
 import { css } from '@emotion/core';
 import styled from '@emotion/styled';
+import { uiColors } from '@leafygreen-ui/palette';
 import { theme } from '../../theme/docsTheme';
+import SearchContext from './SearchContext';
 
 const LINK_COLOR = '#494747';
+// Use string for match styles due to replace/innerHTML
+const SEARCH_MATCH_STYLE = `background-color: ${uiColors.yellow.light2};`;
 
 // Truncates text to a maximum number of lines
 const truncate = maxLines => css`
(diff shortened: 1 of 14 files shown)


sophstad:
Clicking the magnifying glass on my phone leads to some unexpected/confusing zoom behavior on focus. It's definitely possible this was introduced on the first mobile PR, so sorry for not catching. Would love to avoid this if possible!
bug

jestapinski:
> Clicking the magnifying glass on my phone leads to some unexpected/confusing zoom behavior on focus. It's definitely possible this was introduced in the first mobile PR, so sorry for not catching. Would love to avoid this if possible!
> bug

ah I see what is going on. I'm going to bump up the mobile to `Small` from `XSmall` for iPhone X/Plus to also use this size. That will prevent the zoom. I also noticed the container is blocking links from being clicked so will also address that. These were definitely introduced before, thanks for the look! Will push up once this is fixed.

src/components/Searchbar/SearchResult.js
```diff
@@ -86,6 +109,13 @@ const SearchResult = React.memo(({ maxLines = 2, preview, title, url, ...props }
             __html: sanitizePreviewHtml(highlightedPreviewText),
           }}
         />
+        {learnMoreLink && (
+          <MobileFooterContainer>
+            <LearnMoreLink url={url}>
```
sophstad, line 114: Why the `url` prop rather than `href` (since `LearnMoreLink` is a styled `a` tag)? Does this have to do with the nested `<a>` elements?

jestapinski: oops! You are right

jestapinski:
Looking into the iOS use, it raised some small issues but working through them now. That inactive grey bar should not occur!

jestapinski:
Okay here is a summary of changes made to fix iOS/mobile issues:
- For the zooming in, [this is actually iOS Safari-specific behavior](https://stackoverflow.com/questions/2989263/disable-auto-zoom-in-input-text-tag-safari-on-iphone). The best approach is to just bump up the input font size from 14px to 16px. To be honest, it looks nice on my phone and addresses the zoom issue. The alternatives would disable scaling by the user site-wide.
- Fixed a bug where onBlur would unexpectedly re-open the panel for mobile.
- Swapped `isMobile`/media queries to go up to `small` instead of `xsmall` for iPhone X and larger compatibility (previously was the desktop bar)
- Explicly set the search bar to white at all times on mobile

I *think* this should patch up those issues! The links have been updated. Thanks so much for the thoroughness!

src/components/Searchbar/Searchbar.js
```diff
@@ -193,10 +197,18 @@ const SearchbarContainer = styled('div')`
     }
   }
   @media ${theme.screenSize.upToSmall} {
-    height: ${({ isSearching }) => (isSearching ? '100%' : `${SEARCHBAR_HEIGHT}px`)};
+    height: ${({ isExpanded, isSearching }) => (isExpanded && isSearching ? '100%' : `${SEARCHBAR_HEIGHT}px`)};
```
jestapinski, line 200: This fixed an issue where just having the bar up top would also prevent users with interacting with content (links) on the page. Now we only set this height when in use.

src/components/Searchbar/Searchbar.js
```diff
@@ -214,8 +226,8 @@ const Searchbar = ({ getResultsFromJSON, isExpanded, setIsExpanded, searchParams
   const onFocus = useCallback(() => setIsFocused(true), []);
   const onBlur = useCallback(() => {
     setIsFocused(false);
-    setIsExpanded(!!value);
-  }, [setIsExpanded, value]);
+    setIsExpanded(false);
```
jestapinski, line 229: Not sure what behavior we got before, but this definitely seemed cleaner. Perhaps this logic is getting a bit complex so a refactor might be in order :)

PR #237, opened 2020-07-29
Refactor Searchbar into Smaller Components

jestapinski:
[Ecosystem Staging](https://docs-mongodborg-staging.corp.mongodb.com/master/ecosystem/jordanstapinski/refactor-search-component/)
[Prior Branch Ecosystem Staging](https://docs-mongodborg-staging.corp.mongodb.com/master/ecosystem/jordanstapinski/mobile-results/)
[Guides Staging](https://docs-mongodborg-staging.corp.mongodb.com/master/guides/jordanstapinski/refactor-search-component/)

This PR eliminates some building technical debt by splitting the `Searchbar` component into four components:
- The main `Searchbar` parent which acts as an overall controller of the experience
- `CondensedSearchbar` which is the magnifying glass to pop-open the searchbar
- `ExpandedSearchbar` which is the searchbar open
- `SearchTextInput` which is the custom LeafyGreen text input

These were all in one component before and it was getting tricky to manage state and control changes. This approach is much more maintainable and clean! Passing state now also feels more natural as well.

I'll bump the snapshots separately since this is large enough. I looked side by side on this branch versus the old one (and on mobile). The only change is on mobile the magnifying glass while searching is always filled in. This is actually what the design calls for anyway. No other looks or functionality should change.

### DIFF
--- a/src/components/Searchbar/CondensedSearchbar.js
+++ b/src/components/Searchbar/CondensedSearchbar.js
@@ -1,0 +1,46 @@
+import React from 'react';
+import styled from '@emotion/styled';
+import IconButton from '@leafygreen-ui/icon-button';
+import Icon from '@leafygreen-ui/icon';
+import { uiColors } from '@leafygreen-ui/palette';
+import { theme } from '../../theme/docsTheme';
+
+// Defining as a styled component allows us to use as a selector in ExpandButton
+const ExpandMagnifyingGlass = styled(Icon)``;
+
+const ExpandButton = styled(IconButton)`
+  background-color: #fff;
+  background-image: none;
+  border: none;
+  border-radius: ${theme.size.medium};
+  box-shadow: none;
+  height: ${theme.size.large};
+  position: absolute;
+  right: ${theme.size.small};
+  /* 32px button in a 36px container, 2px top gives equal spacing */
+  top: 2px;
+  width: ${theme.size.large};
+  z-index: 1;
+  :hover,
+  :focus {
+    background-color: #f7f9f8;
+    ${ExpandMagnifyingGlass} {
+      color: ${uiColors.gray.dark3};
+      transition: color 150ms ease-in;
+    }
+  }
+  :before {
+    display: none;
+  }
+  :after {
+    display: none;
+  }
+`;
+
+const CondensedSearchbar = ({ onExpand }) => (
+  <ExpandButton aria-label="Open MongoDB Docs Search" onClick={onExpand}>
+    <ExpandMagnifyingGlass glyph="MagnifyingGlass" fill={uiColors.gray.base} />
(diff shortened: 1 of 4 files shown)


src/components/Searchbar/CondensedSearchbar.js
```diff
@@ -0,0 +1,46 @@
...
+  :before {
+    display: none;
+  }
+  :after {
+    display: none;
+  }
+`;
+
+const CondensedSearchbar = ({ onExpand }) => (
+  <ExpandButton aria-label="Open MongoDB Docs Search" onClick={onExpand}>
```
sophstad, line 41: Sorry for missing this the first time around, but why is the `onClick` handler attached here instead of to the `ExpandMagnifyingGlass` `IconButton`?

src/components/Searchbar/Searchbar.js
```diff
@@ -1,182 +1,29 @@
...
 
 const BUTTON_SIZE = theme.size.medium;
-const GO_BUTTON_COLOR = uiColors.green.light3;
-const GO_BUTTON_SIZE = '20px';
 const NUMBER_SEARCH_RESULTS = 9;
 const SEARCH_DELAY_TIME = 200;
-const SEARCHBAR_DESKTOP_WIDTH = 372;
-const SEARCHBAR_HEIGHT = 36;
+const SEARCHBAR_DESKTOP_WIDTH = '372px';
+const SEARCHBAR_HEIGHT = '36px';
```
sophstad, line 16: Good call standardizing these 💯

src/components/Searchbar/CondensedSearchbar.js
jestapinski: Ah the button wraps around the icon and this is where the click handler is meant to be bound. This actually also increases the radius where a user can click and something can happen as well.

src/components/Searchbar/Searchbar.js
jestapinski: Yeah it was bothering me haha

src/components/Searchbar/CondensedSearchbar.js
jestapinski: Ah! So with the IconButton we pass the Icon as a child. We use `styled` here because I use it as a selector when styling the `ExpandButton`. It is really just an `Icon` but it has to be a styled(Icon) to use the selector as such.

jestapinski:
> I noticed that closing the search bar on mobile and reopening it preserves the query that was entered previously. This felt a little unexpected, so just wanted to point it out—perhaps Allison has an opinion on it!

oh interesting! This is the intended behavior but we could clear out when closed. It might be nice if they close by accident or something

jestapinski:
Thanks!

PR #239, opened 2020-07-30
Selectively Autofocus Search Input

jestapinski:
[Ecosystem Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/ecosystem/jordanstapinski/selectively-autofocus/)

This PR adds a new prop to the `Searchbar` to control when `autofocus` is enabled. Previously we would `autofocus` all the time on the searchbar, but this PR chnages the behavior a bit for a good UX:
- If the searchbar is expanded by default (desktop), then we do not autofocus the searchbar
- If the searchbar is condensed by default (mobile/~1200px), when the user clicks the magnifying glass to open the searchbar then we do autofocus, since they should be able to type right away.

Using the SearchContext this was easy to add without prop drilling.

Here is what should happen in the "condensed --> expanded" action. This way the user already has the intent to search, and leaving autofocus off would require them to click again to focus and type.

Screen Shot 2020-07-30 at 1 09 50 PM
Screen Shot 2020-07-30 at 1 09 59 PM

### DIFF
--- a/src/components/Searchbar/SearchContext.js
+++ b/src/components/Searchbar/SearchContext.js
@@ -1,6 +1,6 @@
 import { createContext } from 'react';
 
 // Simple context to pass search results to children
-const SearchContext = createContext('');
+const SearchContext = createContext({ searchTerm: '', shouldAutofocus: false });
 
 export default SearchContext;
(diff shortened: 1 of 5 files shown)


src/components/Searchbar/SearchContext.js
```diff
@@ -1,6 +1,6 @@
 import { createContext } from 'react';
 
 // Simple context to pass search results to children
-const SearchContext = createContext('');
+const SearchContext = createContext({ searchTerm: '', shouldAutofocus: false });
```
jestapinski, line 4: `searchTerm` felt more descriptive than `value`

jestapinski:
💯

PR #243, opened 2020-08-03
Add reset functionality to Select

sophstad:
No staging link because we're not using this anywhere yet. Also included the color fix. Thanks!

### DIFF
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -43,7 +43,7 @@ const Option = styled('li')`
   white-space: nowrap;
   :focus,
   :hover {
-    background-color: rgba(231, 238, 236, 0.2);
+    background-color: rgb(250, 252, 251);
   }
 `;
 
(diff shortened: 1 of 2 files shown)


src/components/Select.js
```diff
@@ -106,9 +107,11 @@ const Select = ({ choices, onChange, defaultText = '', disabled = false, label =
         const currentChoice = choice[0];
         setSelectText(currentChoice.text);
       }
+    } else {
```
jestapinski, line 110: Ah so the reset is coming from the parent right? The `select` itself does not reset but is just passed `null` from the parent.

sophstad: Indeed! Let me know if you think a different approach would make more sense.

jestapinski: This seems reasonable! Also real easy to implement

PR #245, opened 2020-08-03
DOP-1298 Parse Marian Manifests

jestapinski:
No staging link!

This PR adds some logic to parse and format raw Marian Manifest content for use in the Dropdowns for filtering. The approach I took was ultimately creating a data structure that looks like this:

`{ <Property Name>: { <Branch Name> : <Search Property> } }`

So that gives us something like `{"Ecosystem": { "master": "ecosystem-master" }}` for example. This will allow us to easily use this data with the dropdown filters and then go ahead and search for it. The data in the tests can better illustrate this. Ideally I wish we could only run this map once during build time but then we would have to add it to every page context and that could be messy 🤷‍♂️ , any thoughts?

Since the raw data was a bit messy, I went ahead and applied the following transformations:
- Capitalized each word
- Applied a mapping of some terms to better formatted versions called `PROPERTY_MAPPING`
- There was one outlier of odd/duplicated data which I commented on called "Charts 19-06" (there is also "Charts 19.06"). I just removed this


src/utils/parse-marian-manifests.js
```diff
@@ -0,0 +1,38 @@
...
+  'Docs Ruby': 'Ruby Driver',
+  Manual: 'MongoDB Manual',
+  'Mms Cloud': 'MMS Cloud',
+  'Mms Onprem': 'MMS On Prem',
+  'Mongodb Vscode': 'MongoDB VSCode',
+};
+
+const capitalizeString = s => s.charAt(0).toUpperCase() + s.slice(1);
+
+export const getSortedBranchesForProperty = (parsedManifest, property) => {
```
jestapinski, line 15: We don't really need this utility function but it might be nice to have in here.

src/utils/compare-branches-with-version-numbers.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Function to be used with sort() to sort a list of git branch names in the
+ * following order:
+ * - "current" is the first entry, if applicable
+ * - "master" is the second entry, if applicable
+ * - Branch version numbers are then provided in most-recent-first order
+ * @param {string} branchA the first branch
+ * @param {string} branchB
+ */
+export const compareBranchesWithVersionNumbers = (branchA, branchB) => {
```
jestapinski, line 10: Split this out to give more explanation and not bloat the other file. It is also tested when checking sort order but we could get more explicit with testing this.

src/utils/compare-branches-with-version-numbers.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Function to be used with sort() to sort a list of git branch names in the
```
sophstad, line 2: To clarify, we are sorting these so that they appear in the correct order in the dropdown, right?

jestapinski: Yes so in this representation we are using an object, so when we go to compute the labels to put in the dropdown this function would help us to sort the branches in particular, since they have some ordering rules. Does this make sense?

src/utils/parse-marian-manifests.js
```diff
@@ -0,0 +1,38 @@
...
+
+const PROPERTY_MAPPING = {
+  'Bi Connector': 'BI Connector',
+  'Docs Php Library': 'PHP Driver',
+  'Docs Ruby': 'Ruby Driver',
+  Manual: 'MongoDB Manual',
+  'Mms Cloud': 'MMS Cloud',
+  'Mms Onprem': 'MMS On Prem',
+  'Mongodb Vscode': 'MongoDB VSCode',
+};
```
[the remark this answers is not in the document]
jestapinski: Ah I was trying to save the amount of explicit mapping by just `capitalizing` and splitting some of these properties, so they didn't have to be in the mapping (e.g. `kubernetes-operator` --> `Kubernetes Operator`) which seemed general enough to work but I think it might make sense to just include everything into the mapping for now. If a new property is added we could then default to capitalizing?

src/utils/parse-marian-manifests.js
```diff
@@ -0,0 +1,38 @@
+import { compareBranchesWithVersionNumbers } from './compare-branches-with-version-numbers';
+
+const PROPERTY_MAPPING = {
```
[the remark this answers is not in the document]
jestapinski: I think this combined with adding everything into the mapping would solve this?

src/utils/compare-branches-with-version-numbers.js
sophstad: Yes that makes sense! I think it may be desirable to obfuscate the idea of "current" and "master" from the user in some way, because those are more like internal concepts than user-friendly labels. Again, I think it would help to chat with Andrew when they're back to try and understand this. And Allison can help us from a design perspective.

Happy with this for now, and perhaps we'll tweak what's happening here when we have more information!

jestapinski: Got it! We can replace that in `parseMarianManifests` should we want to and then update here as well to reflect stable, since the term "master" is also starting to be decommissioned

src/utils/parse-marian-manifests.js
```diff
@@ -0,0 +1,63 @@
...
+
+const capitalizeString = s => s.charAt(0).toUpperCase() + s.slice(1);
+
+export const getSortedBranchesForProperty = (parsedManifest, property) => {
+  const branches = Object.keys(parsedManifest[property]);
+  branches.sort(compareBranchesWithVersionNumbers);
+  return branches;
+};
+
+export const parseMarianManifest = manifest => {
```
jestapinski, line 35: Split this to allow us to restore dropdown values later

src/utils/parse-marian-manifests.js
```diff
@@ -0,0 +1,63 @@
...
+  'docs-php-library': 'PHP Driver',
+  'docs-ruby': 'Ruby Driver',
+  drivers: 'Drivers',
+  ecosystem: 'Ecosystem',
+  guides: 'Guides',
+  'kafka-connector': 'Kafka Connector',
+  'kubernetes-operator': 'Kubernetes Operator',
+  manual: 'MongoDB Manual',
+  'mms-cloud': 'MMS Cloud',
+  'mms-onprem': 'MMS On Prem',
```
sophstad, line 19: These links actually refer to the [Cloud Manager](https://docs.cloudmanager.mongodb.com) and [Ops Manager](https://docs.opsmanager.mongodb.com/current/) docs, respectively, so I think we could use those terms for the labels.

src/utils/parse-marian-manifests.js
```diff
@@ -0,0 +1,63 @@
...
+  mongocli: 'MongoDB CLI',
+  'mongodb-vscode': 'MongoDB VSCode',
```
sophstad, line 21: [nit] This product goes by [MongoDB for VS Code](https://docs.mongodb.com/mongodb-vscode/).

src/utils/parse-marian-manifests.js
```diff
@@ -0,0 +1,63 @@
+import { compareBranchesWithVersionNumbers } from './compare-branches-with-version-numbers';
+
+const PROPERTY_MAPPING = {
+  atlas: 'Atlas',
+  'bi-connector': 'BI Connector',
+  charts: 'Charts',
+  compass: 'Compass',
+  'database-tools': 'Database Tools',
+  datalake: 'Datalake',
```
sophstad, line 9: I think [Atlas Datalake](Atlas Data Lake) would be a more descriptive label.

jestapinski:
Thanks! Added those in

PR #246, opened 2020-08-06
Update misc select component styling

jestapinski:
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/DOP-909/landing/jordanstapinski/select-updates/legacy)

This PR fixes a few issues with the `Select` component:
1. Fixes the focus border so it again appears normally and does not appear on dropdown elements (we have a hover state here already)
2. Adds max-height/scroll for > 160px (discussed with Allison)
3. Fixes text overflow ellipsis to actually work by using display block :) 
4. Fixes size of element changing when open vs closed
5. Removes disabled selects from the taborder

### DIFF
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -172,7 +172,7 @@ const Select = ({ choices, onChange, defaultText = '', disabled = false, label =
         onKeyDown={showOptionsOnEnter}
         role="listbox"
         showOptions={showOptions}
-        tabIndex="0"
+        tabIndex={enabled ? '0' : null}
       >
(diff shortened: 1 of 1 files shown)


jestapinski:
> Noticing this weird behavior with the focus outline on Safari:

Good catch! There is some odd hover behavior directly related to adding the background color. I couldn't figure out why this was the case, so a happy median here is to allow the outline when focused but remove it when expanded (since we have some other hover states in place:

Screen Shot 2020-08-06 at 1 17 15 PM
Screen Shot 2020-08-06 at 1 17 22 PM


What do you think?

PR #250, opened 2020-08-07
DOP-1312 Actually Filtering Part 1 (UI only/non-functional)

jestapinski:
[Staging Link](https://docs-mongodbcom-staging.corp.mongodb.com/master/ecosystem/jordanstapinski/advanced-filters-redux/)
[Updated Relevant Invision](https://mongodb.invisionapp.com/d/#/console/19596736/417891823/preview)

This PR adds the UI for filtering search results. I was additionally going to include the logic to actually filter here as well but after speaking with Elle and Allison I want to change my approach there so I figured this could be split into two PRs. This PR does the following:
1. Flips footer button ordering of pagination and "Advanced Filters" by request from Allison
2. Adds a `reset` button on the advanced filter pane (that does not do anything yet)
3. Adds the search dropdowns to the advanced filter pane, with logic to control enabled/disabled
4. Adds a custom hook to get Marian results (when Marian is available. For now I hardcoded some samples to see)
5. Cleans up the `searchParamsToURL` function to account for a string filter.

### DIFF
--- a/src/components/Searchbar/AdvancedFiltersPane.js
+++ b/src/components/Searchbar/AdvancedFiltersPane.js
@@ -5,6 +5,7 @@ import Button from '@leafygreen-ui/button';
 import Icon from '@leafygreen-ui/icon';
 import { uiColors } from '@leafygreen-ui/palette';
 import { theme } from '../../theme/docsTheme';
+import SearchFilters from './SearchFilters';
 
 const fadeIn = keyframes`
   from {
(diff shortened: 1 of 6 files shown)


jestapinski:
> Just a thought: should we consider hiding the dropdown for unversioned products? It's a bit awkward to select a product like Atlas, which has one version of documentation, and choose an version from a dropdown of one item. Definitely would like Allison's input here though!

I think this is a great point, but would also require us to change a few things around for it to function properly:
1. We would have to get rid of the `(2)` hardcoded for advanced filtering and find some way to only display `(1)` is this has one branch. This shouldn't be worse than passing another `state` hook
2. Curious if some don't feature the `version` selector if there would be a worse UX due to it being inconsistent?
3. We would want to not render the second dropdown until the first is chosen (instead of the opacity change)
4. Any implications for the filters on the new search page?

@allisonmui9 thanks in advance!

jestapinski:
Thanks Sophie!

PR #251, opened 2020-08-10
DOP-1312 Part 2: Implement Filter Logic

jestapinski:
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/ecosystem/jordanstapinski/filter-logic/)

This PR adds the logic to support filtering via Marian. The pattern I took here was a bit odd but I think makes the most sense given the problem. We want to have a "working" search filter (meaning we edit the dropdowns) but don't actually search until the user hits "apply search criteria", so the parent exposes a state variable to update the working search filter and a callback to "promote" this working filter to the active search filter, which kicks off a new request for a search.

Open to feedback here because the use case is a bit odd but this made the most sense out of what I tried. I also put these values to be passed to children into the context to avoid a ton of prop drilling.

I also noticed we didn't have an empty state so I added one in, as well as disabled the second dropdown if a property only has one branch (Allison's suggestion)

### DIFF
--- a/src/components/Searchbar/AdvancedFiltersPane.js
+++ b/src/components/Searchbar/AdvancedFiltersPane.js
@@ -5,6 +5,7 @@ import Button from '@leafygreen-ui/button';
 import Icon from '@leafygreen-ui/icon';
 import { uiColors } from '@leafygreen-ui/palette';
 import { theme } from '../../theme/docsTheme';
+import SearchFilters from './SearchFilters';
 
 const fadeIn = keyframes`
   from {
@@ -32,7 +33,7 @@ const StyledReturnButton = styled(Button)`
   letter-spacing: 0.5px;
   line-height: ${theme.size.default};
   margin: 0;
-  padding: ${theme.size.tiny};
+  padding: 0;
   /* Below removes default hover effects from button */
   background: none;
   background-image: none;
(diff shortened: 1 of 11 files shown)


jestapinski:
Ah yeah we can add the current/stable in here! I was going to wait but now also seems like a good time so will work on getting that in here.

As for datalake, yes I did see this! We get empty titles from Marian so I am not sure what is going on here. I raised this with Elle and it seems we might want to get some of this data fixed but this seems something coming from either Marian or the writers, so I am not familiar.

jestapinski:
Thanks!

skerschb:
Just wanted to add a UX comment -- behaviorally I expect to be able to navigate through the search results after I've started typing, by hitting the arrow keys. Wondered what @eshwer thought of this as something we should think about for next iteration.

eshwer:
Hmm I would have also expected that behavior. I think we may have discussed it in Slack but definitely would like to include this in a future version if we can't include it in the MVP. @jestapinski do you know if there is a ticket to enable this behavior?

jestapinski:
We don't have a ticket for this functionality yet, but I do have some questions on when this should be active (since typically tab is used in this case). Here is perhaps my understanding of what might make sense:
- If a user is focused in the text box with a query, using the arrow keys should stay within the text box
- If a user is focused on a search result, using arrow keys mimics tabbing

I cannot recall an instance of this pattern though, are there examples of this somewhere on a search UI done well?

skerschb:
Please don't judge me for my eternal quest for lysol wipes.

Screen Shot 2020-08-17 at 9 37 45 AM

eshwer:
This is correct: If a user is focused in the text box with a query, using the arrow keys should stay within the text box
However I think that applies for the left/right arrow. Whereas up/down would move through the results.

Google also does this. Tab and up down arrow move through the results.

skerschb:
Yes on instacart as well, up/down moves through the results, left/right moves within the search query box.

jestapinski:
Lysol wipes are very important in these trying times! Ah the up and down arrow keys make sense (I was thinking left/right for some reason), thanks all for clarifying!

I went ahead and captured this [in a ticket](https://jira.mongodb.org/browse/DOP-1367) and will try to get to this if possible late this sprint.

PR #254, opened 2020-08-14
DOP-1340 Part 1: Instrumentation for SearchQuery and SearchSelection

jestapinski:
[JIRA](https://jira.mongodb.org/browse/DOP-1340)
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/ecosystem/jordanstapinski/search-instrumentation/) (Nothing super exciting)

Adding Elle so she can make sure the reporting structure is good.

This PR is the first of two parts (the second being focus/blur) of adding instrumentation for the search experience. Specifically we add two `reportAnalytics` calls that match a comment left in the ticket.
- `SearchQuery` triggers anytime a search is sent to Marian and includes the query in the payload
- `SearchSelection` triggers when an option is selected and reports both the overall rank of the item and the url.

Here is a sample of this going through to Segment with the proper event name and payload (this example is for `SearchQuery`, but you get the idea for `SearchSelection`. I pulled this from the network tab right off staging.
Screen Shot 2020-08-14 at 1 26 08 PM


jestapinski:
🤙

PR #257, opened 2020-08-18
DOP-1367: Add arrow key support

jestapinski:
[JIRA](https://jira.mongodb.org/browse/DOP-1367)
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/DOP-1337/landing/jordanstapinski/toggle-arrows/)

This PR adds arrow key support within the searchbar in the following way:
- If the user is focused in the search text box, down jumps to the results while up/left/right stays within the box
- If the user is focused in the search results, they jump to the next or previous result (depending on which key is pressed).
- If the user is focused on the first result and the up arrow is pressed, the searchbar is focused
- If the user is focused on the last result and the down arrow is pressed, the first result is focused (loops)

This will end up causing a conflict with the enter key addition on the search results page but should be easy to resolve

I also added a fix for tracking focus and added an active state back in for the Go button

### DIFF
--- a/src/components/Searchbar/ExpandedSearchbar.js
+++ b/src/components/Searchbar/ExpandedSearchbar.js
@@ -1,6 +1,7 @@
 import React, { useCallback, useContext, useMemo, useRef } from 'react';
 import { css } from '@emotion/core';
 import styled from '@emotion/styled';
+import { useEventListener } from '@leafygreen-ui/hooks';
 import Icon from '@leafygreen-ui/icon';
 import IconButton from '@leafygreen-ui/icon-button';
 import { uiColors } from '@leafygreen-ui/palette';
@@ -89,6 +90,7 @@ const ExpandedSearchbar = ({ isFocused, onChange, onMobileClose }) => {
   );
 
   const goButton = useRef(null);
+  const searchTextbox = useRef(null);
 
   const onKeyDown = useCallback(
     e => {
@@ -107,12 +109,14 @@ const ExpandedSearchbar = ({ isFocused, onChange, onMobileClose }) => {
     [searchContainerRef]
   );
 
+  useEventListener('keydown', onKeyDown, { dependencies: [searchTextbox.current], element: searchTextbox.current });
+
   const searchUrl = useMemo(() => searchParamsToURL(searchTerm, searchFilter, false), [searchFilter, searchTerm]);
 
   return (
     <>
       <MagnifyingGlass glyph="MagnifyingGlass" />
-      <SearchTextInput onKeyDown={onKeyDown} isSearching={isSearching} onChange={onSearchChange} value={searchTerm} />
+      <SearchTextInput ref={searchTextbox} isSearching={isSearching} onChange={onSearchChange} value={searchTerm} />
       {shouldShowGoButton && (
         <GoButton ref={goButton} type="submit" aria-label="Go" href={searchUrl}>
           <GoIcon glyph="ArrowRight" fill="#13AA52" />
--- a/src/components/Searchbar/ExpandedSearchbar.js
+++ b/src/components/Searchbar/ExpandedSearchbar.js
@@ -10,7 +10,6 @@ import SearchTextInput from './SearchTextInput';
 import { SearchResultLink } from './SearchResult';
 import { searchParamsToURL } from '../../utils/search-params-to-url';
 import SearchContext from './SearchContext';
-import { SearchResultsContainer } from './SearchDropdown';
 
 const ARROW_DOWN_KEY = 40;
 const CLOSE_BUTTON_SIZE = theme.size.medium;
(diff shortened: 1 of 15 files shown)


src/components/Searchbar/Searchbar.js
```diff
@@ -69,9 +69,11 @@ const Searchbar = ({ getResultsFromJSON, isExpanded, setIsExpanded, searchParams
   // Focus Handlers
   const onExpand = useCallback(() => setIsExpanded(true), [setIsExpanded]);
   const onFocus = useCallback(() => {
+    if (!isFocused) {
```
jestapinski, line 72: I found this bug for reporting also since toggling hits the focus method

src/components/Searchbar/SearchResult.js
```diff
@@ -101,12 +106,64 @@ const sanitizePreviewHtml = text =>
...
     const highlightedPreviewText = highlightSearchTerm(preview, searchTerm);
+    const resultLinkRef = useRef(null);
+
+    const onArrowDown = resultLinkRef => {
+      const nextSibling = getNestedValue(['current', 'nextSibling'], resultLinkRef);
+      if (nextSibling) {
+        nextSibling.focus();
+      } else {
+        // This is the last result, so let's loop back to the top
+        // TODO fix here
```
jestapinski, line 131: Found a bug here, will remove WIP when ready for review

src/components/Searchbar/ExpandedSearchbar.js
```diff
@@ -87,15 +88,26 @@ const ExpandedSearchbar = ({ isFocused, onChange, onMobileClose }) => {
...
+    e => {
+      // On an "Enter", click the Go button
+      if (e.key === 'Enter' || e.keyCode === ENTER_KEY) {
+        goButton && goButton.current && goButton.current.click();
+      } else if (e.key === 'ArrowDown' || e.keyCode === ARROW_DOWN_KEY) {
+        // prevent scrolldown
+        e.preventDefault();
+        // find first result in the dropdown and focus
+        if (searchContainerRef && searchContainerRef.current) {
+          searchContainerRef.current.querySelector(`${SearchResultLink}`).focus();
```
jestapinski, line 103: This was the previous issue. We were querying the entire document for search results, which would break on the search results page 🤦 . Now we added a ref to the searchbar to the context so we can limit the query to just the searchbar.

This PR is again ready for review! Staging has been updated

sophstad:
I think LeafyGreen released an update for all of their components to fix SSR on Friday — do you think it'd be worth using those hooks?

jestapinski:
> I think LeafyGreen released an update for all of their components to fix SSR on Friday — do you think it'd be worth using their hooks?

Sure, great idea! I'll try to use [`useEventListener`](https://github.com/mongodb/leafygreen-ui/blob/master/packages/hooks/src/useEventListener.ts) and see what that gets us

jestapinski:
An update on the hook. I was able to get it going pretty easily for the top searchbar. I started running into trouble when trying to add it to the search results.

The main issue I was seeing was the event went off once for each search result (so a key down would run 3 times and we would end up where we started). I figured this is due to something about how the element was bound with `addEventListener` but was hitting a wall here.

I pushed up the change for the top searchbar so we can take a look at that but curious what you think about the results situation? I'll point it out in the code

src/components/Searchbar/SearchResult.js
```diff
@@ -102,11 +107,58 @@ const sanitizePreviewHtml = text =>
...
+            // find previous result and focus
+            onArrowUp(resultLinkRef);
+          }
+        }
+      },
+      [onArrowDown, searchContainerRef]
+    );
     return (
-      <SearchResultLink href={url} onClick={onClick} {...props}>
+      <SearchResultLink ref={resultLinkRef} href={url} onClick={onClick} onKeyDown={onKeyDown} {...props}>
```
jestapinski, line 161: This is where I tried using the hook instead of the event handler prop and ran into issues. Here it is clear we are binding to the right place, and I tried using the existing `ref` to control when the hook would fire. For some reason every result will still trigger

src/components/Searchbar/SearchResult.js
```diff
@@ -1,11 +1,16 @@
...
+import React, { useCallback, useContext, useRef } from 'react';
 import sanitizeHtml from 'sanitize-html';
 import { css } from '@emotion/core';
 import styled from '@emotion/styled';
 import { uiColors } from '@leafygreen-ui/palette';
 import { theme } from '../../theme/docsTheme';
+import { getNestedValue } from '../../utils/get-nested-value';
 import SearchContext from './SearchContext';
+import { StyledTextInput } from './SearchTextInput';
+// import { SearchResultsContainer } from './SearchDropdown';
```
sophstad, line 10: [nit] Delete?

jestapinski:
Thanks!!

PR #258, opened 2020-08-19
DOP-1339: Fix Search Results Hover Gaps

jestapinski:
[JIRA](https://jira.mongodb.org/browse/DOP-1339)
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/DOP-1337/landing/jordanstapinski/hover-issue/search?q=stitch) (Search something then check out the results page)

This PR fixes the previous hover issue where placing a cursor between two rows on the search results page would just opaque out all results/ The solution here was to place a transparent pseudoelement over the entire new row height.


jestapinski:
🤙

PR #260, opened 2020-08-21
DOP-1339 Part 3: Search Results Empty State

jestapinski:
[JIRA](https://jira.mongodb.org/browse/DOP-1339)
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/DOP-1337/landing/jordanstapinski/results-empty-state/search)
[InVision](https://mongodb.invisionapp.com/d/#/console/19596736/422509204/preview)

This PR adds the empty state for search results centered vertically on the page. Positioning was a bit wonky so super open to suggestions there.

I spoke to Allison and we will be keeping the footer, so the blank template still is fine.

PR #262, opened 2020-08-24
Add onClick for magnifying glasses

jestapinski:
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/DOP-1337/landing/jordanstapinski/focus-text-search/search)

This PR addresses a comment from #260 involving making the empty state magnifying glass clickable. In addition, we also made the magnifying glass in the expanded searchbar clickable.

I also found a bug where using arrow keys with empty search results threw an error so I added the quick fix for that here as well.


jestapinski:
🔍

PR #268, opened 2020-08-28
DOP-1419: Simplify layout rendering

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOP-1419)] [[Datalake Staging](https://docs-mongodborg-staging.corp.mongodb.com/master/datalake/sophstad/DOP-1419/)] [[Landing Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/landing/sophstad/DOP-1419/)] [[Guides Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/guides/sophstad/DOP-1419/)] [[Drivers Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/drivers/sophstad/DOP-1419/)]
- Simplify rendering of layouts by using [`gatsby-plugin-layout`](https://www.gatsbyjs.com/plugins/gatsby-plugin-layout/). This package ensures that layout components mount only once, allowing components shared between pages to avoid re-rendering.
- Modify `getTemplate()` function to return the template component itself, not a path to the component
- Update TOC to be compatible with new way of rendering pages: now that the TOC is mounted only once, we need to use a `useEffect` hook to tell the TOC to update when we navigate to a new page 
- Rename all instances of `__refDocMapping`/`refDocMapping` to `page` in order to:
  1. Standardize across the repo
  2. Use a more meaningful variable name

During testing, would appreciate a close look to make sure TOC behavior is still working as expected. Thanks 😄


src/components/Footer.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
-import { FeedbackFooter } from './Widgets/FeedbackWidget';
+import loadable from '@loadable/component';
+const FeedbackFooter = loadable(() => import('./Widgets/FeedbackWidget/FeedbackFooter'));
```
mango-db, line 3: What is the reason for this change?

sophstad: Now that all of our template files are rendered on the server (by the new plugin), this component is included in the SSR stage for the first time. However, it has a dependency on `mongodb-stitch-browser-sdk`, which is _not_ SSR compatible and yields an error. To handle this, we use [loadable-components](https://loadable-components.com) to avoid errors in the SSR stage but still allow the component to be used/hydrated by the client.

sophstad:
@mango-db Re: your first comment, these are all part of preview mode. It's not functional anymore and due to be ripped out soon anyway in DOP-1201, so I didn't bother 🤷‍♀️

src/utils/is-active-toc-node.js
```diff
@@ -5,9 +5,6 @@ import { endsWith } from './ends-with';
   has a slug matching the giveen page slug, and false otherwise.
```
mango-db, line 5: Nit: "given"

src/components/TOCNode.js
```diff
@@ -28,6 +28,11 @@ const TOCNode = ({ node, level = BASE_NODE_LEVEL }) => {
 
   const [isOpen, setIsOpen] = useState(isActive);
 
+  // If the active state of this node changes, change the open state to reflect it
+  useEffect(() => {
+    setIsOpen(isActive);
+  }, [isActive, isDrawer ? activeSection : null]); // eslint-disable-line react-hooks/exhaustive-deps
```
mango-db, line 34: How come we have to disable listing here? [This issue comment concerns me](https://github.com/facebook/create-react-app/issues/6880#issuecomment-485912528 )

sophstad: Good question, this approach addresses a pretty specific problem we needed to solve. To clarify, the second dependency (`isDrawer ? activeSection : null`) is the one that causes the warning. The situation that causes it is this:

1. User is on a page
2. User opens a drawer (just a drawer — user does not navigate to a new page)
3. User clicks on a _different_ entry in the TOC and navigates to a new page

When this happens, we need to close the drawer. What stage change specifically should cause the drawer to be closed? A change in `activeSection` (i.e. new page loaded). However, only drawers need to listen for this state change. That's why we use a conditional dependency here: if the node is a drawer, pay attention to the `activeSection`. Otherwise, we don't care about this variable—and listening for it would cause a lot of extra `useEffect` calls—so that dependency is `null`.

src/utils/is-active-toc-node.js
```diff
@@ -2,12 +2,9 @@ import { endsWith } from './ends-with';
 /*
   Provided a current page slug, a slug, and a list of node children, returns
   true if either the given slug matches the current page slug or one of its children
-  has a slug matching the giveen page slug, and false otherwise.
+  has a slug matching the given page slug, and false otherwise.
 */
 export const isActiveTocNode = (currentUrl, slug, children) => {
-  // "/" is a special case usually for titles. In this case we don't want other
```
mango-db, line 8: I'm having trouble deciphering what changed in this PR that allows for the removal of this special case

sophstad:
@mango-db [This prop](https://github.com/mongodb/snooty/pull/268/files#diff-c5b1996366f2c4c341246a79e7025d19L10) was removed from the test (because it's no longer a prop used by the component), and `refDocMapping` represents the entire page AST so... it's long 😅

mango-db:
I'm getting a couple of errors: [https://docs-mongodbcom-staging.corp.mongodb.com/master/drivers/cgoecknerwald/DOP-1419/go/](https://docs-mongodbcom-staging.corp.mongodb.com/master/drivers/cgoecknerwald/DOP-1419/go/)

image
image

src/utils/is-active-toc-node.js
sophstad: I was unable to find a case where this clause was actually entered, so I figured I'd remove it... but I added it back because it's possible I just missed an edge case! Thanks.

PR #278, opened 2020-09-03
Remove withPrefix from search results path

jestapinski:
[Staging Link](https://docs-mongodbcom-staging.corp.mongodb.com/DOP-1337/landing/jordanstapinski/search-path/)

This PR removes the Gatsby `withPrefix` method from the search results link, since we just want to go to `/search...` not `/property/search...`. This fixes the issue for production, but on the staging link removing `withPrefix` gives a 404, so not sure if we care to fix it up for different environments since results point to production anyway (open to a suggestion here).

Actually, getting it to work properly on staging would be difficult, since writers would (most likely) not be working on `landing`, so the results page may not exist anyway.


jestapinski:
Looks like this is now auto-magically redirected, so not sure we need this anymore

sophstad (approved):
@jestapinski We're going to be merging this in and removing the redirects once it's released, as that seems like a more sustainable solution :) thanks again for your help!

PR #284, opened 2020-09-16
DOP-1493: Bump delay to Amplitude to 1s

jestapinski:
[JIRA](https://jira.mongodb.org/browse/DOP-1493)
[Staging Link (Landing)](https://docs-mongodbcom-staging.corp.mongodb.com/DOP-1337/landing/jordanstapinski/DOP-1493/)

Hi!
Now that we can check out metrics in Amplitude, there was an ask to increase the debounce time by which we send events to Amplitude. Previously, whenever we sent a query to Marian, we would also send one to Amplitude (200ms delay). This seemed to log a ton of noise so we want to keep the quick search-as-you-type but just report a bit less often to Amplitude (1000ms).

The solution here was just to split out the `reportAnalytics` logic and debounce it at a different interval.

Check out this piece of art (off of the Network tab in Staging) to see the spacing in events (note there is an 800ms delay between events, but a 1000ms delay for Amplitude overall now and the initial event is at around the 400ms mark):

Screen Shot 2020-09-16 at 10 14 48 AM


eshwer (approved):
Yay great thanks

sophstad (approved):
💯 thanks Jordan!

PR #287, opened 2020-09-22
DOP-1549: Fix version dropdown links

sophstad:
[DOP-1549] [[Charts Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/charts/sophstad/master/)] [[BI Connector Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/bi-connector/sophstad/DOP-1549/)] Use values from the [`git.branches.published` table](https://github.com/mongodb/docs-worker-pool/blob/meta/publishedbranches/docs-charts.yaml#L15-L18)  of the published-branches.yaml file. These are the values that correspond to the git branches of each version, and also to the slug that should appear in the URL. Previously, the component was using the values that appear in the `version.active` table, which understandably worked out badly!

### DIFF
--- a/src/components/VersionDropdown.js
+++ b/src/components/VersionDropdown.js
@@ -8,8 +8,9 @@ import Button from '@leafygreen-ui/button';
 
 const zip = (a, b) => {
   // Zip arrays a and b into an object where a is used for keys and b for values
+  const shorter = a.length > b.length ? b : a;
   const dict = {};
-  a.forEach((key, i) => (dict[key] = b[i]));
+  shorter.forEach((key, i) => (dict[a[i]] = b[i]));
   return dict;
 };
 
@@ -54,7 +55,7 @@ const VersionDropdown = ({
   };
 
   // Zip two sections of data to map git branches to their "pretty" names
-  const gitNamedMapping = zip(gitBranches, published);
+  const gitNamedMapping = zip(gitBranches, active);
   const currentBranch = gitNamedMapping[parserBranch] || parserBranch;
 
   const wrapperRef = useRef(null);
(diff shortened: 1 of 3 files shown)


src/components/VersionDropdown.js
```diff
@@ -110,12 +111,12 @@ const VersionDropdown = ({
       </Button>
       {!hidden && (
         <ul className={['dropdown-menu', dropdownStyles.menu].join(' ')} role="menu">
-          {active.map(version => {
-            const url = normalizePath(`${generatePrefix(version)}/${slug}`);
+          {Object.entries(gitNamedMapping).map(([branch, name]) => {
+            const url = normalizePath(`${generatePrefix(branch)}/${slug}`);
```
schmalliso, line 115: Is the branch and slug backwards? 

The link is `https://docs-mongodbcom-staging.corp.mongodb.com/19.12/charts/sophstad/master/` and I would have expected `/charts/19.12/sophstad`, but I also may be confused.

schmalliso (commented):
One question!

src/components/VersionDropdown.js
sophstad: This is indeed a matter of staging vs. production; for production builds, we use [this logic](https://github.com/mongodb/snooty/pull/287/files#diff-a104f82626180d012641864a0fd0dfebR90-R94):
```js
if (pathPrefix) {
  const [project] = pathPrefix.split('/');
  return `/${project}/${version}`;
}
```

But this led me to notice another bug that could result from `PATH_PREFIX` not including a preceding slash, so I fixed that and added a test to ensure we are generating the correct link!

schmalliso (approved):
Oh of course the slug is there because we flip to the specific page. Thanks for the explain and yay for tests!

PR #290, opened 2020-09-30
Footnote-misc: Reduce likelihood of ID collisions

mango-db:
[JIRA](https://jira.mongodb.org/browse/DOP-1622)

## Use case

Previously, if writer were to use a footnote, e.g. `[#mongodb-compatibility]_`, it would take the ID `mongodb-compatibility`. In that case, if the writer were to also use a header link with the ID `mongodb-compatibility`, the user would experience unexpected behavior when clicking links within the page. 

To reduce the chance of accidental ID collision, I have append `"footnote-"` to footnote IDs, as well as a unique identifier based on names to the footnote reference IDs. 

### Bonus

Handle hanging footnotes (i.e., unpaired footnotes that do not have 1 or more footnote references).

## Staging
[Datalake/index -- with add'l footnotes](https://docs-mongodborg-staging.corp.mongodb.com/DOP-931/datalake/cgoecknerwald/footnote-misc/)
[Datalake/reference -- with add'l footnotes](https://docs-mongodborg-staging.corp.mongodb.com/DOP-931/datalake/cgoecknerwald/footnote-misc/reference)
[Drivers/java](https://docs-mongodborg-staging.corp.mongodb.com/master/drivers/cgoecknerwald/footnote-misc/java/)
[Manual -- #slow-oplogs](https://docs-mongodbcom-staging.corp.mongodb.com/master/manual/cgoecknerwald/footnote-misc/reference/log-messages#footnote-slow-oplogs)

PR #292, opened 2020-10-06
Footnote-context: refresh footnotes before every render

mango-db:
Currently, footnotes were not being updated when navigating between sub-pages. (Example: try navigating from [/index](https://docs-mongodbcom-staging.corp.mongodb.com/footnotes-misc/datalake/cgoecknerwald/footnote-caching/) to [/reference](https://docs-mongodbcom-staging.corp.mongodb.com/footnotes-misc/datalake/cgoecknerwald/footnote-caching/reference) while watching the `footnotes` object logged to console -- see that it doesn't change until you refresh.)

While this change will cause additional calls to `getPageNodes` (formerly `preprocessPageNodes`), this additional computation is necessary to keep footnotes up-to-date while a user navigates.

Note: hanging footnotes are handled in the [footnotes-misc PR](https://github.com/mongodb/snooty/pull/290).

## Staging
- [Datalake /index](https://docs-mongodborg-staging.corp.mongodb.com/footnotes-misc/datalake/cgoecknerwald/footnote-context/)
- [Datalake /reference](https://docs-mongodbcom-staging.corp.mongodb.com/footnotes-misc/datalake/cgoecknerwald/footnote-context/reference)
- [Datalake /administration](https://docs-mongodbcom-staging.corp.mongodb.com/footnotes-misc/datalake/cgoecknerwald/footnote-context/administration)
- [Drivers /java](https://docs-mongodborg-staging.corp.mongodb.com/master/drivers/cgoecknerwald/footnote-context/java)
- [Manual -- #slow-oplogs](https://docs-mongodbcom-staging.corp.mongodb.com/master/manual/cgoecknerwald/footnote-context/reference/log-messages#slow-oplogs)


src/layouts/index.js
```diff
@@ -138,6 +134,8 @@ export default class DefaultLayout extends Component {
 
   render() {
     const { children, pageContext } = this.props;
+    const pageNodes = this.getPageNodes(pageContext);
+    const footnotes = this.getFootnotes(pageNodes);
```
sophstad, line 138: I'm not sure that these operations belong in the [`render()`](https://reactjs.org/docs/react-component.html#render) method, as they are somewhat computationally expensive; have you tried using [`componentDidUpdate()`](https://reactjs.org/docs/react-component.html#componentdidupdate)?

src/layouts/index.js
```diff
@@ -40,15 +38,13 @@ export default class DefaultLayout extends Component {
     }));
   };
 
-  preprocessPageNodes = () => {
-    const { pageContext } = this.props;
+  getPageNodes = pageContext => {
```
sophstad, line 41: What prompted this change? AFAICT, the bug could be fixed by running `preprocessPageNodes` upon route change.

src/layouts/index.js
mango-db: In this case, it seems that the` footnote`s within `pageNodes` are updating between page clicks, and `componentDidUpdate` runs too late in the lifecycle (i.e. after `render`) -- thus this implementation. As discussed, we will change this implementation to instantiate the footnote context provider at the page level rather than the layout level.

mango-db:
Staging has been updated to reflect changes.

src/components/DocumentBody.js
```diff
@@ -1,12 +1,82 @@
...
+  // The nth footnote on a page is associated with the nth reference on the page. Since
+  // anon footnotes and footnote references are anonymous, we assume a 1:1 pairing, and
+  // have no need to query nodes
+  const getAnonymousFootnoteReferences = index => {
+    return `id${index + 1}`;
+  };
+
+  // Standardize cssclass nodes that appear on the page
+  normalizeCssClassNodes(pageNodes, 'name', 'cssclass');
+  const footnotes = getFootnotes(pageNodes);
```
sophstad, line 76: This actually seems like a good opportunity to use a class component: we can call these only once in the `constructor()` and avoid the computational expense of calling them on every re-render, as is happening now.

src/components/DocumentBody.js
```diff
@@ -1,12 +1,82 @@
...
+    return footnoteReferences.filter(node => node.refname === refname).map(node => node.id);
+  };
+
+  // They are used infrequently, but here we match an anonymous footnote to its reference.
+  // The nth footnote on a page is associated with the nth reference on the page. Since
+  // anon footnotes and footnote references are anonymous, we assume a 1:1 pairing, and
+  // have no need to query nodes
+  const getAnonymousFootnoteReferences = index => {
+    return `id${index + 1}`;
+  };
```
sophstad, line 72: I think we can move all of these functions outside the component since they do not rely on any `props` or other variables defined by the class.

sophstad (changes requested):

mango-db:
Staging has been updated to reflect changes

sophstad (approved):

PR #297, opened 2020-10-13
DOP-1635: Fix GTM tracking

sophstad:
[DOP-1635] [[Atlas Staging](https://docs-mongodbcom-staging.corp.mongodb.com/snooty-setup/cloud-docs/sophstad/DOP-1635/)] To verify this fix:

1. Open the staging link and click between some pages
2. Console log the `window.dataLayer` object

You should see lots of `gatsby-route-change` events. These are missing from next-gen properties in production.


src/html.js
```diff
@@ -35,21 +35,9 @@ const HTML = ({ body, bodyAttributes, headComponents, htmlAttributes, preBodyCom
         <link rel="stylesheet" href={withPrefix('docs-tools/mongodb-docs.css')} type="text/css" />
       )}
       <link rel="stylesheet" href={withPrefix('docs-tools/navbar.min.css')} type="text/css" />
-      <script async src={withPrefix('scripts/gtm.js')} />
       {headComponents}
     </head>
     <body {...bodyAttributes}>
-      {/* Google Tag Manager (noscript) */}
```
i80and, line 42: I'm okay with removing the `<noscript>` stanza, but could you just explain the reasoning briefly for posterity?

i80and (approved):

i80and (changes requested):
I'm seeing double-reporting of the send event on initial load, actually?

Screen Shot 2020-10-13 at 12 17 30 PM
screen_shot_2020-10-13_at_12 17 49_pm

src/html.js
sophstad: The Gatsby plugin handles `<noscript>`/`<iframe>` generation itself! See [here](https://github.com/gatsbyjs/gatsby/blob/0ecba231f37a76255333cc9846bcd5b392394ce0/packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js#L11)

PR #299, opened 2020-10-15
DOP-1604: Allow code blocks to render with caption

rayangler:
[[DOP-1604](https://jira.mongodb.org/browse/DOP-1604)]
[[Realms Staging /android/mongodb](https://docs-mongodbcom-staging.corp.mongodb.com/master/realm/raymundrodriguez/DOP-1604/android/mongodb)]

This PR introduces the inclusion of captions to code blocks. `caption` options are handled as strings. Code blocks now use `showWindowChrome`.


src/components/Code.js
```diff
@@ -60,6 +60,17 @@ const Code = ({
...
     >
+      {/* This implementation of caption should change with DOP-1639 */}
+      {caption && (
+        <span
+          className="caption-text"
+          css={css`
+            font-weight: bold;
+          `}
+        >
+          {caption}
```
sophstad, line 71: Just flagging: once [this](https://github.com/mongodb/snooty-parser/pull/242#discussion_r505615068) is implemented, we'll probably need to use a `ComponentFactory` here to render the caption.

src/components/Code.js
```diff
@@ -40,7 +40,7 @@ const getLanguage = lang => {
 };
 
 const Code = ({
-  nodeData: { copyable, emphasize_lines: emphasizeLines, lang, linenos, value },
+  nodeData: { copyable, caption, emphasize_lines: emphasizeLines, lang = 'none', linenos, value },
```
sophstad, line 43: Did you encounter a bug without this default `none` for `lang`? I recently changed this component to handle that in `getLanguage()`, so wondering if you're seeing something weird!

src/components/Code.js
```diff
@@ -60,6 +60,17 @@ const Code = ({
         margin: ${theme.size.default} 0;
       `}
     >
+      {/* This implementation of caption should change with DOP-1639 */}
+      {caption && (
```
sophstad, line 64: The ticket was definitely unclear about this, but we actually want to render the caption as the component's [`chromeTitle`](https://github.com/mongodb/leafygreen-ui/tree/master/packages/code#properties)`, if it exists!

src/components/Code.js
```diff
@@ -72,6 +83,7 @@ const Code = ({
         highlightLines={emphasizeLines}
         language={getLanguage(lang)}
         showLineNumbers={linenos}
+        showWindowChrome={true}
```
sophstad, line 86: We only want this property to be `true` if a caption is specified.

sophstad (changes requested):
The A/C for this ticket definitely could have been more clear, but the goal is to show the caption in LeafyGreen's "chrome window" text, if it's specified. But so far so good—lemme know if I can clarify anything for you!

src/components/Code.js
rayangler: Ah, my mistake! I accidentally left `lang = 'none'` because of a merge conflict. I'll fix this

src/components/Code.js
rayangler: Oh, gotcha. Will change

rayangler:
With the recent [comment](https://jira.mongodb.org/browse/DOP-1639?focusedCommentId=3448465&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-3448465) on DOP-1639, it seems like we do not need to worry about breaking the caption down into individual components. Captions are just strings now.

sophstad (approved):

PR #304, opened 2020-10-21
DOP-1667: Fix root and step components

sophstad:
[DOP-1667] [[BI Connector Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/bi-connector/sophstad/DOP-1667/)]
- Render children of `root` nodes
- Render children of `steps` nodes
  - Simplify `Include` component, which used to handle numbering of `Step` components


i80and (approved):
This strikes me as cleaner, too! 🚀

PR #303, opened 2020-10-21
Move page-specific components to DocumentBody

sophstad:
[[Atlas Staging](https://docs-mongodbcom-staging.corp.mongodb.com/snooty-setup/cloud-docs/sophstad/page-layout/)] [[BI Connector Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/bi-connector/sophstad/page-layout/)] [[Drivers Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/drivers/sophstad/page-layout/)]

PR #306, opened 2020-10-22
Fix feedback context error

sophstad:
[[BI Connector Staging](https://docs-mongodborg-staging.corp.mongodb.com/master/bi-connector/sophstad/footnote-fix/)] Move `Footer` component into `DocumentBody` component to ensure it is always a child of `FeedbackProvider`. Fixes error introduced in #303

### DIFF
--- a/src/components/DocumentBody.js
+++ b/src/components/DocumentBody.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import Helmet from 'react-helmet';
 import ComponentFactory from './ComponentFactory';
 import FootnoteContext from './footnote-context';
+import Footer from './Footer';
 import Widgets from './Widgets';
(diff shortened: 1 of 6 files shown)


src/components/DocumentBody.js
```diff
@@ -85,7 +86,8 @@ export default class DocumentBody extends Component {
 
   render() {
     const {
-      pageContext: { location, metadata, page, slug },
+      location,
```
rayangler, line 89: Why did location have to be moved out of page context?

rayangler (approved):
Everything looks good! Just a question for my understanding!

src/components/DocumentBody.js
sophstad: It was an error 😅 Gatsby passes that value outside of `pageContext`.

PR #308, opened 2020-10-26
DOP-1673: Add Segment analytics to Breadcrumbs

mango-db:
[DOP-1673](https://jira.mongodb.org/browse/DOP-1673). [Datalake](https://docs-mongodbcom-staging.corp.mongodb.com/master/datalake/cgoecknerwald/DOP-1673/). [Manual](https://docs-mongodbcom-staging.corp.mongodb.com/master/manual/cgoecknerwald/DOP-1673/aggregation).

Bonus: [Issue referenced](https://github.com/gatsbyjs/gatsby/issues/8357) in comment is fixed. Remove comment.


allisonmui9 (approved):

src/components/Breadcrumbs.js
```diff
@@ -12,9 +13,18 @@ const Breadcrumbs = ({ parentPaths, slugTitleMapping }) => (
...
             <li key={path}>
-              {/* TODO: Replace <a> with <Link> when back button behavior is fixed for the component.
-              GitHub issue: https://github.com/gatsbyjs/gatsby/issues/8357 */}
-              <Link to={path}>{formatText(title)}</Link>
+              <Link
+                to={path}
+                onClick={() => {
+                  reportAnalytics('BreadcrumbClick', {
+                    areaFrom: 'Breadcrumb',
+                    parentPaths: parentPaths,
```
eshwer, line 21: Does 'Breadcrumb' return the Breadcrumb they selected? or is that 'path'
And does 'ParentPaths' return the list of breadcrumbs they could have selected?

mango-db: Supposing the breadcrumbs: 'Java 1' > 'Java 2' > 'Java 3'. If the user selects 'Java 2', Segment will report the following:

```
areaFrom: 'Breadcrumb',
parentPaths: {
    0: 'Java 1',
    1: 'Java 2',
    2: 'Java 3',
}
breadcrumbClicked: 'Java 2'
```

mango-db: You can test for yourself [here](https://docs-mongodbcom-staging.corp.mongodb.com/master/manual/cgoecknerwald/DOP-1673/core/aggregation-pipeline-optimization)!

eshwer: Oh that's great! I'm not sure if we need the `areaFrom: 'Breadcrumb',` in this instance since it will always be "breadcrumb" and the title of the analytics implies its from the breadcrumb. Unless it would make more sense for `areaFrom` to return the URL. Either way, I think this will certainly give us what we need to figure out how to do our breadcrumbs in the design.

mango-db: I just added it to mirror the ['areaFrom' pattern](https://github.com/mongodb/snooty/search?q=areaFrom) seen in other Segment tracking -- can certainly remove if desired, just lmk.

eshwer: Hmm I'll deflect to @sophstad to weigh in here on best practices

sophstad: It looks like `areaFrom` is used in those cases when the same event is triggered by different components—for example, whether the user clicks on the searchbar in the top nav bar or on the search page.

As such, I don't think it is necessary to include here.

eshwer (approved):

PR #314, opened 2020-11-05
Remove segment script

sophstad:
[[Datalake Staging](https://docs-mongodbcom-staging.corp.mongodb.com/DOP-1643/datalake/sophstad/segment-fix/)] Get rid of `Segment snippet included twice.` error as thrown by the GA plugin. GTM handles loading of Segment on its own.


mango-db (approved):
LGTM!

PR #315, opened 2020-11-05
DOP-1701: Update image writes

sophstad:
[DOP-1701] [[Charts Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/charts/sophstad/DOP-1701/)] Write a file to the filesystem for each path used to refer to a given image.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -33,7 +33,7 @@ let RESOLVED_REF_DOC_MAPPING = {};
 // stich client connection
 let stitchClient;
 
-const assets = {};
+const assets = new Map();
 
 exports.sourceNodes = async () => {
   // setup env variables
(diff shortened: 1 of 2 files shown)


gatsby-node.js
```diff
@@ -33,7 +33,7 @@ let RESOLVED_REF_DOC_MAPPING = {};
 // stich client connection
 let stitchClient;
 
-const assets = [];
+const assets = {};
```
i80and, line 36: Might I suggest a Map here? It's more idiomatic in ES6+ for this type of application, and there's fewer quirks (like the "in" operator scanning the prototype)

gatsby-node.js
```diff
@@ -64,7 +64,15 @@ exports.sourceNodes = async () => {
     const pageNode = getNestedValue(['ast', 'children'], val);
     const filename = getNestedValue(['filename'], val) || '';
     if (pageNode) {
-      assets.push(...val.static_assets);
+      val.static_assets.forEach(asset => {
+        const checksum = asset.checksum;
+        if (checksum in assets) {
```
i80and, line 69: I don't see it being hit easily in this case, but the `in` operator has a problem:

```
> "toLocaleString" in {}
true
```

I recommend Maps

i80and (commented):
One remark on object quirkiness -- I suggest this change, but your call on if it's important enough to push this merge back

i80and (approved):

PR #317, opened 2020-11-10
DOP-1545: Build 404 page

rayangler:
[[JIRA Ticket](https://jira.mongodb.org/browse/DOP-1545)]
[[Landing Staging: /404](https://docs-mongodbcom-staging.corp.mongodb.com/master/landing/raymundrodriguez/DOP-1545/404/)]

This PR introduces the creation of a 404 page for docs-landing, found at the slug `/404`.


src/components/Footer.js
```diff
@@ -1,16 +1,31 @@
...
-    <FeedbackFooter />
-    <div className="footer">
-      <div className="copyright">
-        <p>© MongoDB, Inc 2008-present. MongoDB, Mongo, and the leaf logo are registered trademarks of MongoDB, Inc.</p>
+const Footer = props => {
+  const { enableFeedback } = props;
+
+  return (
+    <>
+      {enableFeedback && <FeedbackFooter />}
```
rayangler, line 11: Not sure if we needed feedback to be available on a 404 page so this was my idea of bypassing this. If there's a better way to do this or if we do need feedback, I can go ahead and change this!

src/components/NotFound.js
```diff
@@ -0,0 +1,110 @@
...
+  padding: 0px ${theme.size.small};
+
+  @media ${theme.screenSize.upToSmall} {
+    padding: 0px ${theme.size.default};
+  }
+`;
+
+const supportButtonColor = '#3D4F58';
+
+const SupportButton = styled('a')`
```
sophstad, line 19: [nit] I think this would be clearer if it were named `SupportLink`, since we aren't actually intending for it to function as a button.

gatsby-node.js
```diff
@@ -114,10 +114,38 @@ exports.createPages = async ({ actions }) => {
...
+    }
+
     resolve();
   });
 };
 
+exports.onCreatePage = ({ page, actions }) => {
+  const { createPage } = actions;
+  // Allows 404 page to be visible in development.
+  if (page.path === '/dev-404-page/' && process.env.GATSBY_SITE === 'landing') {
```
sophstad, line 143: Interesting, I don't see `dev-404-page` anywhere else in this PR—is it a Gatsby pattern?

sophstad (changes requested):
Great work Raymund! I left a few comments.

Also, not sure if you've been run this past @allisonmui9, but you can send the staged site over/add her as a reviewer on here (mostly thinking of the mobile 404 page since that wasn't included in the designs) [but thanks for going the extra mile to ensure that this looks good on small screens as well!].

src/components/Footer.js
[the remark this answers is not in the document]
rayangler: Haha, good call! I originally called it `disableFeedback` but was concerned about if it would need default props or not so I turned it into `enableFeedback` as a "safer" bet. Will change!

gatsby-node.js
rayangler: `dev-404-page` is Gatsby's default 404 page when in development. If you go to any non-existent page while in development, such as `/asdf`, it will show you the 404 page (which for some reason is named `dev-404-page`). Alternatively, you can reach the same page by going directly to `/404`.

gatsby-node.js
```diff
@@ -114,10 +114,43 @@ exports.createPages = async ({ actions }) => {
...
+    page.context = {
+      layout: '404',
+      page: {
+        ast: {},
+      },
+    };
+    createPage(page);
+  }
+};
```
sophstad, line 152: For the sake of keeping code clean in production, I think it may be good to remove this—assuming you can still develop for the page by going to `/404`. Correct me if this isn't the case!

sophstad (approved):
Just one request. Thanks, Raymund!

gatsby-node.js
rayangler: Yes, that is correct! Done!

PR #327, opened 2020-12-14
DOP-1748: OpenAPI component updates page hash

sophstad:
[DOP-1748] [[Realm Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/realm/sophstad/DOP-1748/admin/api/v3#realm-administrator-apis)]
- Update page hash when an OpenAPI card is expanded
- If URL includes page hash on load, automatically expand this card

@nlarew Added you as a reviewer for acceptance testing! Thanks in advance 😄


nlarew:
@sophstad LGTM on those two bullets! Thank you!

I'm still seeing one more minor issue - `:ref:` roles that point to a given endpoint don't actually point to the endpoint when resolved.

Not sure if it's something we can handle on the frontend or if it would require a parser update too.

e.g. consider this rst:

```
Application ID
~~~~~~~~~~~~~~

To find your Application ID, make a request to the {+service-short+}
:ref:`get-/groups/{groupid}/apps` API endpoint.
```

That ref _does_ render as a link but points to the top of the admin API page (i.e. it doesn't include the ref target in the URL hash)
image

image

nlarew (approved):

mango-db (changes requested):
1. I'm not seeing consistent scroll behavior on smaller screens - the scroll is usually above the card by a significant amount. (Is this possibly a separate issue?)
2. The toggle for `/​auth/​providers/​{provider}/​login` isn't maintained upon refresh
3. Upon refresh, card expansion and scroll causes the excess page state updates - watch the Search box

mango-db (approved):
As discussed, #1 is a separate issue, and we were unable to reproduce #3! LGTM

PR #330, opened 2021-01-11
DOP-1860: Remove .live banner

rayangler:
[[Jira Ticket](https://jira.mongodb.org/browse/DOP-1860)]
[[Atlas Staging](https://jira.mongodb.org/browse/DOP-1860)]

I decided to keep the .live images in case we wanted to easily reuse it in the future, but I'm totally down to just delete it if it's not necessary to keep.

### DIFF
--- a/src/theme/docsTheme.js
+++ b/src/theme/docsTheme.js
@@ -1,10 +1,5 @@
 // Set bannerContent = null when banner should be removed
-const bannerContent = {
-  altText: 'Register Now for MongoDB.live, free & fully virtual',
-  imgPath: 'assets/live-fall.png',
-  mobileImgPath: 'assets/live-fall-mobile.png',
-  url: 'https://www.mongodb.com/live?tck=dotlivefallbannerdocs',
-};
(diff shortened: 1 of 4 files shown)


src/theme/docsTheme.js
```diff
@@ -1,10 +1,5 @@
 // Set bannerContent = null when banner should be removed
```
sophstad, line 1: I should've done this before, but how about a comment here specifying the object structure required to add a banner? We'll make lives easier for our future selves 🙌

sophstad (approved):
Wahoo! 💯

src/theme/docsTheme.js
rayangler: What do you think about the idea of adding an additional bool field `enabled` to `bannerContent`? In places where we check if `bannerContent` is null or not null, we check if `bannerContent.enabled` is `true`. This way, we can keep the banner details and fields up, and just change the bool field when we don't want it anymore.

sophstad (approved):
🎉

PR #334, opened 2021-01-13
DOP-1881: Add canonical link to all pages

sophstad:
[DOP-1881] [[Landing Staging](https://docs-mongodborg-staging.corp.mongodb.com/master/landing/sophstad/DOP-1881/)] [[Realm Staging](https://docs-mongodborg-staging.corp.mongodb.com/master/realm/sophstad/DOP-1881/)] Add `<link rel="canonical" href={url} />` to all pages, where `url` must include a trailing slash.

Sorry this is kinda a big PR! A little bonus cleanup:
- Standardize how the root site url is determined (via [gatsby-config.js](https://www.gatsbyjs.com/docs/reference/config-files/gatsby-config/#sitemetadata)) [note that I don't think it's important to bother incorporating the staging URL here]
- Remove use of `process.env.GATSBY_SITE` throughout repo; replace with `useSiteMetadata()` hook
- Fix active navbar highlighting for some repos
- Change usage of `manual` as project name to `docs` since that's what's in use [here](https://github.com/schmalliso/docs/blob/snooty-setup/snooty.toml#L1) (note, I would love to make this all less manual some day, pun not intended 😭)

From Croud:

> Acceptance Criteria
> 
> - Ensure all canonical tag contain URLs with a trailing slash
> - All canonical URLs must be 200 OK
> - All canonical URLs must be indexable


src/components/DocumentBody.js
```diff
@@ -93,27 +97,50 @@ export default class DocumentBody extends Component {
...
+                user
+              }
+            }
+          }
+        `}
+        render={data => {
+          const {
+            site: { siteMetadata },
+          } = data;
+          const path = normalizePath(`${generatePathPrefix(siteMetadata)}/${slug}`);
```
rayangler, line 121: Question about `generatePathPrefix`: Does the autobuilder handle adding the correct path prefix when building the site for production/deployment? Asking because the full `articleUrl` becomes `https://docs.mongodb.com/master/realm/sophstad/DOP-1881/<slug>` in staging.

rayangler (approved):
LGTM! Just a few questions for my understanding

src/components/VersionDropdown.js
```diff
@@ -82,7 +82,7 @@ const VersionDropdown = ({
 
   const generatePrefix = version => {
     // Manual is a special case because it does not use a path prefix (found at root of docs.mongodb.com)
-    const isManualProduction = project === 'manual' && snootyEnv === 'production';
+    const isManualProduction = project === 'docs' && snootyEnv === 'production';
     if (isManualProduction) {
```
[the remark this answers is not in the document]
sophstad: This is a good point — however, I'm somewhat inclined to say that `isDocsProduction` could be more confusing, because isn't every site really a docs build? I think the variables as they are currently more clearly indicate that we're building the server docs, inconsistent though it feels.

Really I would just like to update the server docs to use `manual` as their key, but that's a whole different job 😀 and thanks a lot for raising this!

src/components/DocumentBody.js
sophstad: Not really: if there is no `pathPrefix` provided (i.e. all staging builds; it's only provided for production), we construct the prefix ourself using a combination of env variables [here](https://github.com/mongodb/snooty/blob/master/src/utils/generate-path-prefix.js#L12). This is definitely Confusing, and something that we could certainly stand to improve!

PR #337, opened 2021-01-22
(DOP-1921): Fix - Feedback widget tooltips appear behind the floating card

nlarew:
https://jira.mongodb.org/browse/DOP-1921

image


src/components/Widgets/FeedbackWidget/components/StarRating.js
```diff
@@ -9,6 +9,15 @@ import loadable from '@loadable/component';
 import { StarIcon } from '../icons';
 const Tooltip = loadable(() => import('./LeafygreenTooltip'));
 
+// Given a string, convert all regular space characters to non-breaking spaces
+function convertSpacesToNbsp(someString) {
```
sophstad, line 13: Any reason not to use [`replace()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace) here? Something like `someString.replace(/\s/g, '&nbsp;')` may be cleaner/faster.

nlarew: Only reason is that I didn't think of it! Good suggestion, I'll update now.

sophstad (approved):
Great find with `usePortal`! I'm glad we didn't have to enter z-index hell 😁 thanks a lot for taking care of this, Nick!

PR #338, opened 2021-01-22
DOP-1896: Render OpenAPI content using Redoc

rayangler:
[[Jira Ticket](https://jira.mongodb.org/browse/DOP-1896)]
[[Atlas Staging - Using Redoc](https://docs-mongodbcom-staging.corp.mongodb.com/DOP-1896/cloud-docs/raymundrodriguez/DOP-1896/testing/)]
[[Realm Staging - Using Snooty](https://docs-mongodbcom-staging.corp.mongodb.com/master/realm/raymundrodriguez/DOP-1896/admin/api/v3/)] (should look the same as the currently deployed site)

This is the basic implementation of Snooty utilizing [Redoc](https://github.com/Redocly/redoc) to render OpenAPI specs. Writers will have the option to render using snooty or Redoc.

Current use of the `openapi` directive should change their rst to:
```
.. openapi:: /file/to/spec.yaml
   :uses-rst:
```

Writers wanting to use the new Redoc renderer should create a separate page that looks something along the lines of:
```
:template: openapi

.. openapi:: /file/to/spec.yaml
```
Alternatively, writers can use a url to a remote spec file as the argument.

This implementation currently uses the open source version of Redoc called Redoc Standalone. There may be paid versions that we can use as needed for increased performance (and as we commit further into using Redoc). (Would need further investigation though)

`redoc` and its peer dependencies have been added to `package.json` as specified by [their docs](https://docs.mongodb.com/drivers/java-drivers/).

This PR should be merged after https://github.com/mongodb/snooty-parser/pull/282 is approved and merged. Will wait to open this PR until the parser PR is approved.


sophstad (commented):
Some minor notes -- will wait until the parser work is merged for full approval!

package.json
```diff
@@ -65,6 +65,7 @@
     "@leafygreen-ui/tooltip": "^6.1.1",
     "@leafygreen-ui/typography": "^7.1.1",
     "@loadable/component": "^5.14.1",
+    "core-js": "^3.8.3",
```
sophstad, line 68: I know this is dependency stuff is confusing but: I don't think we need to include `core-js` here as it is already a dependency of our project (you can find it in `package-lock.json`). I managed to successfully build without adding this line, and I think pinning to a specific version here that we would have to manually upgrade could introduce more conflicts in the future.

sophstad (changes requested):
This looks great! Just one dependency change I think we ought to make.

sophstad (approved):
🎉

PR #341, opened 2021-01-27
DOP-1899: Add Segment tracking on codeblock

rayangler:
[[Jira ticket](https://jira.mongodb.org/browse/DOP-1899)]
[[Realm staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/realm/raymundrodriguez/DOP-1899/deploy/create-with-cli)]

### DIFF
--- a/src/components/Code.js
+++ b/src/components/Code.js
@@ -1,4 +1,4 @@
-import React, { useContext } from 'react';
+import React, { useContext, useCallback } from 'react';
 import ReactDOMServer from 'react-dom/server';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/core';
@@ -13,6 +13,7 @@ import { TabContext } from './tab-context';
 import URIText from './URIWriter/URIText';
 import { isBrowser } from '../utils/is-browser';
 import { theme } from '../theme/docsTheme';
+import { reportAnalytics } from '../utils/report-analytics';
 
 const URI_PLACEHOLDERS = [
   URI_PLACEHOLDER,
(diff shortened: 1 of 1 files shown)


src/components/Code.js
```diff
@@ -84,9 +90,9 @@ const Code = ({
         `}
         highlightLines={emphasizeLines}
         language={getLanguage(lang)}
+        onCopy={reportCodeCopied}
         showLineNumbers={linenos}
         showWindowChrome={caption ? true : false}
-        chromeTitle={caption ? caption : ''}
```
sophstad, line 89: Alphabetization 😊👍

src/components/Code.js
```diff
@@ -54,6 +55,10 @@ const Code = ({
     if (isBrowser) code = htmlDecode(code);
   }
 
+  const reportCodeCopied = useCallback(() => {
+    reportAnalytics('CodeblockCopied', { code });
```
sophstad, line 59: This event was [previously](https://github.com/mongodb/snooty/blob/0c2d67e316910218ed44469191e17a9410525880/src/components/Code.js#L124) named `Codeblock Copied` — nitpicky I know but just want to ensure this is the event name Elle wants! 😄

src/components/Code.js
sophstad, line 89: Alphabetization 😊👍

sophstad (approved):
Just one note, looks great to me!

src/components/Code.js
rayangler: The Jira ticket title stated it should be `CodeblockCopied` so I just went with that

rayangler: @eshwer Do we want the event for this to be `Codeblock Copied` like the previous version, or `CodeblockCopied`?

eshwer: let's do the latter please (no space)

PR #342, opened 2021-01-27
DOP-1943: Fix double active page bug

sophstad:
[DOP-1943] [[Realm Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/realm/sophstad/DOP-1943/tutorial/react-native)] Fix toctree bug where pages that ended with the same string were mistakenly both identified as active.

### DIFF
--- a/src/components/TableOfContents.js
+++ b/src/components/TableOfContents.js
@@ -1,46 +1,33 @@
-import React, { useEffect, useState } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/core';
 import TOCNode from './TOCNode';
 import { TOCContext } from './toc-context';
-import { isBrowser } from '../utils/is-browser';
 
 /**
  * Overall Table of Contents component, which manages open sections as children
  */
-const TableOfContents = ({ height, toctreeData: { children } }) => {
-  // Want to check this on each re-render
-  let currentPage;
-  if (isBrowser) {
-    currentPage = window.location.pathname;
-  }
-
-  const [activeSection, setActiveSection] = useState(currentPage);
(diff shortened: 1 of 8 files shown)


nlarew (approved):
LGTM!

src/components/TableOfContents.js
```diff
@@ -1,46 +1,33 @@
...
-        {children.map(c => {
-          const key = c.slug || c.url;
-          return <TOCNode node={c} key={key} />;
-        })}
-      </ul>
-    </TOCContext.Provider>
-  );
-};
+const TableOfContents = ({ activeSection, height, toctreeData: { children } }) => (
+  <TOCContext.Provider value={{ activeSection }}>
```
rayangler, line 11: Is there a benefit to using React context here? Is it possible to pass `activeSection` directly as a prop to `TOCNode` since it's only one level down? (A benefit of using context here would be to save us from having to pass the prop to `TOCNodes` further down the tree, but wondering if there's anything else I'm missing)

rayangler (approved):
LGTM! Just one question out of curiosity

src/components/TableOfContents.js
nlarew: FWIW I had the same thought when I was browsing through this code earlier. For the current functionality this could definitely just drill props instead of using context. That said, it probably doesn't really matter either way in terms of performance.

sophstad: Yeah, I think the main benefit here is hoisting `activeSection` so that we aren't prop drilling through potentially ♾️ levels of `TOCNodes`. Plus, I have to imagine it's more expensive to pass an identical updated prop to all of those child components than to only update the context.

PR #346, opened 2021-02-01
Fix PageNotFound layout to handle templates properly

rayangler:
[[Docs landing - 404 page](https://docs-mongodbcom-staging.corp.mongodb.com/master/landing/raymundrodriguez/fix-notfound-layout/404/)]

`PageNotFoundLayout` does not utilize changes made to `getTemplate` from a previous PR. This should prevent any awkward templating for the 404 page before the next time `docs-landing` is rebuilt.

Bonus:
* Updated npm packages.


src/layouts/index.js
```diff
@@ -52,7 +50,7 @@ const DefaultLayout = props => {
   const { project } = useSiteMetadata();
 
   if (pageContext.layout === '404') {
-    return PageNotFoundLayout(props);
+    return PageNotFoundLayout(project, props);
```
sophstad, line 53: I supposed I missed this when you added this code, but one possible improvement could be to write this using JSX syntax, which is more in line with our approach in the rest of the repo:

```suggestion
    return <PageNotFoundLayout {...props} project={project} />
```

sophstad (approved):
LGTM, just a suggestion!

sophstad:
No longer applicable due to #354

PR #349, opened 2021-02-05
DOP-1971: Correctly handle mailto links

sophstad:
[DOP-1971] [[Cloud Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/cloud-docs/sophstad/DOP-1971/tutorial/delete-atlas-account/#email-atlas-support-to-finalize-deletion-process)] I couldn't think of any other special-case URL formats besides `mailto:` so I just went with the easiest approach... but if you can think of other use cases I'd be happy to reassess!


schmalliso (approved):
What a delightfully simple patch!

PR #352, opened 2021-02-16
DOP-1838: Update top menu/navbar

rayangler:
[[Jira Ticket](https://jira.mongodb.org/browse/DOP-1838)]
[[Drivers Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/drivers/raymundrodriguez/DOP-1838/)]

**Main changes:**
*New navbar design*
- Navbar should match design (check desktop and mobile sizes).
- Menu button should make sidebar TOC appear. Functionality and layout of sidebar may change when work on the new sidebar is done.
- Search icon/bar should function the same. Changes to its position on mobile were made so that navbar contents are spaced more evenly.

*Removed divs*
- Old nested divs from legacy for both the `document` and `ecosystem-index` layouts have been removed/updated. Staged site and live site should pretty much look the same.

*Removed unused constants*
- Constants in `constants.js` that had no references to them have been removed.


src/components/Navbar.js
```diff
@@ -1,106 +1,71 @@
...
-const getActiveSection = (slug, urlItems) => {
-  const urlMapping = Object.entries(urlItems).find(([, value]) => value.includes(slug));
-  if (urlMapping) {
-    return urlMapping[0];
+const NavbarContainer = styled('div')`
+  align-items: center;
+  background-color: #fff;
+  box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.2);
+  display: flex;
+  height: 45px;
```
sophstad, line 17: As part of this work I think it would really behoove us to move away from a fixed height navbar with a corresponding top margin applied to the content container. This has been a headache in the past as we've seen the navbar cover up content and have to do some extra work to make banners look good. It's probably a good idea to go with a flex box applied to the entire page using `flex-direction: column;` to provide some more, well, flexibility. Devhub is using this approach to great effect [here](https://github.com/mongodb/devhub/blob/staging/src/components/dev-hub/layout.js#L61-L62)!

src/components/Navbar.js
```diff
@@ -1,106 +1,71 @@
...
-
-const NavbarContainer = styled('div')`
-  ${({ isExpanded, shouldOpaqueWhenExpanded }) => isExpanded && shouldOpaqueWhenExpanded && 'opacity: 0.2;'};
-
-  ${theme.bannerContent.enabled
-    ? `top: unset !important;
+const NavbarLeft = styled('div')`
+  align-items: center;
+  display: flex;
+  padding-left: 20px;
```
sophstad, line 36: We can probably apply some standardized x-margins to the `NavbarContainer` to ensure things stay symmetrical. This could also allow us to remove the right padding on the search bar [here](https://github.com/mongodb/snooty/blob/master/src/components/Searchbar/CondensedSearchbar.js#L19).

sophstad (changes requested):
This is great work! I have a few notes mainly focused on avoiding some of the navbar-related headaches I've had to deal with in the past 😅 there is a surprising amount of CSS-related tech debt in these components and you've done a great job of working through it!

src/templates/document.js
```diff
@@ -37,38 +41,39 @@ const Document = ({
...
+        <div className={`left-column ${style.leftColumn} ${renderStatus}`} id="left-column">
+          <Sidebar
+            slug={slug}
+            publishedBranches={publishedBranches}
+            toctreeData={toctree}
+            toggleLeftColumn={toggleLeftColumn}
+          />
+        </div>
+      )}
+      <MainColumn>
```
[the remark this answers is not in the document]
rayangler: The component itself doesn't offer much in terms of logic or functionality but I created it as a separate component to house the `main-column` class css for future use (i.e. after css refactoring). `MainColumn` is also used in `ecosystem-index` [here](https://github.com/mongodb/snooty/pull/352/files#diff-3cc54f2f1eec9f4ea14e7a865956ce36cf1727fae2d655019bae03da8ae26828R9). 

Should I maybe save the css for `main-column` as reusable css and import it to each of the two layouts? Or maybe revert the `MainColumn` used in `ecosystem-index` to just be a `div` with `main-column` class again? I'm interested to get your opinion on this!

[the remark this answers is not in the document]
sophstad: Ah you make a great point about the ecosystem index template (which is [way more hardcoded than it should be, anyway](https://jira.mongodb.org/browse/DOP-929)...). I think therefore your implemented approach makes a lot of sense — as we flesh out the rest of the new nav, we may need to make style changes, at which point having `MainColumn` separated from legacy CSS will come in handy. Thanks for pointing that out!

src/components/Navbar.js
rayangler: Created [DOP-2004](https://jira.mongodb.org/browse/DOP-2004) to handle the rearrangement of components and css for the new grid layout

src/components/Navbar.js
```diff
@@ -1,110 +1,67 @@
...
-        return 'cloud';
-      default:
-        return null;
-    }
-  }
-
-  return null;
-};
+import SidebarMobileMenuButton from './SidebarMobileMenuButton';
+import { Logo } from '@leafygreen-ui/logo';
```
sophstad, line 10: [nit] General rule of thumb is to put external libraries first in the list of imports... I wish our linter caught this! 😤

src/components/Navbar.js
```diff
@@ -1,110 +1,67 @@
...
-      `}
-    >
-      <div
-        css={css`
-          background-image: url(${imgPath});
-          background-position: center;
-          background-size: cover;
-          height: 100%;
+const NavLabel = styled('div')`
+  color: #616161;
```
sophstad, line 48: [nit] Perhaps we should switch to a LeafyGreen color here! https://www.mongodb.design/component/palette/example/

src/components/Searchbar/CondensedSearchbar.js
```diff
@@ -16,7 +16,6 @@ const ExpandButton = styled(IconButton)`
   box-shadow: none;
   height: ${theme.size.large};
   position: absolute;
```
sophstad, line 18: Is it possible to get rid of this altogether now that we are moving towards a non-fixed layout? Rogue `position` rules can always come back to bite ya 😩 or maybe this will become addressable in DOP-2004!

src/components/Searchbar/Searchbar.js
```diff
@@ -48,6 +48,7 @@ const SearchbarContainer = styled('div')`
   @media ${theme.screenSize.upToSmall} {
     ${theme.bannerContent.enabled ? `margin-top` : `top`}: ${SEARCHBAR_HEIGHT_OFFSET};
     height: ${({ isExpanded, isSearching }) => (isExpanded && isSearching ? '100%' : SEARCHBAR_HEIGHT)};
+    ${({ isExpanded }) => !isExpanded && 'position: static;'}
```
sophstad, line 51: Ditto

src/components/SidebarMobileMenuButton.js
```diff
@@ -0,0 +1,42 @@
...
+import IconButton from '@leafygreen-ui/icon-button';
+import { SidebarContext } from './sidebar-context';
+import { theme } from '../theme/docsTheme';
+import { uiColors } from '@leafygreen-ui/palette';
+
+// This container prevents the leafygreen components from flashing when the media query is true
+const MenuButtonContainer = styled('div')`
+  @media ${theme.screenSize.smallAndUp} {
+    display: none;
+  }
```
sophstad, line 13: Can we use the new `display-none.js` util here?

src/templates/document.js
```diff
@@ -36,38 +40,39 @@ const Document = ({
...
-            </div>
-          </div>
+        <div
+          className="body"
+          css={css`
+            margin-left: 25px;
+
+            @media print {
+              margin: 0 !important;
+            }
```
sophstad, line 66: Have you noticed these `print` rules having an impact on render? Perhaps worth making a concerted effort to make sure that printing our pages looks good at the end of this work rather than adding in rules ad hoc now.

sophstad:
One more thing: do we still need [either of these files](https://github.com/mongodb/snooty/blob/master/Makefile#L41-L42)?

src/templates/document.js
rayangler: Good point! I have removed the current print statements for now

src/components/Navbar.js
```diff
@@ -1,110 +1,66 @@
...
-  }`
-    : ''}
+  align-items: center;
+  background-color: #fff;
+  box-shadow: 0 2px 4px 0 rgba(0, 0, 0, ${props => (props.isTransparent ? '0.1' : '0.2')});
+  display: flex;
+  height: 45px;
+  justify-content: space-between;
+  padding: 0px 20px;
+  position: fixed;
```
sophstad, line 21: Another `position` rule 😤 perhaps another rule that will go away with CSS grid?

src/components/Navbar.js
```diff
@@ -1,110 +1,66 @@
...
+  top: 0;
+  width: 100%;
+  z-index: 9999;
```
sophstad, line 24: This is probably something else we should get rid of with grid

src/components/Searchbar/ExpandedSearchbar.js
```diff
@@ -30,6 +30,31 @@ const removeDefaultHoverEffects = css`
   }
 `;
 
+const ExpandedSearchbarContainer = styled('div')`
+  display: flex;
```
sophstad, line 34: Something to keep in mind when you start working on grid stuff: Would it preferable to use a grid approach here too? I am wary of mixing two approaches so at least something to think about!

src/components/Searchbar/SearchTextInput.js
```diff
@@ -10,24 +10,23 @@ import SearchContext from './SearchContext';
 const SEARCHBAR_HEIGHT_OFFSET = '5px';
 
 const activeTextBarStyling = css`
-  background-color: #fff;
-  border: none;
-  color: ${uiColors.gray.dark3};
+  // background-color: #fff;
+  // border: none;
+  // color: ${uiColors.gray.dark3};
```
sophstad, line 15: Intentional?

src/components/Searchbar/SearchTextInput.js
```diff
@@ -10,24 +10,23 @@ import SearchContext from './SearchContext';
...
-    border: none;
-    background-color: ${uiColors.gray.light3};
-    border-radius: ${theme.size.medium};
-    color: ${uiColors.gray.dark1};
-    /* 24 px for magnifying glass plus 16px margin */
-    padding-left: 40px;
-    padding-right: ${theme.size.large};
+    background-color: transparent;
+    border: none !important;
+    // color: ${uiColors.gray.dark1};
```
sophstad, line 22: Ditto

sophstad (approved):
Nice work, LGTM! I left a few comments, mostly just to flag stuff that may be relevant to the grid implementation.

Also, should [this line](https://github.com/mongodb/snooty/blob/master/src/html.js#L37) be deleted?

src/components/Navbar.js
rayangler: Yup, I kept this `position` here to keep the navbar up fixed at the top while scrolling. I hope to remove this when I get the grid layout working properly

src/components/Searchbar/ExpandedSearchbar.js
rayangler: I'm not too sure yet tbh. We could certainly opt to transition into using `grid` instead of `flex` for certain things. However, my understanding was that `flex` was very good for one-dimensional layouts (such as this horizontal searchbar) while `grid` is very useful for two-dimensional layouts.

src/components/Searchbar/SearchTextInput.js
rayangler: Definitely not intentional. Thanks for catching!

rayangler:
> Also, should [this line](https://github.com/mongodb/snooty/blob/master/src/html.js#L37) be deleted?

That line should already have been deleted in one of the more recent [commits](https://github.com/mongodb/snooty/pull/352/files#diff-023eb4be0e5096f5a778ef1dfee421162b529c473ad2c2c7f3e12534eeaa44b7L37)!

PR #354, opened 2021-02-18
DOP-1669: Move to handling 404's as a separate property

i80and:


sophstad (commented):
We had to do some special `PageNotFound` layout stuff ([here](https://github.com/mongodb/snooty/blob/master/src/layouts/index.js#L42-L56)), I think mostly to work around the `404` page not having any page data (which it now does thanks to docs-404). So probably worth deleting this code as well! (@rayangler would know more though)

rayangler:
> We had to do some special `PageNotFound` layout stuff ([here](https://github.com/mongodb/snooty/blob/master/src/layouts/index.js#L42-L56)), I think mostly to work around the `404` page not having any page data (which it now does thanks to docs-404). So probably worth deleting this code as well! @rayangler would know more though.

+1 to deleting this!

src/utils/get-template.js
```diff
@@ -1,4 +1,5 @@
 import { Blank, Document, DriversIndex, Guide, GuidesIndex, Landing, OpenAPITemplate } from '../templates';
+import NotFound from '../components/NotFound';
```
rayangler, line 2: It might be a good idea to move this file from the `components` folder to the `templates` folder. I think I made this into a component before to have `gatsby-node` use it alongside the `landing` template [here](https://github.com/mongodb/snooty/pull/354/files#diff-a6f3904c41211be170d9b0eb87e49ea8d02d0fd5a7224d6772174b425dacf90fL125). My apologies!

sophstad (approved):
😎 woohoo!

rayangler (approved):

PR #355, opened 2021-02-19
DOP-1988: Prevent RightColumn and FeedbackTab from appearing on mobile first

rayangler:
[[Jira Ticket](https://jira.mongodb.org/browse/DOP-1988)]
[[Node Driver Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/node/raymundrodriguez/DOP-1988/)]

**Changes:**

Added css media queries for `RightColumn` and `FeedbackTab` to prevent them from rendering on mobile load.
- On a mobile screen, go to any page on staging to see that neither components load before rendering properly. Desktop screen should load and look the same as deployed site.
- See that Cumulative Layout Shift (CLS) score is generally down (or around the same score depending on page) on staging when compared to deployed site. This can be checked manually by using the `Performance` tab in Chrome Dev Tools while on a mobile responsive screen size, or using the [PageSpeed Insights](https://developers.google.com/speed/pagespeed/insights/) site and looking for the page's CLS. Example: page `/drivers/node/fundamentals/crud/write-operations/change-a-document/` [staging](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fdocs-mongodbcom-staging.corp.mongodb.com%2Fmaster%2Fnode%2Fraymundrodriguez%2FDOP-1988%2Ffundamentals%2Fcrud%2Fwrite-operations%2Fchange-a-document%2F&hl=en) vs. [deployed site](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fdocs.mongodb.com%2Fdrivers%2Fnode%2Ffundamentals%2Fcrud%2Fwrite-operations%2Fchange-a-document%2F&tab=mobile&hl=en).

Fixed small CLS issue with waiting for MongoDB logo to load causes `Documentation` span to shift upon the image loading at times.
* This does not impact the CLS score very much but might as well have included it.

**Note:**

CLS scores may be around the same or slightly higher for some pages on the staging site vs. the deployed site. This is mostly due to the shift caused by the rendering of `FeedbackHeading` and `Contents` (but mostly on `Contents`, since it's often the bigger component). 
- For `Contents`, this may likely be due to the process of looking for heading nodes client-side. This could potentially be fixed by processing headings on the backend instead. 
- Not sure what's causing `FeedbackHeading` to render later than the rest of the content.

### DIFF
--- a/src/components/RightColumn.js
+++ b/src/components/RightColumn.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/core';
+import { theme } from '../theme/docsTheme';
 
 const RightColumn = ({ children }) => (
   <div
@@ -9,6 +10,10 @@ const RightColumn = ({ children }) => (
       margin: 70px 24px 40px 54px;
       min-width: 180px;
       order: 2;
+
+      @media ${theme.screenSize.upToLarge} {
+        display: none;
+      }
     `}
   >
     <div
--- a/src/components/RightColumn.js
+++ b/src/components/RightColumn.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/core';
-import { theme } from '../theme/docsTheme';
+import { displayNone } from '../utils/display-none';
 
 const RightColumn = ({ children }) => (
   <div
@@ -11,9 +11,7 @@ const RightColumn = ({ children }) => (
       min-width: 180px;
       order: 2;
 
-      @media ${theme.screenSize.upToLarge} {
-        display: none;
-      }
+      ${displayNone.onMobileAndTablet};
     `}
   >
     <div
--- a/tests/unit/FeedbackWidget.test.js
+++ b/tests/unit/FeedbackWidget.test.js
@@ -7,6 +7,7 @@ import {
   FeedbackFooter,
 } from '../../src/components/Widgets/FeedbackWidget';
 import { BSON } from 'mongodb-stitch-server-sdk';
+import { matchers } from 'jest-emotion';
 
 import { FEEDBACK_QUALIFIERS_POSITIVE, FEEDBACK_QUALIFIERS_NEGATIVE } from './data/FeedbackWidget';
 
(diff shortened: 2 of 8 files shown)


src/components/RightColumn.js
```diff
@@ -9,6 +10,10 @@ const RightColumn = ({ children }) => (
       margin: 70px 24px 40px 54px;
       min-width: 180px;
       order: 2;
+
+      @media ${theme.screenSize.upToLarge} {
+        display: none;
+      }
```
sophstad, line 16: Just a thought... since we're using this in a few places do you think it could be worth saving it as a util, called like `hideOnMobile` or something?

tests/unit/FeedbackWidget.test.js
```diff
@@ -49,6 +51,8 @@ async function mountFormWithFeedbackState(feedbackState = {}, options = {}) {
   return wrapper;
 }
 
+expect.extend(matchers);
```
sophstad, line 54: Ooooh, what is this?

sophstad (approved):
Nice job Raymund, I _really_ like the approach you took to the util function. The usage looks so clean!

Do you think at some point in the future it'd be wise to get rid of all of our instances of `useScreenSize()`?

tests/unit/FeedbackWidget.test.js
rayangler: It's part of the `jest-emotion` package! It just gives access to more functions such as `toHaveStyleRule` that can help test for css specifically.

Since the `FeedbackTab` will use `display: none` based on a media query now, I figured it would make sense to put it to use [here](https://github.com/mongodb/snooty/pull/355/files#diff-22787a664a953565c4585f32c3629ec0ca0e8ae20145f08605319c88befe8ac1R106-R108) to make sure the css remains what we want it to be. (Plus I took this as an opportunity to test the package out because using `setMobile()` and checking if `display: none` wasn't working; the default display `block` would be returned instead on Friday iirc)

rayangler:
> Do you think at some point in the future it'd be wise to get rid of all of our instances of `useScreenSize()`?

In the future, yes! I wanted to try out removing the hook for these two components at the very least just to make sure that the display logic is what we want it to be and to verify that this helps resolve some CLS issues. If this css seems to work and there aren't any adverse effects (re: SEO, double content, etc.), it might be for the best to remove `useScreenSize`.

PR #360, opened 2021-02-25
Upgrade Husky to v5

sophstad:
Upgrade Husky to v5, which is [reportedly](https://dev.to/typicode/what-s-new-in-husky-5-32g5) smaller and faster! I think the setup is much more confusing though... 🤷‍♀️


.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
rayangler, line 1: What does this mean, exactly?

.husky/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
```
rayangler, line 2: Also wondering what this means?

rayangler (approved):
Changes look fine but just wanted to know what a couple of lines mean. The old setup definitely seemed simpler

.husky/pre-commit
sophstad: Good questions! I set up husky using the `npx husky install` script ([source](https://typicode.github.io/husky/#/?id=manual)). It looks like `.husky/_/husky.sh` is the main script that husky uses to run hooks, and the install script places that file in a directory called `_` ([source](https://github.com/typicode/husky/blob/main/src/commands/install.ts#L25-L35)]). If I had to guess, it's named `_` so that users don't attempt to delete it.

PR #363, opened 2021-02-26
DOP-2004: Apply global grid layout

rayangler:
[[Jira Ticket](https://jira.mongodb.org/browse/DOP-2004)]
Staging links by template:
[[Landing Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/landing/raymundrodriguez/DOP-2004/)]
[[Blank Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/landing/raymundrodriguez/DOP-2004/search?q=test)] - Note: `/search` will not have the sidenav shown for the time-being since we may choose to change the template for the page.
[[Ecosystem-Index - Drivers Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/drivers/raymundrodriguez/DOP-2004/)]
[[Document - Atlas Staging](https://docs-mongodbcom-staging.corp.mongodb.com/snooty-test-24022021/cloud-docs/raymundrodriguez/DOP-2004/)]
[[Openapi - Atlas Staging](https://docs-mongodbcom-staging.corp.mongodb.com/snooty-test-24022021/cloud-docs/raymundrodriguez/DOP-2004/testing)] - Note: Adding the `openapi` template to work with the new grid layout causes Redoc Standalone's sidebar to not update on scroll anymore. This is something that I'd need to spend more time looking into getting working again. However, links to tags and operations still work properly, which is what we really need.
[[NotFound - docs-404 Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/docs-404/raymundrodriguez/DOP-2004/)]

Summary:
* Used `grid` to create top level layout for header (navbar), sidebar, and content. Major templates should generally look the same as the current site.
* Refactored how active heading is chosen for `Contents` using `Intersection Observer API`. Allows `Contents` to work with new layout (and I think it's supposed to be more modern than relying on `scroll` events).


src/layouts/index.js
```diff
@@ -38,30 +23,119 @@ const globalCSS = css`
...
+              <Header />
+              {!isBrowser || showLeftColumn || isMobileMenuOpen ? (
+                <div
+                  className={`left-column ${style.leftColumn} ${renderStatus}`}
+                  css={css`
+                    grid-area: sidebar;
+                    width: 330px;
+                  `}
+                  id="left-column"
+                >
```
sophstad, line 107: Do you think this outer `<div>` can be removed when we start using LG's `SideNav`? (just for my own clarification!)

src/components/MainColumn.js
```diff
@@ -14,9 +14,8 @@ const MainColumn = ({ className, children }) => (
       max-width: 800px;
```
sophstad, line 14: I feel like `display: inline-block;` doesn't make sense here? Perhaps some legacy code we can nix! The `flex-grow`, `justify-content`, and `-ms-flex-*` rules are also maybe vestiges of the flexbox implementation that we can nix now that `MainColumn` is a part of a grid

src/templates/document.js
```diff
@@ -1,61 +1,29 @@
...
-          <Sidebar
-            slug={slug}
-            publishedBranches={publishedBranches}
-            toctreeData={toctree}
-            toggleLeftColumn={toggleLeftColumn}
-          />
-        </div>
-      )}
+    <div className={['content', className].join(' ')}>
       <MainColumn>
```
sophstad, line 26: Just wondering: is there any reason the `.content` classname can't be applied to `MainColumn`? That could knock out a bit of extra cruft.

sophstad (commented):
A few questions! There's a lot to untangle in our layouts, I suppose 😩

src/templates/landing.js
```diff
@@ -6,10 +6,10 @@ import { useTheme } from 'emotion-theming';
 import { uiColors } from '@leafygreen-ui/palette';
 import PropTypes from 'prop-types';
 
-const Wrapper = styled('main')`
```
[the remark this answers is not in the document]
rayangler: Changed this back since this was actually something I left by mistake! I meant to keep the `Wrapper` as `main` and its container as a `div`

src/components/sidebar-context.js
```diff
@@ -1,22 +1,8 @@
-import React, { createContext, useState } from 'react';
+import { createContext } from 'react';
 
 const SidebarContext = createContext({
   isMobileMenuOpen: false,
   setIsMobileMenuOpen: null,
 });
 
-const SidebarContextProvider = ({ children }) => {
```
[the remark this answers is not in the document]
rayangler: I decided that I'd prefer to keep the current implementation as is. I was getting a random error when I tried to do something like you suggested, but I think it also makes sense to keep this as-is since we directly use `isMobileMenuOpen`'s state in `index.js`. This would also follow the pattern that `SearchContext` follows. Open to future changes for conformity though!

src/layouts/index.js
```diff
@@ -38,30 +23,119 @@ const globalCSS = css`
...
     template,
   } = pageContext;
   const Template = getTemplate(project, slug, template);
+
+  const [isMobileMenuOpen, setIsMobileMenuOpen] = useState(false);
+  const { isTabletOrMobile } = useScreenSize();
+  const [showLeftColumn, setShowLeftColumn] = useState(!isTabletOrMobile);
+  // TODO: Check if this styling is still necessary after current Sidebar is replaced with the LG Sidebar
+  /* Add the postRender CSS class without disturbing pre-render functionality */
+  const renderStatus = isBrowser ? style.postRender : '';
```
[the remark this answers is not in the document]
rayangler: I’m not exactly sure. I kept this around when I transferred this code over from [document.js](https://github.com/mongodb/snooty/blob/master/src/templates/document.js#L25). It was brought in to fix the issue seen in [DOP-873](https://jira.mongodb.org/browse/DOP-873).   

Since `isBrowser` would typically be true after being rendered on a browser, it makes sense that this boolean would be used for a styling meant for a post-render. (Removing it didn’t seem to cause any issues though.) My hope was that this styling would be unnecessary with the implementation of the new sidebar and we could just decide to remove it then.

src/layouts/index.js
rayangler: Yes, for sure! It seems like it only exists to contain the current `Sidebar` that we have.

src/templates/document.js
rayangler: Good point! I decided to move the class's styling for font and line height to [body](https://github.com/mongodb/snooty/pull/363/files#diff-fe6b5f1430d78777f1dcebf229294971ed452eb6f104e8d526df0f082fb7a684R18-R20) to allow `MainColumn` and `RightColumn` and any other children of the different templates to utilize that styling. But maybe it's something we should keep to `MainColumn` and add to `RightColumn` specifically? I'd like to get your opinion on this.

rayangler:
Updated the staging links to separate all of the templates!

I decided to pass `className` instead of using `...props` to the templates because (1) we specifically want the `className` for the containing div of each template to be used for the purpose of css anyways, and (2) adding `{...props}` to a `div` like that would cause all props from `layouts/index.js` to be passed unnecessarily.

src/templates/blank.js
```diff
@@ -1,17 +1,26 @@
...
 import PropTypes from 'prop-types';
+import { css } from '@emotion/core';
+import MainColumn from '../components/MainColumn';
+import { TEMPLATE_CLASSNAME } from '../constants';
 import landingStyles from '../styles/landing.module.css';
 
-const Blank = ({ children }) => (
-  <div className="content">
-    <div className={`main-column ${landingStyles.fullWidth}`} id="main-column">
+const Blank = ({ children, className }) => (
```
sophstad, line 8: I recently spoke to Allison M about the [`/legacy` page](https://docs.mongodb.com/legacy/), which uses the `blank` template. I think we want to preserve the total blankness of these pages (i.e. no sidebar). It looks like your staging link's /legacy page still has the sidebar ([here](https://docs.mongodb.com/legacy/)), so would appreciate moving this to total blank behavior! Sorry for the lack of clarity there.

sophstad (changes requested):

src/templates/blank.js
```diff
@@ -1,17 +1,26 @@
...
+  <div
+    className={`${TEMPLATE_CLASSNAME} ${className}`}
+    css={css`
+      display: flex;
```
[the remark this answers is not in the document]
rayangler: This was an attempt at realigning the contents within the `blank` template to be more like how they are on the live site. I'll change this up with some padding/margins instead!

src/templates/blank.js
rayangler: So we want to keep the sidebar for `/search` and not `/legacy`, right?

sophstad: Yeah! I think we ultimately ought to change the `:template:` field for the search page ([source](https://github.com/mongodb/docs-landing/blob/master/source/search.txt#L1)) as it's not truly "blank" anymore. It seems like it's going to be using more or less the `landing` template, don't you think?

(I wish we had a better naming system for our templates 😁)

src/components/Searchbar/ExpandedSearchbar.js
```diff
@@ -34,6 +34,7 @@ const ExpandedSearchbarContainer = styled('div')`
   display: flex;
   background-color: ${uiColors.gray.light3};
   border-radius: ${theme.size.medium};
+  transition: background-color 150ms ease-in;
```
sophstad, line 37: This line applies the fade-in fade-out animations you see when navigating between pages on legacy. It doesn't seem to be working on next gen (and imo feels slightly slow on legacy), so maybe we could just remove it?

src/components/Navbar.js
```diff
@@ -13,14 +15,13 @@ import { searchParamsToURL } from '../utils/search-params-to-url';
...
   display: flex;
   height: 45px;
   justify-content: space-between;
   padding: 0px 20px;
-  position: fixed;
-  top: 0;
-  width: 100%;
+  // Keep relative with z-index so that box shadow will be on top of every component
+  position: relative;
   z-index: 9999;
```
sophstad, line 25: I couldn't find a side-effect of removing this—unless you know of one, let's get rid of it so as to avoid extraneous z-indexes!

src/components/Navbar.js
```diff
@@ -13,14 +15,13 @@ import { searchParamsToURL } from '../utils/search-params-to-url';
...
+  box-shadow: 0 2px 4px 0 rgba(0, 0, 0, ${(props) => (props.isTransparent ? '0.1' : '0.2')});
   display: flex;
   height: 45px;
   justify-content: space-between;
   padding: 0px 20px;
-  position: fixed;
-  top: 0;
-  width: 100%;
+  // Keep relative with z-index so that box shadow will be on top of every component
+  position: relative;
```
sophstad, line 24: Interesting, where did you notice the shadow issue? I can't reproduce.

src/components/Navbar.js
```diff
@@ -107,4 +109,8 @@ const Navbar = () => {
   );
 };
 
+Navbar.propTypes = {
+  isBlankTemplate: PropTypes.bool,
```
sophstad, line 113: Did you mean `isSidebarEnabled`?

src/components/SidebarMobileMenuButton.js
```diff
@@ -23,16 +24,16 @@ const MenuButton = styled(IconButton)`
...
-    setIsMobileMenuOpen(state => !state);
-  }, [setIsMobileMenuOpen]);
+    setIsSidebarMenuOpen((state) => !state);
+  }, [setIsSidebarMenuOpen]);
 
   return (
     <MenuButtonContainer className={className}>
       <MenuButton aria-label="View All Products" onClick={clickMenu}>
-        <Icon glyph={isMobileMenuOpen ? 'X' : 'Menu'} size="large" />
+        <Icon glyph={isSidebarMenuOpen && isBrowser ? 'X' : 'Menu'} size="large" />
```
sophstad, line 36: Interesting, did this fail without an `isBrowser` check? That seems unexpected!

src/components/contents-context.js
```diff
@@ -55,62 +54,64 @@ const findSectionHeadings = (nodes) => {
...
-      let bestMatch = [Infinity, 0];
-
-      headingPositions.forEach((headingPosition, index) => {
-        const delta = Math.abs(headingPosition - currentPosition);
-        if (delta <= bestMatch[0]) {
-          bestMatch = [delta, index];
+    // Callback is first performed upon page load. Ignore checking entries on first load to allow
+    // headings[0] to be counted first when headings[0] and headings[1] are both intersecting
+    let firstLoad = true;
+    let defaultActiveSectionId = window.location.hash.slice(1) || headingNodes[0]?.id;
```
sophstad, line 91: [nit] The ultimate safest way to do this would be:
```suggestion
    let defaultActiveSectionId = window.location.hash.slice(1) || headingNodes?.[0]?.id;
```

src/layouts/index.js
```diff
@@ -38,30 +25,72 @@ const globalCSS = css`
...
   const { project } = useSiteMetadata();
-
   const {
     metadata: { title },
     page,
     slug,
     template,
   } = pageContext;
   const Template = getTemplate(project, slug, template);
+  const isSidebarEnabled = template !== 'blank';
```
sophstad, line 70: Interesting approach... I like how this allows us to not repeat the sidebar code in each template, but also worry that perhaps it obfuscates the actual behavior of each template (i.e. what's included in the template file is not necessarily what you get via render). What was your thinking here?

sophstad (changes requested):

src/components/Searchbar/ExpandedSearchbar.js
rayangler: This transition is for the `ExpandedSearchbar` in particular. It's used when hovering over the searchbar itself. I realized that I needed to move where the transition [previously was](https://github.com/mongodb/snooty/pull/363/files/670c9174bbae7a54ee92ee5c2b37213e51f859d6#diff-fc8ec9a68dc65b81d0503a31a47c094dd2f7c0a57fcd8fc2d89681d4b042713eL18) after the previous `docs-nav` ticket rearranged some searchbar stuff. The change is subtle but I decided to just keep it anyways.

Could [this](https://github.com/mongodb/snooty/pull/363/files/670c9174bbae7a54ee92ee5c2b37213e51f859d6#diff-e79e4c2c4c39a8e639b6c2f0d78c7717e5d614c1aa095500ed80e02d223a3ba7L18) be the transition you were referring to?

src/components/Navbar.js
rayangler: If you remove the `z-index` and/or `position: relative`, the box shadow of the navbar can sometimes be covered by other components while scrolling down. This is most noticeable with LG components such as a code block. Here is an example page: ([link](https://docs-mongodbcom-staging.corp.mongodb.com/master/drivers/raymundrodriguez/DOP-2004/c/)).

src/components/Navbar.js
rayangler: Please refer to the [other comment](https://github.com/mongodb/snooty/pull/363#discussion_r592440008) pertaining to the box shadow!

src/components/Navbar.js
rayangler: Thanks for catching this! I meant to remove this after introducing `isSidebarEnabled` to this component as part of context

src/components/SidebarMobileMenuButton.js
rayangler: In an attempt to [simplify](https://github.com/mongodb/snooty/pull/363/files/670c9174bbae7a54ee92ee5c2b37213e51f859d6#diff-a5373604f289396d47dac4670c26c970cea3ab77a0e9dd24c78c82b5484fa798R31) the [previous logic](https://github.com/mongodb/snooty/pull/363/files/670c9174bbae7a54ee92ee5c2b37213e51f859d6#diff-4ed84ff2669e916bd0ac773815d1678118db7bdc226291c85a0e7c4befa038abL43) for showing the sidenav, this causes a side-effect due to the `useScreenSize` we use to determine whether to show the sidenav by default or not on desktop vs. mobile. Without checking to see if `isBrowser` now, the glyph ends up flashing as `X` before becoming `Menu`.

I wonder if it might even be better to just return the logic the way it was since `isMobileMenuOpen` didn't need to depend on `useScreenSize` previously.

src/components/contents-context.js
rayangler: Despite being the ultimate safest way to do this, I wonder if this is necessary. `headingNodes` should be expected to exist because we get it from [`findHeadings`](https://github.com/mongodb/snooty/pull/363/files/670c9174bbae7a54ee92ee5c2b37213e51f859d6#diff-b45effd2f2fad9df3bc7f776ee52abfa2b52a8f9a73ea8648d4ba355fd383a7fR13), which at the very least should return an empty array.

src/layouts/index.js
rayangler: My thinking was to follow a pattern of having the content (template), the navbar, and the sidenav in the same layout since that is what was expected of the high level grid layout we have/would expect for the site. I was also thinking that if we wanted to expand disabling the sidebar for more templates in the future, we could have some sort of array to keep track of them all.

I saw the template as more of a template for the actual content on the page as opposed to the overall layout of the page. I'm not opposed to moving the sidebar/sidenav component into each template though. It was more of a choice based on how I saw layout vs. template 😅

src/components/Searchbar/ExpandedSearchbar.js
sophstad: Ah yep totally! My bad!

src/components/Navbar.js
sophstad: Ah I see now 😭 good catch!

sophstad (approved):
LGTM! I'm happy to go with this approach for templates, we can always change it in the future if need be. Thanks Raymund!!

PR #371, opened 2021-03-17
DOP-1842: Add Banner component

rayangler:
[DOP-1842]
[[Drivers Staging Link](https://docs-mongodborg-staging.corp.mongodb.com/master/drivers/raymundrodriguez/DOP-1842/)]

* Added `Banner` component to fetch data from Realm/Stitch.
* Updated `utils/stitch.js` to use `getStitchClient`.
* Updated `docs-tools` to actually use the same version snooty uses on the `main` branch.
* Removed unused widgets.

On Realm:

* Added `banner` value with banner content schema. I included `isEnabled` in hopes that it would be a minor convenience so that when changing/removing the banner, we could just change the flag instead of changing the function that uses this object.
* Added `getBanner` function to return banner content if enabled.
* Added `live-fall` banners to host under the `bnrs/` folder to avoid issues with adblock when images were saved under `banners/`.

Feel free to toggle the `isEnabled` bool in the `banners` value on Realm and deploy to update whether it shows on the staging site.


src/components/Searchbar/Searchbar.js
```diff
@@ -20,10 +21,10 @@ const SEARCHBAR_HEIGHT = '36px';
 const SEARCHBAR_HEIGHT_OFFSET = '5px';
 const TRANSITION_SPEED = '150ms';
 
-const expandedCss = css`
+const expandedCss = (isBannerEnabled) => css`
   position: fixed;
   right: 16px;
-  top: 5px;
+  top: calc(${isBannerEnabled && `${theme.header.bannerHeight} +`} ${SEARCHBAR_HEIGHT_OFFSET});
```
sophstad, line 27: Something smells fishy here—do we still need the fixed positioning for search? Can its position just be dependent on that of the navbar?

src/components/Header.js
```diff
@@ -9,8 +29,8 @@ const Header = () => {
         grid-area: header;
```
sophstad, line 29: Could we not apply the z-index rule here instead? Or maybe that would reintroduce the modal bug you mentioned yesterday! 🤔

src/components/Banner.js
```diff
@@ -0,0 +1,60 @@
...
+  if (bannerContent == null) {
+    return null;
+  }
+
+  return (
+    <a
+      css={css`
+        display: block;
+        height: ${theme.header.bannerHeight};
+        width: 100vw;
```
sophstad, line 38: [nit] This is just personal preference, but... I kinda prefer Emotion's `styled` API over the `css` prop 🙊 If you don't have any strong feelings I'd rather use that! I think it helps keep the JSX neater.

src/components/Navbar.js
```diff
@@ -54,7 +51,7 @@ const NavLabel = styled('div')`
   user-select: none;
 `;
 
-const Navbar = () => {
+const Navbar = ({ ...props }) => {
```
sophstad, line 54: [nit] `{...props}` is equivalent to `props` 😁

src/components/SearchResults/SearchResults.js
```diff
@@ -37,7 +37,7 @@ const EmptyResultsContainer = styled('div')`
   must account for the navbar, any margins added from using the blank landing
   template, and half of the height of the empty state component */
   margin-top: calc(
-    50vh - ${theme.navbar.height} - ${LANDING_MODULE_MARGIN} - ${LANDING_PAGE_MARGIN} - ${EMPTY_STATE_HEIGHT} / 2
+    50vh - ${theme.header.navbarHeight} - ${LANDING_MODULE_MARGIN} - ${LANDING_PAGE_MARGIN} - ${EMPTY_STATE_HEIGHT} / 2
```
sophstad, line 40: Same thoughts as above with regard to the search bar. It would be nice to untangle this now that we don't have to deal with the awkward positioning rules of the legacy navbar.

sophstad (changes requested):

src/components/Searchbar/Searchbar.js
rayangler: `position: fixed` has remained to allow the searchbar to expand up to its [assigned width](https://github.com/mongodb/snooty/blob/master/src/components/Searchbar/Searchbar.js#L18) on smaller screens. Removing this will truncate or cut off the right side of the sidebar when below 600px of screen width. The `fixed` position also allows the searchbar to show expanded on top of the navbar on mobile. 

Maybe the mobile fix can be to not render anything else on the navbar, or displaying them as `none` when the searchbar is expanded on mobile. I'm not so sure how we could get the expanded searchbar to appear above the logo on the nav / other navbar elements if we were to add them.

src/components/Header.js
rayangler: Applying `z-index` to either the `Header` or the `Banner` component doesn't really matter for that particular bug afaik. However, placing the rule in the `Header` would cause the `SearchDropdown` to appear behind the navbar. Normally, it would be expected to appear directly behind the searchbar, but above the navbar.

One fix for the dropdown issue that would allow us to apply the `z-index` to the `Header` instead of each of its children would be to apply `position: fixed` to the searchbar, regardless of screen size (since the searchbar is currently only set to be `fixed` when it is expanded when the screen size is smaller than 670px).

src/components/SearchResults/SearchResults.js
rayangler: Ah, I believe this css is used in [`/search`](https://docs.mongodb.com/search/) to get empty search results rendered onto the middle of the page. I could try to recreate this through alignment instead.

src/components/Navbar.js
```diff
@@ -19,9 +19,7 @@ const NavbarContainer = styled('div')`
   height: 45px;
   justify-content: space-between;
   padding: 0px 20px;
-  // Keep relative with z-index so that box shadow will be on top of every component
   position: relative;
```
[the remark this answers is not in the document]
rayangler: The purpose of this `position` is so that the `Seachbar` has a parent to be [absolutely positioned](https://github.com/mongodb/snooty/pull/371/files#diff-836ccf349cba8d638c65003b0497966e57eab394d6f630b9bac900a7fca4a7b4R24) to when expanded at screen sizes smaller than 670px. This allows the searchbar to appear stacked on top of its siblings on the navbar without it having to be positioned as `fixed` and relying on the navbar + banner height for placement as it did [before](https://github.com/mongodb/snooty/pull/371/files/f81df7828d117680e70e390862a8bbf699ab7448#diff-836ccf349cba8d638c65003b0497966e57eab394d6f630b9bac900a7fca4a7b4R27)

sophstad (approved):
Looks great!!

PR #377, opened 2021-04-01
DOP-1947: Anchor scroll to clicked tab component

casthewiz:
### Stories/Links:

DOP-1947

### Staging Links:

https://docs-mongodbcom-staging.corp.mongodb.com/master/realm/cassidy.schaufele/DOP-1947/sdk/ios/data-types/embedded-objects/

### Notes:
Replicates some of the logic from here: https://github.com/mongodb/docs-tools/blob/master/themes/mongodb/src/js/componentTabs.js#L94-L133

Also adds a minor delay on scroll to account for renders in anything dependent on `TabContext` propagation.

Utilizes an empty `<div/>` tag with a ref to anchor our scroll logic to the parent tabset clicked.

### DIFF
--- a/src/components/Tabs.js
+++ b/src/components/Tabs.js
@@ -61,6 +61,7 @@ const Tabs = ({ nodeData: { children, options = {} }, ...rest }) => {
     (index) => {
       const tabId = tabIds[index];
 
+      // Calculate an offset of current top of viewport from the scroll anchor ref vs. scrollY position
       const offsetY = window.scrollY - getPosition(scrollAnchorRef.current).y;
 
       setActiveTab({
@@ -72,11 +73,10 @@ const Tabs = ({ nodeData: { children, options = {} }, ...rest }) => {
         tabSet: tabsetName,
       });
 
+      // Delay preserving scroll behavior by 40ms to allow other tabset content bodies to render
       window.setTimeout(() => {
         window.scrollTo(0, getPosition(scrollAnchorRef.current).y + offsetY);
       }, 40);
-
-      // setPosition(getPosition(scrollAnchorRef.current).y);
     },
     [setActiveTab, tabIds, tabsetName] // eslint-disable-line react-hooks/exhaustive-deps
   );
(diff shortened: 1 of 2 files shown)


src/components/Tabs.js
```diff
@@ -59,31 +72,38 @@ const Tabs = ({ nodeData: { children, options = {} }, ...rest }) => {
         tabId,
         tabSet: tabsetName,
       });
+
+      // Delay preserving scroll behavior by 40ms to allow other tabset content bodies to render
+      window.setTimeout(() => {
+        window.scrollTo(0, getPosition(scrollAnchorRef.current).y + offsetY);
+      }, 40);
```
casthewiz, line 79: I'm not completely satisfied by this, but seems to be less overhead than dealing with stateful soup in trying to manage observing/listening for other tab sets to finish rendering child content.

src/components/Tabs.js
```diff
@@ -59,31 +72,38 @@ const Tabs = ({ nodeData: { children, options = {} }, ...rest }) => {
...
-          <LeafyTab key={tabId} name={tabTitle}>
-            {tab.children.map((child, i) => (
-              <ComponentFactory {...rest} key={`${tabId}-${i}`} nodeData={child} />
-            ))}
-          </LeafyTab>
-        );
-      })}
-    </StyledTabs>
+    <>
+      <div ref={scrollAnchorRef} aria-hidden="true"></div>
```
casthewiz, line 86: Not attached to having an invisible div as a ref for scroll anchoring purposes, but I don't believe the LeafyGreen tabset supports [ref forwarding](https://reactjs.org/docs/forwarding-refs.html) from what I could see in the [implementation](https://github.com/mongodb/leafygreen-ui/blob/main/packages/tabs/src/Tabs.tsx)

It might also be preferable to own the scroll anchor ref in our own layer regardless, so we don't have to worry about compatibility or a regression if we update our leafygreen version 🤷‍♂️

casthewiz:
Very open to any thoughts about how this could be done differently or cleaner.

There is some observable occasional screen jitter that I could smooth scroll over, but it would also make the scrolling experience more pronounced.

src/components/Tabs.js
sophstad: Good point, we could reach out to see if they'd be interested in supporting ref forwarding, but I agree that it may be preferable to own that ourselves.

sophstad (approved):
LGTM! I ultimately think this approach does a nice job of keeping things simple. Appreciate your comments too!

PR #381, opened 2021-04-12
DOP-1846: Add ProductsList component

rayangler:
### Stories/Links:

DOP-1846

### Staging Links:

[Drivers Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/drivers/raymundrodriguez/DOP-1846/)
[Landing Staging](https://docs-mongodbcom-staging.corp.mongodb.com/docs-nav/landing/raymundrodriguez/DOP-1846/)

### Notes:
* Removed usage of `Sidebar` and replaced with empty LG `SideNav` for demo of new component.
* Added `ProductsList` component ("View all Products").
* Added page option to show `ProductsList` on specific pages only: `index` and `search` of `docs-landing`. ([Example commit](https://github.com/rayangler/docs-landing/commit/6ad23868a3deb6238907842742a4abb94f64ab71))


src/components/ProductsList.js
```diff
@@ -0,0 +1,109 @@
...
+
+  useEffect(() => {
+    const fetchAllProducts = async () => {
+      const client = getStitchClient(SNOOTY_STITCH_ID);
+      await client.auth.loginWithCredential(new AnonymousCredential()).catch(console.error);
+      const products = await client.callFunction('fetchAllProducts', [database]);
+      setProducts(products);
+    };
+    fetchAllProducts();
+  }, [database, setProducts]);
```
sophstad, line 85: I think this data ultimately should not be fetched client-side. The spec says

> Upon build of the docs-landing repo, Realm will be used to query for all of the documents in the collection.

and I think since the component is not used across repos, situating it in `gatsby-node.js` would optimize the user experience (i.e., this data is included in the static HTML). A great approach here may be to [create GraphQL nodes](https://www.gatsbyjs.com/docs/reference/config-files/actions/#createNode) to query with a [static query](https://www.gatsbyjs.com/docs/how-to/querying-data/use-static-query/). [This](https://www.gatsbyjs.com/docs/conceptual/data-fetching/) is another doc I find helpful—let me know if you'd like to hop on a call to discuss any of this!

src/components/ProductsList.js
```diff
@@ -0,0 +1,109 @@
...
+  display: inline-block;
+  font-size: ${theme.fontSize.small};
+  letter-spacing: 0;
+  padding-left: 44px;
+  padding-right: ${theme.size.medium};
+  width: 100%;
+
+  :hover {
+    color: ${uiColors.gray.dark2};
+    font-weight: bold;
```
sophstad, line 53: I think applying `font-weight: bold` here is causing some layout shifts across the other list items. This is kinda surprising because it is usually a pattern seen with horizontal layouts ([source](https://css-tricks.com/bold-on-hover-without-the-layout-shift/)). Perhaps it has to do with the `line-height` of each `<li>`?

https://user-images.githubusercontent.com/9298431/114454925-222bb480-9ba9-11eb-9bcb-4af06e3a8dcd.mov

src/layouts/index.js
```diff
@@ -82,18 +85,17 @@ const DefaultLayout = (props) => {
             <GlobalGrid>
               <SidebarContextProvider isSidebarEnabled={isSidebarEnabled}>
                 <Header />
-                {isSidebarEnabled && <Sidenav pageContext={pageContext} />}
```
sophstad, line 85: This seems like the right call to me 👍

src/components/Sidenav.js
```diff
@@ -1,65 +1,31 @@
...
+const StyledLeafygreenSideNav = styled(LeafygreenSideNav)`
+  grid-area: sidebar;
+  z-index: 1;
+`;
 
-  useEffect(() => {
-    setIsSidebarMenuOpen(!isTabletOrMobile);
-  }, [isTabletOrMobile, setIsSidebarMenuOpen]);
+const Sidenav = ({ pageContext }) => {
+  const showAllProducts = pageContext?.page?.options?.['nav-show-all-products'];
```
sophstad, line 15: This is a great approach! 💯

sophstad (changes requested):
This is looking awesome, Raymund!

src/components/ProductsList.js
rayangler: Thanks for noticing the shifting! I switched around where the padding was being added so that hovering over a product shouldn't cause other list items to shift

src/components/ProductsList.js
rayangler: Thanks for the docs! Super helpful

gatsby-node.js
```diff
@@ -83,6 +85,20 @@ exports.sourceNodes = async () => {
...
+  products.forEach((product) => {
+    createNode({
+      id: createNodeId(`Product-${product.title}`),
+      internal: {
+        contentDigest: createContentDigest(product),
+        type: 'Product',
+      },
+      title: product.title,
+      url: product.baseUrl + product.slug,
+    });
```
sophstad, line 100: AFAICT `children` and `parent` are required properties:
```suggestion
      parent: null,
      children: [],
    });
```

src/components/ProductsList.js
```diff
@@ -0,0 +1,85 @@
...
+
+const ProductsListContainer = styled('div')`
+  line-height: 20px;
+  margin-top: ${theme.size.large};
+  width: 100%;
+`;
+
+const ProductsListHeading = styled('div')`
+  align-items: center;
+  color: ${(props) => (props.isOpen ? uiColors.gray.dark3 : uiColors.gray.dark1)};
```
sophstad, line 27: [nit] Could tidy up 🧹 
```suggestion
  color: ${({ isOpen }) => (isOpen ? uiColors.gray.dark3 : uiColors.gray.dark1)};
```

src/components/Sidenav.js
```diff
@@ -1,65 +1,29 @@
...
-            toggleLeftColumn={toggleLeftColumn}
-          />
-        </div>
-      ) : (
-        <span className={`showNav ${style.showNav} ${renderStatus}`} id="showNav" onClick={toggleLeftColumn}>
-          Navigation
-        </span>
-      )}
-    </div>
+    <StyledLeafygreenSideNav aria-label="Side navigation">
```
sophstad, line 16: So much simpler! *chef's kiss*

src/templates/document.js
```diff
@@ -18,46 +37,30 @@ const Document = ({
...
-        css={css`
-          grid-area: right;
-        `}
-      >
-        <TabSelectors />
-        <Contents />
-      </RightColumn>
-    </div>
+    <>
+      <Sidenav pageContext={pageContext} />
```
sophstad, line 47: We can probably get away with just passing the `page` property of `pageContext` here, since only the `page.options` object is used by `Sidenav`.

src/templates/landing.js
```diff
@@ -127,6 +129,13 @@ const Landing = ({ children, className }) => {
 Landing.propTypes = {
   children: PropTypes.arrayOf(PropTypes.node),
   className: PropTypes.string,
+  pageContext: PropTypes.shape({
+    metadata: PropTypes.shape({
+      publishedBranches: PropTypes.object,
+      toctree: PropTypes.object,
+    }).isRequired,
+    slug: PropTypes.string,
+  }).isRequired,
```
sophstad, line 138: [nit] Maybe extraneous since none of these props are used by this component

tests/unit/Banner.test.js
```diff
@@ -11,6 +11,15 @@ const mockBannerContent = {
...
 
+const mockClient = {
+  auth: {
+    loginWithCredential: jest.fn(() => {
+      return new Promise((resolve) => resolve());
+    }),
+  },
+  callFunction: jest.fn(() => Promise.resolve(mockBannerContent)),
+};
+
```
sophstad, line 22: This is probably unnecessary after moving the Realm call to the server!

sophstad (changes requested):

src/components/SidebarMobileMenuButton.js
```diff
@@ -32,7 +32,7 @@ const SidebarMobileMenuButton = ({ className }) => {
 
   return (
     <MenuButtonContainer className={className}>
-      <MenuButton aria-label="View All Products" onClick={clickMenu}>
+      <MenuButton aria-label="View sidenav" onClick={clickMenu}>
```
sophstad, line 35: It looks like the menu button has kinda lost control of the side nav. Is this something worth addressing later/with LeafyGreen, or should it be working now?

https://user-images.githubusercontent.com/9298431/114600675-98d8b880-9c62-11eb-8c75-f84caad9f067.mov

src/components/ProductsList.js
```diff
@@ -0,0 +1,85 @@
...
+
+const ProductLink = styled(Link)`
+  color: ${uiColors.gray.dark2};
+  display: inline-block;
+  font-size: ${theme.fontSize.small};
+  letter-spacing: 0;
+  padding: ${theme.size.tiny} ${theme.size.medium} ${theme.size.tiny} 44px;
+  width: 100%;
+
+  :hover {
```
sophstad, line 46: I think this hover state looks slightly awkward when it causes an entry to break onto a new line on hover. This may also be addressed by the wider `SideNav` component, so let's keep an eye on it.

src/components/ProductsList.js
```diff
@@ -0,0 +1,85 @@
...
+import { theme } from '../theme/docsTheme';
+
+const HeadingTitle = styled('span')`
+  padding-left: ${theme.size.small};
+`;
+
+const Products = styled(`ul`)`
+  list-style-type: none;
+  padding: 0;
+  ${({ isOpen }) => !isOpen && 'display: none;'}
```
sophstad, line 16: Nice 💯

src/components/SidebarMobileMenuButton.js
rayangler: I think this could be one of the potential asks for LeafyGreen. It would be nice to have a way to pass a collapsed/expanded state to the LG `SideNav`. (This would also help with the design for screen sizes ≤ 767px that have the `SideNav` closed by default)

It would be possible to have that menu button control the `display` of the `SideNav` as it currently is, but the following scenario might make it buggy:
1) Close/collapse the `SideNav` on a screen size > 420px using the `SideNav`'s collapse button.
2) Change the screen size to ≤ 420px.
3) Use the menu button to control the `display` of the `SideNav`.
4) The items in the `SideNav` will not be visible unless hovering over them because the `SideNav` would still be in its closed state.

An alternative solution (that just came to mind) to the above scenario may be to have our `Sidenav` component use a completely different parent element instead of the LG `SideNav` when on mobile screen size! (But having a way to control the state for screen sizes > 420px and ≤ 767px would still be good)

rayangler: I'm happy to continue looking into other options in the meantime, but I'm also good with addressing all sidenav-related stuff in DOP-1839

src/components/ProductsList.js
rayangler: Yeah, I'm not too sure what to do about it at the moment or how it could be handled to make it less awkward. I'm hoping a wider component will help

src/components/SidebarMobileMenuButton.js
sophstad: Yup I think this makes perfect sense, so let's raise with them. Thanks!

sophstad (approved):

PR #384, opened 2021-04-16
DOP-2105: Replace version dropdown with LG Select component

sophstad:
### Stories/Links:

DOP-2105

### Staging Links:

- [BI Connector v2.14](https://docs-mongodbcom-staging.corp.mongodb.com/master/bi-connector/sophstad/DOP-2105/)
- [BI Connector v2.13](https://docs-mongodbcom-staging.corp.mongodb.com/v2.13/bi-connector/sophstad/DOP-2105/)

Note that other links in the dropdown will 404 because I didn't stage every version.

### Notes:

- Using Enzyme's `shallow` for testing now that we aren't managing the Select component's internals.
- A bit of hacking was necessary to remove LeafyGreen's default `null` option ([demo](https://www.mongodb.design/component/select/example/)), which doesn't really makes sense here. I was able to target and hide it by setting `usePortal` to false, which required a bit of z-index adjustment to correctly stack the dropdown.


src/components/VersionDropdown.js
```diff
@@ -97,34 +74,28 @@ const VersionDropdown = ({
...
+      placeholder={null}
+      size={Size.Large}
+      usePortal={false}
+      value={parserBranch}
+    >
+      {Object.entries(gitNamedMapping).map(([branch, name]) => (
+        <Option key={branch} value={branch}>
+          {prefixVersion(name)}
+        </Option>
+      ))}
```
casthewiz, line 96: Not blocking, open question:
Do we want to pass the `<a>` as children to the `Option` here with
```
    const legacyDocsURL = `https://docs.mongodb.com/legacy/?site=${project}`;
    const destination = value === 'legacy' ? legacyDocsURL : normalizePath(`${generatePrefix(value)}/${slug}`);
```
as the logic for the href value instead of using `navigate`? It would be more semantic, but might introduce some style issues

src/components/VersionDropdown.js
```diff
@@ -14,6 +15,19 @@ const zip = (a, b) => {
...
+  margin: 8px 12px 12px 22px;
+
+  span {
+    font-size: 16px;
+  }
+
+  /* Remove "null" selection from dropdown */
+  ul > li:first-of-type {
+    display: none;
+  }
```
casthewiz, line 28: might be worth flagging to leafygreen that [deselection](https://github.com/mongodb/leafygreen-ui/blob/e2bd0abfc5d9897e3748b86058cd726f038b79e2/packages/select/src/Select.tsx#L346-L371) should be nullable

casthewiz (approved):
Just a small nonblocking question, otherwise LGTM! 

🚢

src/components/VersionDropdown.js
sophstad: Great point, just pushed up this addition. I ended up using a dual `<a>`/`navigate` approach because it made dealing with the styling much simpler. Quite similar to the first ticket you worked on! Let me know what you think.

I'll also mention to LeafyGreen that it'd be nice to have an `as` prop (or something similar) for `Option`. Currently, we get a console error `Invalid prop children supplied to Option` because of the `<a>` tag. _And_ [this page](https://www.mongodb.design/component/select/documentation/) suggests any `node` is valid, so... 🤷‍♀️

PR #385, opened 2021-04-19
DOP-2056: Add updateNav script

sophstad:
### Stories/Links:

DOP-2056

### Notes:

- `context` is a global object available in Realm functions ([source](https://docs.mongodb.com/realm/functions/context/)), so we can ignore the failing ESLint test.
- I created a `snooty_test` DB whose `navigation` collection has been populated via this function. Let me know if you'd like to run the function yourself in Realm and I can drop the existing data in `snooty_test`!

### DIFF
--- a/scripts/updateNav.js
+++ b/scripts/updateNav.js
@@ -1,0 +1,110 @@
+const getDatabase = (env) => {
+  const dbMap = new Map([
+    ['development', 'snooty_dev'],
+    ['staging', 'snooty_stage'],
+    ['production', 'snooty_prod'],
+  ]);
+
+  if (dbMap.has(env)) {
+    return dbMap.get(env);
+  }
+
+  throw new Error('Invalid env argument supplied');
+};
+
+const getWriteOperations = (iatree, parentPaths, slugToTitle) => {
+  const getTitle = (slug) => slugToTitle[slug] || [];
+
+  const makeUrl = (slug) => {
+    const baseUrl = `https://docs.mongodb.com`;
+    return `${baseUrl}/${slug}/`;
+  };
+
+  const makeUpdateOperation = (projectName, parents) => {
+    const filter = { projectName };
+    const update = { $set: { parents } };
+    return {
+      updateOne: { filter, update },
+    };
+  };
+
+  const getProjects = (tree) => {
+    const dive = (obj) => {
+      if (obj['project_name']) {
+        projectNames.push(obj['project_name']);
+      }
+
+      if (obj.children) {
+        return obj.children.forEach(dive);
+      }
+    };
+
+    const projectNames = [];
+    dive(tree);
+    return projectNames;
+  };
+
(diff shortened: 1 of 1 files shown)


scripts/updateNav.js
```diff
@@ -0,0 +1,110 @@
...
+    dive(tree);
+    return projectNames;
+  };
+
+  const getWrites = (projectNames) =>
+    projectNames.map((project) => {
+      const projectParents = parentPaths[project] || [];
+      const parentEntries = projectParents.map((parent) => ({
+        title: getTitle(parent),
+        url: makeUrl(parent),
```
rayangler, line 52: Should we validate that a parent is a project's `primary` parent before writing it into the `navigation` collection?

In `snooty_test.navigation` and `snooty_test.metadata`, I see that `cloud-docs`'s parent is `view-analyze`. However, [this commit](https://github.com/sophstad/docs-landing/commit/1edbd7563372a574cb88cbeb532c1fc896bff1e1#diff-82e309fc89985f46c95e770e2ec3f98535f4d75c6654e01c3fbedd7f2f2918ddR9-R12) shows that the `primary` parent of `cloud-docs` is `launch-manage`.

scripts/updateNav.js
```diff
@@ -0,0 +1,110 @@
...
+const getWriteOperations = (iatree, parentPaths, slugToTitle) => {
+  const getTitle = (slug) => slugToTitle[slug] || [];
+
+  const makeUrl = (slug) => {
+    const baseUrl = `https://docs.mongodb.com`;
+    return `${baseUrl}/${slug}/`;
+  };
+
+  const makeUpdateOperation = (projectName, parents) => {
+    const filter = { projectName };
```
rayangler, line 24: How should we handle projects that don't have formal project/snooty names but might still be worth adding an `ia` for? For example: Ops Manager, Cloud Manager, and Atlas Search

sophstad: Interesting question -- I'm somewhat inclined to say it's a non-issue for Ops Manager & Cloud Manager. If they don't have a snooty name (i.e. are legacy), they're never going to be updated to leverage links to the parent.

Atlas Search is a different issue, but I'm inclined to say there's similarly not much use for a parent because the documentation is fully integrated within Atlas. This is something I'd be interested to keep an eye on for future uses, though.

scripts/updateNav.js
sophstad: Whoops, I think the collection was last updated _before_ I added [this commit](https://github.com/mongodb/snooty-parser/pull/299/commits/f8e7fc834574685ed6e72e135884d447a6993098) to the parser. I just dropped the `navigation` collection and ran the function again and it looks good now. Sorry for the confusion!

rayangler (approved):
LGTM! Thanks for answering my questions

sophstad:
Deployed to Realm, thanks!

PR #387, opened 2021-04-20
DOP-2083: Only keydown event listen on searchbar focus

casthewiz:
### Stories/Links:

DOP-2083

### Staging Links:

https://docs-mongodbcom-staging.corp.mongodb.com/master/realm/cassidy.schaufele/DOP-2083/

### Notes:

The event listener for the keydown event was inexplicably triggering on page load - this is why arrow down navigation wasn't working until clicking into another element, or until clicking into and then out of the searchbar.

This changeset changes the event listener to only activate when the searchbar has focus, and passes `isFocused` into `ExpandedSearchbar`. We had the prop present in `ExpandedSearchbar` before, we just never drilled it down from `Searchbar`.


sophstad (approved):
Perfecto

PR #389, opened 2021-04-22
DOP-1841: Update page breadcrumbs to fetch parents

rayangler:
### Stories/Links:

DOP-1841

### Staging Links:

[Drivers Staging](https://docs-mongodborg-staging.corp.mongodb.com/master/drivers/raymundrodriguez/DOP-1841/)
[Atlas Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/cloud-docs/raymundrodriguez/DOP-1841/)
[Landing Staging](https://docs-mongodbcom-staging.corp.mongodb.com/docs-nav/landing/raymundrodriguez/DOP-1841/view-analyze/)

### Notes:
* Separated `BreadcrumbContainer` into its own component so `BreadcrumbSchema` is added to page source.
* Created Realm function `fetchProjectParents` to return the parents of a given project.
* Re-added some code to `Document.js` that were removed during one of the merges from `master` to `docs-nav`.

### DIFF
--- a/src/components/BreadcrumbContainer.js
+++ b/src/components/BreadcrumbContainer.js
@@ -58,7 +58,11 @@ const BreadcrumbContainer = ({ lastCrumb }) => {
   useEffect(() => {
     const fetchBreadcrumbData = async () => {
       let parentCrumbs = await fetchProjectParents(SNOOTY_STITCH_ID, database, project);
-      const breadcrumbs = [{ title: 'Docs Home', url: 'https://docs.mongodb.com/' }, ...parentCrumbs, lastCrumb];
+      const breadcrumbs = [
+        { title: 'Docs Home', url: project === 'landing' ? '/' : 'https://docs.mongodb.com/' },
+        ...parentCrumbs,
+        lastCrumb,
+      ];
       setBreadcrumbs(breadcrumbs);
     };
     fetchBreadcrumbData();
--- a/src/components/BreadcrumbContainer.js
+++ b/src/components/BreadcrumbContainer.js
@@ -6,29 +6,12 @@ import { css } from '@emotion/core';
 import Link from './Link';
 import { SNOOTY_STITCH_ID } from '../build-constants';
 import { useSiteMetadata } from '../hooks/use-site-metadata';
-import { theme } from '../theme/docsTheme';
 import { formatText } from '../utils/format-text';
 import { reportAnalytics } from '../utils/report-analytics';
 import { fetchProjectParents } from '../utils/stitch';
 
 const activeColor = css`
   color: ${uiColors.gray.dark3};
-`;
-
-const StyledNav = styled('nav')`
-  font-size: ${theme.fontSize.small};
-
-  * {
-    color: ${uiColors.gray.dark1};
-  }
-
-  & > p {
-    margin-top: 0;
-  }
-`;
-
-const P = styled('p')`
-  min-height: ${theme.size.medium};
 `;
 
 const StyledArrow = styled('span')`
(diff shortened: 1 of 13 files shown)


src/components/BreadcrumbContainer.js
```diff
@@ -0,0 +1,100 @@
...
+  ${({ isLast }) => isLast && activeColor}
+  cursor: default;
+`;
+
+const StyledLink = styled(Link, {
+  shouldForwardProp: (prop) => prop !== 'isLast',
+})`
+  ${({ isLast }) => isLast && activeColor}
+
+  :hover, :focus {
```
sophstad, line 41: [nit] Could this be an opportunity to use `:last-of-type` rather than a JS prop?

src/components/MainColumn.js
```diff
@@ -17,7 +17,7 @@ const MainColumn = ({ children, className }) => (
 
 MainColumn.propTypes = {
   className: PropTypes.string,
-  children: PropTypes.array,
+  children: PropTypes.node,
```
sophstad, line 20: We probably want to handle both of these cases like we do [here](https://github.com/mongodb/snooty/blob/master/src/layouts/index.js#L71)
```suggestion
  children: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.node), PropTypes.node]).isRequired,
```

src/components/EcosystemHomepageTiles.js
```diff
@@ -84,8 +84,8 @@ const EcosystemHomepageTiles = () => {
 
   return (
     <ul className={tileStyles.tileOuter}>
-      {tileValues.map((element, index) => (
-        <Link to={element.slug} className={tileStyles.tileAnchor}>
+      {tileValues.map((element) => (
+        <Link to={element.slug} className={tileStyles.tileAnchor} key={element.slug}>
```
sophstad, line 88: Nice improvement 😊

src/components/BreadcrumbContainer.js
```diff
@@ -0,0 +1,105 @@
...
+  :last-of-type {
+    ${activeColor}
+  }
+
+  :hover,
+  :focus {
+    text-decoration: none;
+
+    :not(:last-of-type) {
+      ${activeColor}
```
sophstad, line 49: It doesn't look like this rule is being properly applied on Atlas 🤔 it's working on Drivers though! Maybe just a staging mixup?

https://user-images.githubusercontent.com/9298431/115900287-c3c6c780-a42d-11eb-8306-6c28f147bbdd.mov

rayangler: Is this issue still happening for you? I just tried it and it seems to be fine at the moment

sophstad: Ah yup, looks resolved. Weird, thanks for looking into it!

sophstad (approved):
Great work! I think these breadcrumbs look super slick. 💯

rayangler:
> Just noticed that breadcrumbs appear on new docs-landing content pages (e.g. "Develop Applications"). Can we make sure these are working as expected?

Utilized the `pageTitle` prop in `Breadcrumbs` and introduced a `:breadcrumbs-page-title:` option to each of the 3 project parent pages on landing to work this out. I updated the staging links and added one for Landing. Let me know if this seems like a reasonable approach! @sophstad

src/templates/document.js
```diff
@@ -35,19 +35,26 @@ const Document = ({
   pageContext: {
     slug,
     page,
-    metadata: { parentPaths, slugToTitle: slugTitleMapping, toctreeOrder },
+    metadata: { parentPaths, slugToTitle: slugTitleMapping, title, toctreeOrder },
   },
 }) => {
   const pageOptions = page?.options;
   const showPrevNext = !(pageOptions && pageOptions.noprevnext === '');
+  const breadcrumbsPageTitle = pageOptions?.['breadcrumbs-page-title'] === '';
```
sophstad, line 43: Hmm interesting approach. It seems to me like we are splitting the logic for which "type" of breadcrumb to use based on two conditions: here we check for this page option, but later we depend on the project name being `landing`; theoretically it would be nice to consolidate that. Is there any reason we couldn't use a project check here?

rayangler: Changed! Thanks for pointing that out!

src/components/BreadcrumbContainer.js
```diff
@@ -0,0 +1,109 @@
...
+        lastCrumb,
+      ];
+      setBreadcrumbs(breadcrumbs);
+    };
+    fetchBreadcrumbData();
+  }, [database, lastCrumb, project]);
+
+  if (breadcrumbs.length === 0) {
+    return null;
+  }
```
sophstad, line 69: One final nitpicky request: it seems possible that this has a negative effect on our CLS—I even noticed some jumpiness when looking at staging on a slow connection. Rather than returning `null`, it would be great to place some sort of empty space here that has the same height as breadcrumb content. This way when the data is returned, rendering it doesn't affect the positioning of the rest of the page.

sophstad (changes requested):

src/components/BreadcrumbContainer.js
rayangler: I added a `min-height` to the `p` [here](https://github.com/mongodb/snooty/pull/389/files#diff-c0df34c9dda01b2ba93ec946f2e1f4db562cbd56a5e60499d101e8967e060356R30-R32) to have some empty space while the breadcrumbs load. I also had to [add](https://github.com/mongodb/snooty/pull/389/files#diff-aae663948a10f0ae83084fc520ef95f1067c43faef5c92b8abb5d504d627903dR11) a `Wrapper` to the `Breadcrumbs` component to actually allot some space to the loadable component while it loads.

sophstad (approved):
Looks immaculate. Thank you Raymund!

PR #392, opened 2021-04-27
DOP-2115: Add Additional Links to SideNav

rayangler:
### Stories/Links:

DOP-2115

### Staging Links:

[Landing Staging](https://docs-mongodborg-staging.corp.mongodb.com/docs-nav/landing/raymundrodriguez/DOP-2115/
)

### Notes:
* Added additional links to the bottom of the new SideNav.
* I set one of the elements of LG SideNav to be a flex item so that we can have a child flex grow. Open to any better alternatives so that we don't have to modify the css of the SideNav!

### DIFF
--- a/src/components/Sidenav.js
+++ b/src/components/Sidenav.js
@@ -3,10 +3,8 @@ import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
 import Icon from '@leafygreen-ui/icon';
 import { uiColors } from '@leafygreen-ui/palette';
-import { SideNav as LeafygreenSideNav } from '@leafygreen-ui/side-nav';
-import Link from './Link';
+import { SideNav as LeafygreenSideNav, SideNavItem } from '@leafygreen-ui/side-nav';
 import ProductsList from './ProductsList';
-import { theme } from '../theme/docsTheme';
 
 const StyledLeafygreenSideNav = styled(LeafygreenSideNav)`
   grid-area: sidebar;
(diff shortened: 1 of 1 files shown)


src/components/Sidenav.js
```diff
@@ -1,20 +1,76 @@
...
     <StyledLeafygreenSideNav aria-label="Side navigation">
       {showAllProducts && <ProductsList />}
+      <Spaceholder />
+      <AdditionalLinksContainer>
+        {additionalLinks.map(({ glyph, title, url }) => {
+          return (
+            <AdditionalLink to={url} key={title}>
+              <StyledIcon glyph={glyph} />
+              <LinkTitle>{title}</LinkTitle>
+            </AdditionalLink>
```
sophstad, line 70: Hmm, I think this is probably a good use case for LG's `SideNavItem` component (which even has a somewhat undocumented `glyph` prop). I suppose this will be a slight departure from Allison's designs, so maybe check in with her about it, but I believe would be preferable to really leverage this package's offerings! cc @allisonmui9

src/components/Sidenav.js
```diff
@@ -1,20 +1,53 @@
...
+// Allows AdditionalLinks to always be at the bottom of the SideNav
+const Spaceholder = styled('div')`
+  flex-grow: 1;
 `;
 
+const StyledSideNavItem = styled(SideNavItem)`
+  letter-spacing: 0;
+
+  :hover {
+    color: ${uiColors.gray.dark2};
```
sophstad, line 29: Nice catch! Perhaps worth using `color: unset` instead so that this component will be responsive to hypothetical color changes on LG's end. I've also been leaving comments [like this](https://github.com/mongodb/snooty/blob/master/src/components/CaptionLegend.js#L14) around styling that is obviously addressing legacy CSS tech debt.

src/components/Sidenav.js
```diff
@@ -1,20 +1,53 @@
...
+
+  // Allows Spaceholder element to flex grow for AdditionalLinks
+  & > div > nav > div > ul {
+    display: flex;
+    flex-direction: column;
+  }
+`;
+
+// Allows AdditionalLinks to always be at the bottom of the SideNav
+const Spaceholder = styled('div')`
```
sophstad, line 21: I like this!

sophstad (approved):
Looks great!

PR #396, opened 2021-05-03
Proposal: Fork mongodb-docs.css

sophstad:
### Staging Links:

- [Landing](https://docs-mongodbcom-staging.corp.mongodb.com/master/landing/sophstad/fork-css/)
- [BI Connector](https://docs-mongodbcom-staging.corp.mongodb.com/master/bi-connector/sophstad/fork-css/)
- [Drivers](https://docs-mongodbcom-staging.corp.mongodb.com/master/drivers/sophstad/fork-css/)

### Notes:

While working on the new navigation, I kept coming across rules from `mongodb-docs.css` that were messing with styling due to their (lack of) scoping. Since we are not actively developing any new features for the legacy toolchain, a centralized stylesheet is limited in usefulness. On the off chance we do need to make a change to CSS in legacy and next-gen, we'll have to do so twice. However, this is not much different than the current workflow that requires updating the submodule commit.

Moving away from the submodule has a few advantages:

- Leverage Gatsby's [recommended way](https://www.gatsbyjs.com/docs/how-to/styling/built-in-css/) of importing a CSS file via JS
- Allow us to make CSS tweaks without potentially affecting legacy docs
- More easily complete CSS refactor by having our CSS version controlled and easily editable in one place
- Simplify workflow: once docs nav work is completed, we can get rid of the submodule altogether
- Increase visibility into our static dependencies (i.e. fonts and images referenced in CSS)

I haven't made any changes to the CSS file besides removing lines that caused a build failure due to invalid imports. Interested in any thoughts y'all have, thanks in advance!


src/html.js
```diff
@@ -29,11 +29,6 @@ const HTML = ({ body, bodyAttributes, headComponents, htmlAttributes, preBodyCom
         href="https://docs.mongodb.com/osd.xml"
         title="MongoDB Help"
       />
-      {process.env.GATSBY_SITE === 'guides' ? (
-        <link rel="stylesheet" href={withPrefix('docs-tools/guides.css')} type="text/css" />
-      ) : (
-        <link rel="stylesheet" href={withPrefix('docs-tools/mongodb-docs.css')} type="text/css" />
-      )}
```
casthewiz, line 36: Do we need to replace the logic for `docs-tools/guides.css` here, or otherwise replace it as well?

sophstad: Guides is built on legacy tooling right now, and Q2 has a guides redesign project scheduled which will likely result in an Emotion-based implementation. As a result I think there's no need to load the guides CSS file, though it will still be accessible [here](https://github.com/mongodb/docs-tools/tree/master/themes/mongodb/src/css/guides) if need be!

casthewiz (approved):
👏 👏 👏

rayangler:
For clarification, will the `mongo-docs.css` file be expected to remain even after the css refactor?

sophstad:
@rayangler Ultimately no, I think! There are much better ways of writing CSS at this point (although I imagine a smaller global stylesheet for things like fonts).

rayangler (approved):
LGTM!

PR #397, opened 2021-05-04
DOP-2095: Add Sidebar component into PLP

mango-db:
### Stories/Links:

[DOP-2095](https://jira.mongodb.org/browse/DOP-2095)

### Staging Links:

[Docs-compass](https://docs-mongodbcom-staging.corp.mongodb.com/DOP-1979/compass/cgoecknerwald/DOP-2095/)

### Notes:

There were some interesting design questions arising from the 2-column / 1-column layout boundary. Should the layout switch to 2 columns at **medium** (`max-width: 767px`), **large** (`max-width: 1023px`), or somewhere in-between at **medium-large** (`max-width: 895px`)?

(@allisonmui9: the current layout boundary options are [here](https://github.com/mongodb/snooty/blob/master/src/theme/docsTheme.js#L47-L58). "Medium-large" would need to be added.)

Currently, I have opted to have the layout change to 2 columns at  **large** (`max-width: 1023px`). The downside is that the hero image becomes very massive on slightly-under-large screens (e.g. `1020px`) with a collapsed Sidebar. I opted to constrain the `max-width` of the hero image to `600px`. Here's what that looks like:

image

Note the copious amounts of whitespace. For comparison, if we constrain the hero image to `800px`:

image

Versus if we don't constrain the `max-width` of the hero image at all:

image

#### Changing the layout boundary:

If the boundary was reverted to **medium**, and if we expand the sidebar, we get the following result on slightly-under-medium screens (e.g. `770px`):

image

Since the above is so gross, I opted to move the layout boundary to **large** (`max-width: 1023px`). 

Two questions result:

1. Would it be desirable to create a **medium-large** layout boundary at `~895px` (splitting the difference between **medium** and **large**)?
2. If I leave the layout boundary at **large**, what `max-width` should I constrain the hero image to?

### DIFF
--- a/src/components/Tabs.js
+++ b/src/components/Tabs.js
@@ -59,7 +59,7 @@ const landingTabStyling = css`
     display: block;
   }
 
-  @media ${theme.screenSize.upToMedium} {
+  @media ${theme.screenSize.upToLarge} {
     display: block;
   }
 `;
--- a/src/components/landing/Procedure.js
+++ b/src/components/landing/Procedure.js
@@ -7,7 +7,7 @@ import { theme } from '../../theme/docsTheme';
 const StyledProcedure = styled('div')`
   max-width: 500px;
   padding-left: ${theme.size.large};
-  @media ${theme.screenSize.upToMedium} {
+  @media ${theme.screenSize.upToLarge} {
     padding-bottom: ${theme.size.large};
   }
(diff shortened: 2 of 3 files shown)


allisonmui9 (approved):

rayangler (approved):
LGTM

PR #403, opened 2021-05-10
DOP-2160: Add IA sidebar transition

sophstad:
### Stories/Links:

DOP-2160

### Staging Links:

- [Landing](https://docs-mongodbcom-staging.corp.mongodb.com/nav/landing/sophstad/DOP-2160/)
- [Drivers](https://docs-mongodbcom-staging.corp.mongodb.com/master/drivers/sophstad/DOP-2160/) (sidebar still rendering correctly in ecosystem-index template)

### Notes:

- @allisonmui9 adding you for UAT for transition animations!
- Add sidebar to shared layout component
- Add content animation (fade out-in) to all page transitions

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,6 +55,7 @@
         "react-dom": "^16.14.0",
         "react-helmet": "^6.1.0",
         "react-loadable": "^5.5.0",
+        "react-transition-group": "^4.4.1",
         "realm-web": "^1.2.1",
         "redoc": "^2.0.0-rc.48",
         "sanitize-html": "^2.2.0",
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
     "react-dom": "^16.14.0",
     "react-helmet": "^6.1.0",
     "react-loadable": "^5.5.0",
+    "react-transition-group": "^4.4.1",
     "realm-web": "^1.2.1",
     "redoc": "^2.0.0-rc.48",
     "sanitize-html": "^2.2.0",
--- a/src/components/ContentTransition.js
+++ b/src/components/ContentTransition.js
@@ -1,0 +1,51 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { SwitchTransition, CSSTransition } from 'react-transition-group';
+import { css, Global } from '@emotion/core';
+
+const fadeOut = css`
+  .fade-exit {
+    opacity: 1;
+  }
+  .fade-exit-active {
+    opacity: 0;
+    transition: opacity 100ms;
+  }
+`;
+
+const fadeIn = css`
+  .fade-enter {
+    opacity: 0;
+  }
+  .fade-enter-active {
+    opacity: 1;
+    transition: opacity 300ms;
+  }
+`;
+
+const fadeInOut = css`
+  ${fadeOut}
+  ${fadeIn}
+`;
+
+const ContentTransition = ({ children, slug }) => (
+  <>
+    <Global styles={fadeInOut} />
+    <SwitchTransition>
+      <CSSTransition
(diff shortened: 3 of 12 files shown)


rayangler (commented):
This looks amazing! I think everything looks good to me except I'm running into an issue where the "On This Page" box doesn't change the active heading. (For example, go to [this page](https://docs-mongodbcom-staging.corp.mongodb.com/master/drivers/sophstad/DOP-2160/reactive-streams/) and use it to navigate to the next page "PHP Driver")

Is this happening for you as well?

src/components/DocumentBody.js
```diff
@@ -88,29 +91,54 @@ export default class DocumentBody extends Component {
...
+            site {
+              siteMetadata {
+                project
+              }
+            }
+          }
+        `}
+        render={({
+          site: {
+            siteMetadata: { project },
```
rayangler, line 113: Is this different from `useSitemetadata`?

src/components/DocumentBody.js
```diff
@@ -88,29 +91,54 @@ export default class DocumentBody extends Component {
...
+              <SEO pageTitle={pageTitle} siteTitle={siteTitle} />
+              <Widgets
+                location={location}
+                pageOptions={page?.options}
+                pageTitle={pageTitle}
+                publishedBranches={getNestedValue(['publishedBranches'], metadata)}
+                slug={slug}
+              >
+                <FootnoteContext.Provider value={{ footnotes: this.footnotes }}>
+                  <ContentsProvider headingNodes={page?.options?.headings}>
```
rayangler, line 128: For my understanding, how does putting the `ContentsProvider` here fix the issue with the `Contents` component as opposed to when it used to be in `layouts/index`?

src/components/RootProvider.js
```diff
@@ -1,20 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { ContentsProvider } from './contents-context';
 import { HeaderContextProvider } from './header-context';
 import { NavigationProvider } from './navigation-context';
 import { SidebarContextProvider } from './sidebar-context';
 import { TabProvider } from './tab-context';
 
 const RootProvider = ({ children, headingNodes, isSidebarEnabled, pageTitle, selectors }) => (
```
rayangler, line 8: Is `headingNodes` still needed?

src/components/DocumentBody.js
sophstad: Certainly! So, `layouts/index` is wrapped around the entire _site_ via Gatsby's `wrapPageElement` API (which is in turn implemented by `gatsby-plugin-layout`). The point of this plugin is to prevent components from rerendering unnecessarily on page changes ([source](https://www.gatsbyjs.com/docs/how-to/routing/layout-components/#how-to-prevent-layout-components-from-unmounting)); for example, we never really need to remount the top navbar. The `contents` directive, however, is really not shared by pages across the site, and is unique to each page. Putting the context in this `DocumentBody`, which is instantiated by the `createPage` action [here](https://github.com/mongodb/snooty/blob/master/gatsby-node.js#L104), ensures that the sidebar is unique per page. I think the weird behavior we saw before was caused by new props constantly being passed to a component that was not unmounting as expected.

rayangler: That makes a ton of sense! Very useful to know. Thanks!

src/components/DocumentBody.js
rayangler: Ohhh, right. Do you think in the future, we could transform this into a functional component, or is this something that needs to be a class component?

sophstad: Definitely we could! The only thing that looked a bit funny was the `normalizeCss()` stuff. Didn't feel like untangling it now, and potentially is worth deprecating in the future :)

PR #408, opened 2021-05-12
DOP-2138: Add padding for additional levels of TOC

rayangler:
### Stories/Links:

DOP-2138

### Staging Links:

[BI-Connector](https://docs-mongodbcom-staging.corp.mongodb.com/DOP-2138/bi-connector/raymundrodriguez/DOP-2138/reference/known-issues/)

### Notes:
* Added padding for a couple of extra levels in the old TOC.
* Stand by for related docs-tools PR to keep the css as similar as possible, and in case something like this were to happen with legacy or old gen docs. (https://github.com/mongodb/docs-tools/pull/506)

### DIFF
--- a/src/styles/mongodb-docs.css
+++ b/src/styles/mongodb-docs.css
@@ -11993,6 +11993,12 @@ div.sphinxsidebar li.toctree-l4 > a {
 div.sphinxsidebar li.toctree-l5 > a {
   padding-left: 130px;
 }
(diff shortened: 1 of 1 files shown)


sophstad (approved):
Lol, first time we've had to make an update to legacy CSS in ages and it's right after we've split the codebases 😆 

looks great to me!

PR #411, opened 2021-05-14
DOP-2184: Upgrade LG table to use 16px font

sophstad:
### Stories/Links:

DOP-2184

### Staging Links:

[Atlas Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/cloud-docs/sophstad/DOP-2184/reference/api/database-users-update-a-user/#request-body-parameters)


tests/unit/__snapshots__/ListTable.test.js.snap
```diff
@@ -132,14 +132,15 @@ exports[`when rendering a list table with fixed widths renders correctly 1`] = `
   vertical-align: middle;
   font-size: 14px;
   line-height: 20px;
+  line-height: 20px;
```
rayangler, line 135: I'm very confused as to why this extra `line-height` is here but it seems fine on the actual page! 😅

rayangler (approved):
LGTM!

tests/unit/__snapshots__/ListTable.test.js.snap
sophstad: Interesting — I think it's actually happening on LeafyGreen's end [here](https://github.com/mongodb/leafygreen-ui/blob/main/packages/table/src/Cell.tsx#L47-L48). `getCommonCellStyles()` and `baseStyles` both set `line-height`. Nothing to worry about on our end I think, but good eye!

PR #418, opened 2021-05-25
DOP-2193: Apply css changes and theming to redoc component

rayangler:
### Stories/Links:

DOP-2193

### Staging Links:

[Atlas Staging](https://docs-mongodbcom-staging.corp.mongodb.com/openapi/cloud-docs/raymundrodriguez/DOP-2193/reference/api/openapi/)
[Original Atlas Staging](https://docs-atlas-staging.mongodb.com/cloud-docs/docsworker-xlarge/DOCSP-15099/reference/api/openapi/) (for comparison)

### Notes:
* Used known theme options to change appearance of redoc components. Additional css was added to components with exposed selectors. This may be brittle if an update to the Redoc component changes the selectors for certain components.
* Used modified versions of docs-nav's side nav components as components in Redoc's left sidebar menu.


allisonmui9:
Looks great. Few comments: 

- Can we change the right hand side background color to #21313C?
- Can we even the top & bottom margin on the badges in the TOC? 
- Probably separate and part of the nav project - but can we make the arrow next to back to atlas docs cuter? haha perhaps the arrow Claire used in the PLP or a LG icon? https://www.mongodb.design/component/icon/example/
- I think the loading state feels a bit off to me, since it jumps to only display the footer. Are there any workarounds to this? I know we had removed the big spinner... perhaps putting a smaller spinner or even including a loading bar like the docker example? 

looks great though!!! ty raymund :')

rayangler:
> Probably separate and part of the nav project - but can we make the arrow next to back to atlas docs cuter? haha perhaps the arrow Claire used in the PLP or a LG icon? https://www.mongodb.design/component/icon/example/

I'll try how the arrow was done in PLP! I initially tried it with the LG Icon but was having trouble getting it to render properly. I'll spend some more time trying to get that working if anything.

src/components/OpenAPI.js
```diff
@@ -1,16 +1,155 @@
...
+    :hover {
+      background-color: ${uiColors.gray.light2};
+    }
+    &.active {
+      background-color: ${uiColors.green.light3};
+    }
+  }
+`;
+
+const spanHttpCss = css`
```
sophstad, line 54: It looks like Verdana is the font currently being used for these badges, would be nice to apply Akizidenz here too!
image

src/components/OpenAPI.js
```diff
@@ -1,16 +1,155 @@
...
+  margin: ${theme.size.medium} ${theme.size.default};
+`;
+
+const MenuTitleContainer = ({ siteTitle, pageTitle }) => {
+  const docsTitle = siteTitle + ' Docs';
+  const text = `← Back to ${formatText(docsTitle)}`;
+
+  return (
+    <>
+      <SidebarBack border={<Border />} enableGlyph={false} textOverride={text} url={'/'} />
```
sophstad, line 138: This is a nice way to reuse this component!

allisonmui9:
Thanks Raymund! Few comments: 

- Could we adjust to these syntax highlighting colors? Sorry about the inconvenience. Just trying to stay consistent where we can with the rest of MDB.
Purple: #C678DD
Red: #E06C75
Green: #98C379
Blue: #61AEEE

- Could we add the dark container to the back of the badge and code snippets? 
Like the dark black ones here behind get & code: https://docs-atlas-staging.mongodb.com/cloud-docs/docsworker-xlarge/DOCSP-15099/reference/api/openapi/

- Could we change the non-link blues & teals to use the same: #0D324F? 

- I also noticed the loading screen flashes from the big loader then to the small loader then to the API docs.

allisonmui9 (changes requested):
Left comments above ^. Thank you Raymund!

casthewiz (approved):
One question around how some of the classnames get generated, but otherwise LGTM!

I like the use of `TODO:` to keep track of what we might need to change in the near future 😄

src/components/OpenAPI.js
```diff
@@ -1,16 +1,209 @@
...
+
+  // Code block divs
+  div.hoverable {
+    color: ${uiColors.white};
+  }
+`;
+
+const inlineCodeCss = css`
+  // InlineCode inside of Parameters and Schemas
+  span.sc-eLgOdN {
```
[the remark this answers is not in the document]
rayangler: `sc-eLg0dN` is a classname generated by the styling that Redoc uses for that specific component. It should remain the same as long as the Redoc component's styling doesn't change. (You can read more about how the classname is generated and hashed in this [article](https://medium.com/styled-components/how-styled-components-works-618a69970421), if you'd like.) 

I'm not a big fan of using this as selectors because if we were to update our version of Redoc and the component's styling changes, the classname would change as well. I tried to leave comments for each of the times we do use these classnames as selectors so we could have a sort of reference to double-check on the page if it changes.

The classname also seems to remain even if using a totally different openapi spec file (sanity check)
sc-sanity-check

mango-db (commented):
Closed Q's:
- [x] Updates the div radii & the formerly invisible colons in the code (thank you!)

Open Q's:

- Are the response types (red and green on dark grey, with black-on-dark-grey borders) sufficiently high-contrast?

image

 - Has the expanded text box been checked for stylistic adherence?
 
image

- Can/should we run an accessibility audit on this page? 

- The following redoc "Patch" has a different weight (400) from other similar UI items on the page:

image
Other similar elements with 600/700 weight:
image
image

rayangler:
@allisonmui9   

* Syntax colors issue seems to have been resolved as discussed offline 
* Set code block background and badge background on the right sidebar to `#061621`. This also affects tabs on the right sidebar for the status codes, e.g. “200,” “400,” etc., as this is just part of Redoc’s theming. Let me know if you’d prefer me to change it back to dark gray to match the sidebar’s background color. 
* Set non-link blue and teal text to `#0D324F`. Let me know if I missed something! 
* Removed using Redoc’s default loading screen and recreated it locally to prevent the large spinner -> small spinner  

@mango-db 
- I don’t have an opinion on the current contrast. Maybe @allisonmui9 does? 
- Text in the text box looks good to me. It uses our Akzidenz font and 16px font size
- I don’t mind an accessibility audit in the future, but that would be out of scope for this PR. Just note that Redoc maintains the layout of the content rendered so there may be limitations to things we can change 
- Updated the patches on the right sidebar to also have a bold font weight

Thanks for all of your reviews so far!

allisonmui9 (commented):
Thanks Raymund! 
What do you think about this for inactive tab state? 

-- update in a bit--

Everything else looks good to me!

mango-db:
I'm noticing what _appears_ to be the FeedbackWidget flashing on the loading screen. Also, occasionally on refresh, the FeedbackWidget requests feedback (in the form of the bottom-of-page stars) during the loading form. Is there a way we can disable it entirely until the page is loaded? This may need to be a separate ticket. 

https://user-images.githubusercontent.com/58707794/122982928-6f769000-d369-11eb-8b46-52b2bdfbebbe.mov

https://user-images.githubusercontent.com/58707794/122983485-0e02f100-d36a-11eb-8c1c-d69039a166ea.mov

src/components/SidebarBack.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon from '@leafygreen-ui/icon';
+import { SideNavItem } from '@leafygreen-ui/side-nav';
+import Link from './Link';
+import { formatText } from '../utils/format-text';
+
+// Adapted from docs-nav branch
+// TODO: Resolve merge conflicts when merging docs-nav into master
+const SidebarBack = ({ border, enableGlyph = true, textOverride, title, url, ...props }) => {
```
mango-db, line 10: Is there a more descriptive component name e.g. "SidebarBackArrow" / "SidebarBackIcon" we could use?

rayangler:
> I'm noticing what _appears_ to be the FeedbackWidget flashing on the loading screen. Also, occasionally on refresh, the FeedbackWidget requests feedback (in the form of the bottom-of-page stars) during the loading form. Is there a way we can disable it entirely until the page is loaded? This may need to be a separate ticket.

The feedback widget almost always flashes like that on just about every page that uses the feedback widget. We could make a separate ticket but I think it would also make sense to have it be a requirement when building out the new feedback widget

src/components/SidebarBack.js
rayangler: It's less about the arrow and the icon and more about the functionality of it being a sort of back button. How does `SidebarBackButton` sound?

src/components/OpenAPI.js
```diff
@@ -1,16 +1,271 @@
...
+import styled from '@emotion/styled';
+import { uiColors } from '@leafygreen-ui/palette';
 import ComponentFactory from './ComponentFactory';
+import SidebarBack from './SidebarBack';
+import Spinner from './Spinner';
+import { theme } from '../theme/docsTheme';
+import { formatText } from '../utils/format-text';
 
-const OpenAPI = ({ nodeData: { argument, children, options = {} }, ...rest }) => {
+const NAVBAR_HEIGHT = 45;
```
mango-db, line 14: Heads up! docsTheme has a `navbar` `baseHeight` if that works for your use case!

src/components/OpenAPI.js
```diff
@@ -1,16 +1,271 @@
...
+    }
+  }
+
+  // Prevent long enum names from overlapping with right sidebar code blocks
+  table {
+    word-break: break-word;
+  }
+`;
+
+// Copied over from docs-nav; consolidate title style after docs-nav merge to master
```
mango-db, line 205: Perhaps add a `// TODO:` flag as well?

src/components/OpenAPI.js
```diff
@@ -1,16 +1,271 @@
...
+
+  // "Array of" keyword next to data types under query parameters
+  span.sc-GqfZa {
+    color: ${schemaDataTypeColor};
+  }
+
+  // Parenthesized data types under query paramters; ex - "(ObjectId)"
+  span.sc-dwfUOf {
+    color: ${schemaDataTypeColor};
+  }
```
mango-db, line 117: Is there any way we could apply `${schemaDataTypeColor}` to two or more of these items simultaneously? E.g. a classname or anything? [No is a fine answer]

src/components/OpenAPI.js
```diff
@@ -1,16 +1,271 @@
...
+    color: ${uiColors.green.dark2};
+  }
+  span.put {
+    border: ${badgeBorderType} ${uiColors.yellow.light2};
+    color: ${uiColors.yellow.dark2};
+  }
+  span.patch {
+    border: ${badgeBorderType} ${uiColors.yellow.light2};
+    color: ${uiColors.yellow.dark2};
+  }
```
mango-db, line 136: Can these be consolidated? ['No' is still a fine answer]

src/components/OpenAPI.js
```diff
@@ -1,16 +1,271 @@
...
+  margin: 180px 0 ${theme.size.medium} 0;
+`;
+
+const LoadingMessage = styled('div')`
+  font-size: ${theme.fontSize.h1};
+  margin-bottom: ${theme.size.small};
+`;
+
+const MenuTitle = styled('div')`
+  ${titleStyle}
```
mango-db, line 234: Since `${titleStyle}` CSS is only used in `MenuTitle`, maybe we should keep it here?

src/components/OpenAPI.js
```diff
@@ -1,16 +1,271 @@
...
+
+  margin: ${theme.size.medium} ${theme.size.default};
+`;
+
+const LoadingWidget = ({ className }) => (
+  <LoadingContainer className={className}>
+    <LoadingMessage>Loading</LoadingMessage>
+    <Spinner size={48} />
```
mango-db, line 242: Do we want to pull this from docsTheme size options? (Granted, there's only `large: '32px', xlarge: '64px'`... the only thing stopping me from adding `48px` in myself, is that I don't know what to call the size in-between `large` and `xlarge` 😿 )

src/components/OpenAPI.js
```diff
@@ -1,16 +1,271 @@
...
+  </LoadingContainer>
+);
+
+const MenuTitleContainer = ({ siteTitle, pageTitle }) => {
+  const docsTitle = siteTitle + ' Docs';
+  const text = <>&#8592; Back to {formatText(docsTitle)}</>;
+
+  return (
+    <>
+      <SidebarBack border={<Border />} enableGlyph={false} textOverride={text} title={docsTitle} url={'/'} />
```
mango-db, line 252: Since we pass in `textOverride={text}` here, do we also need `{textOverride || `Back to ${formatText(title)}`}` on line 16 of `SidebarBack.js`?

src/components/OpenAPI.js
```diff
@@ -22,15 +277,121 @@ const OpenAPI = ({ nodeData: { argument, children, options = {} }, ...rest }) =>
   if (!specOrUrl) {
     return null;
   }
+  const pageTitle = page?.options?.title || '';
+  const tempLoadingDivClassName = 'openapi-loading-container';
+  const siteTitle = metadata?.title;
```
mango-db, line 282: Do we need to handled for the case that `siteTitle` is `undefined`? E.g. 

`const siteTitle = metadata?.title || '';`

src/components/OpenAPI.js
```diff
@@ -1,16 +1,271 @@
...
+
+const LoadingWidget = ({ className }) => (
+  <LoadingContainer className={className}>
+    <LoadingMessage>Loading</LoadingMessage>
+    <Spinner size={48} />
+  </LoadingContainer>
+);
+
+const MenuTitleContainer = ({ siteTitle, pageTitle }) => {
+  const docsTitle = siteTitle + ' Docs';
```
mango-db, line 247: Do we want to handle for `siteTitle` `undefined`?

src/components/OpenAPI.js
```diff
@@ -22,15 +277,121 @@ const OpenAPI = ({ nodeData: { argument, children, options = {} }, ...rest }) =>
...
+            colors: {
+              error: {
+                main: uiColors.red.dark1,
+              },
+              http: {
+                get: uiColors.blue.light3,
+                post: uiColors.green.light3,
+                put: uiColors.yellow.light3,
+                patch: uiColors.yellow.light3,
+                delete: uiColors.red.light3,
```
mango-db, line 330: I'm a little confused as to why these colors (and generally all the other styles in this options) seem to be defined twice. Once, here; and also, above with `globalCSS`

src/components/OpenAPI.js
rayangler: I don't think we have too many related things of size `48px` at the moment anyways. Want to circle back on this when we have a more defined name for the in-between size?

src/components/OpenAPI.js
rayangler: I'd prefer to keep it as it would preserve the original function of the component. It would also be useful to keep for when I merge it into the `docs-nav` branch

src/components/OpenAPI.js
rayangler: Ideally, `siteTitle` should never be `undefined` since this should be the same title in the repo's `snooty.toml`. Just in case, I added a check for it in the `MenuTitleContainer` that uses the site title

src/components/OpenAPI.js
rayangler: This option only affects the background color of each badge. The one added through global css changes the text and border colors.

Redoc's theme options seem to be pretty limiting on what we could change, and for which components. To get around this, the global css will fill in any of the gaps for the components that we want. It's not the most robust but it's what works with our current implementation

src/components/OpenAPI.js
```diff
@@ -22,15 +265,122 @@ const OpenAPI = ({ nodeData: { argument, children, options = {} }, ...rest }) =>
...
+    <>
+      <Global styles={globalCSS} />
+      {/* Temporary loading widget to be removed once the Redoc component loads */}
+      <LoadingWidget className={tempLoadingDivClassName} />
+      <RedocStandalone
+        onLoaded={() => {
+          if (load) {
+            return;
+          }
+          // Remove temporary loading margin from DOM
```
mango-db, line 282: Is this removing a margin, or the `LoadingWidget`?

mango-db (approved):
LGTM! Thanks Raymund! :)

src/components/OpenAPI.js
rayangler: Thanks for catching this outdated comment!

PR #419, opened 2021-05-25
DOP-2210: Match legacy sidebar breakpoint with snooty's

rayangler:
### Stories/Links:

DOP-2210

### Staging Links:

[Realm Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/realm/raymundrodriguez/DOP-2210/)

### Notes:
* Legacy css for the sidebar had the component's breakpoint at 1120px while [snooty's is 1024px](https://github.com/mongodb/snooty/blob/master/src/templates/document.js#L22). Changing the legacy css to match 1024px seems to solve this issue.
* To test, resize viewport to be between 1024px and 1120px. The left nav should be visible by default. Compare with the current live site for drivers ([link](https://docs.mongodb.com/drivers/c/)) and note that the sidebar no longer disappears.

### DIFF
--- a/src/styles/mongodb-docs.css
+++ b/src/styles/mongodb-docs.css
@@ -11993,26 +11993,15 @@ div.admonition {
     margin-left: 25px;
   }
 }
-@media (max-width: 1120px) {
-  .expand-toc-icon {
+@media (max-width: 1024px) {
+  #showNav {
+    cursor: pointer;
+    position: fixed;
     display: block;
-    padding-top: 10px;
-    padding-right: 10px;
-    color: #fff;
   }
-  .expand-toc-icon:active,
-  .expand-toc-icon:hover {
-    color: #fff;
-    text-decoration: none;
-  }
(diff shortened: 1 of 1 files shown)


casthewiz (approved):
*Acceptance*
Capture

*Production*
Capture2

🚀

mango-db (approved):
LGTM on the standard browsers, & thanks!

src/styles/mongodb-docs.css
```diff
@@ -12241,15 +12241,6 @@ div.admonition {
     margin-right: 15px;
     float: none;
   }
-  #showNav {
```
mango-db, line 12244: Nit picking, I wonder if we could move all or a substantial part of the original `@media (max-width: 1120px)` CSS block?

rayangler: Done! Thanks for pointing that out. I didn't see those classes in our codebase and things seem to be okay on staging

PR #421, opened 2021-05-26
Dop 2211: Fix the floating version select

casthewiz:
### Stories/Links:

DOP-2211

### Staging Links:

https://docs-mongodbcom-staging.corp.mongodb.com/master/bi-connector/cassidy.schaufele/DOP-2211/

### Notes:
Looks like this regression was caused by a [change](https://github.com/mongodb/leafygreen-ui/commit/857a680a03d3523aafcc8773c6a23ca0cc1422a7#) in leafygreen that changed default value for the `usePortal` prop on the `Select`, which changed default behavior for the `Select` component when we bumped versions.

Fixed version:
Screen Shot 2021-05-26 at 2 33 29 PM

### DIFF
--- a/src/components/TableOfContents.js
+++ b/src/components/TableOfContents.js
@@ -13,7 +13,6 @@ const TableOfContents = ({ activeSection, height, toctreeData: { children } }) =
       className="current"
       css={css`
         height: calc(100% - ${height}px);
-        z-index: -1;
         overflow-y: auto;
         position: absolute;
         width: 100%;
--- a/src/components/TableOfContents.js
+++ b/src/components/TableOfContents.js
@@ -14,7 +14,6 @@ const TableOfContents = ({ activeSection, height, toctreeData: { children } }) =
       css={css`
         height: calc(100% - ${height}px);
         overflow-y: auto;
-        position: absolute;
         width: 100%;
       `}
     >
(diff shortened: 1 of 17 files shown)


src/components/TableOfContents.js
```diff
@@ -13,6 +13,7 @@ const TableOfContents = ({ activeSection, height, toctreeData: { children } }) =
       className="current"
       css={css`
         height: calc(100% - ${height}px);
+        z-index: -1;
```
rayangler, line 16: It seems like applying the -1 z-index to the toc `ul` causes the links to be unclickable. Is it possible to instead add a z-index of 1 to the `StyledVersionDropdown` instead?

casthewiz: Hmmm. Having a non-negative z-index on the toc `ul` results in this UI regardless of z-index for the absolute positioned container on the `VersionDropdown`. 

Screen Shot 2021-05-26 at 3 50 08 PM

I'm going to move this to draft status and dig a little deeper.

casthewiz:
So in a nutshell - nested z-indexes were conflicting with specified z-indexes for the popover element.

I've resolved by adding a style applying to buttons in the select (the popover, when not a `Portal`, is nested under a `button` with z-index of 0) in addition to upgrading our version selector to gain access to a z-index prop on the `Select` itself.

Explicit ToC z-index has been removed.

sophstad (approved):
Works great! I wonder if this z-index sorcery will be necessary when we move to LeafyGreen's SideNav package—worth keeping an eye out for.

Thanks Cassidy!

rayangler (approved):
LGTM! Thanks for looking into the z-index fiasco

PR #410, opened 2021-05-14
DOP-2108: Fix code block line numbers

sophstad:
### Stories/Links:

DOP-2108

### Staging Links:

[Atlas Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/cloud-docs/sophstad/DOP-2108/best-practices-connecting-to-aws-lambda/#connection-example)

### Notes:

- Remove some styles that were causing issues with LG code blocks. A previous fix is incompatible new versions of LG's `Code`, so I've opted instead to remove the old CSS directly (plus some more for good measure, now that LG should be managing our tables and code blocks entirely).

PR #423, opened 2021-06-02
DOP-2219: Add .live banner

rayangler:
### Stories/Links:

DOP-2219

### Staging Links:

[Atlas](https://docs-mongodbcom-staging.corp.mongodb.com/master/cloud-docs/raymundrodriguez/DOP-2219/)
[Realm](https://docs-mongodbcom-staging.corp.mongodb.com/master/realm/raymundrodriguez/DOP-2219/)
[Landing](https://docs-mongodbcom-staging.corp.mongodb.com/master/landing/raymundrodriguez/DOP-2219/)

### Notes:
* Deleted old banner assets and replaced them with new ones.
* Re-enabled banner content.
* Re-add `max-width` to `.figure` and `img` that was previously removed ([link](https://github.com/mongodb/snooty/pull/410/files#diff-b421998eea4952dd4f1576d95f7abbdf2dd17b85786fd0c179c749c88ba94177L9620)). This fixes images on the homepage from being larger than they should be.


casthewiz (approved):
Capture
Approving, but the image we have for desktop isn't the best on widescreens - I imagine it's been like this in the past for other banners, though, so I wouldn't call it a blocking regression.

We might want to bump the height to 64px past a width of 1600px? 

64px height at 1600px width


64px height at 2450ish px width
25xx


(screenshots taken on a 27 inch monitor, 16:9 aspect ratio)

rayangler:
Just finished updating the test snapshots and staging links with the new height! I want to say that the height we set for the .live banner previously was to match past implementation of the .live banner on the https://www.mongodb.com/ site. But I think that's changed since then so I'm totally okay with the new heights we have!

Also fixed heading targets appearing under the top navbar (when a banner is present) and re-added `max-width: 100%` to figures and images to allow landing hero image to not appear full size.

mango-db:
Is there any way to handle the image text cutoffs? 

image

image

image

mango-db:
I'm getting a title cutoff for the "Connect to a Cluster" headerlink here, on smaller screen widths: https://docs-mongodbcom-staging.corp.mongodb.com/master/cloud-docs/raymundrodriguez/DOP-2219/connect-to-cluster/#connect-to-a-cluster

rayangler:
> I'm getting a title cutoff for the "Connect to a Cluster" headerlink here, on smaller screen widths.

Unfortunately, this is a problem that persists on our live site even without the banner (although probably not as apparent as when the banner is present). This only seems to happen to `h1`s at the top of the page anyways though. Have you been able to reproduce with a heading tag of a different size?

> Is there any way to handle the image text cutoffs?

I've been looking into it and it's kind of tricky to the get the dimensions right to have the .live banner to fit across all screen sizes without having anything get cut off. We currently don't have the luxury to change the height of the banner to retain the image's aspect ratio because of how we maintain the margins of the headerlink's targets (so that they appear below the navbar).

If it's any consolation, future .live banners and headerlinks may not have these issues when docs-nav releases because the navbar actually flows with the rest of the document now ([example](https://docs-mongodbcom-staging.corp.mongodb.com/upstream-nav/landing/raymundrodriguez/DOP-2161/) from docs-nav branch; click on the headings, banner still cuts off though because it has a similar implementation as on this branch, will update accordingly).

casthewiz (approved):
Approving with a caveat - it looked good across most emulated devices and viewports except for Ipad Mini.

Took a picture of the staging website on a physical Ipad Mini to confirm.
PXL_20210604_174856756


🤔 I think we might be off by 1px in our screenSize logic in docsTheme.js - but I think that should be investigated in a separate ticket, and shouldn't block shipping this.

PR #429, opened 2021-06-10
DOP-2235: Fix page scroll location when clicking on Links

rayangler:
### Stories/Links:

DOP-2235

### Staging Links:

[Landing](https://docs-mongodbcom-staging.corp.mongodb.com/upstream-nav/landing/raymundrodriguez/DOP-2235/)
[Realm](https://docs-mongodbcom-staging.corp.mongodb.com/master/realm/raymundrodriguez/DOP-2235/sdk/android/fundamentals/object-models-and-schemas/)
[Drivers](https://docs-mongodbcom-staging.corp.mongodb.com/master/drivers/raymundrodriguez/DOP-2235/)

### Notes:
* Fixed implementation of the `ContentTransition` that we have for navigating between pages.
* Utilized `gatsby-browser`'s [shouldUpdateScroll](https://www.gatsbyjs.com/docs/reference/config-files/gatsby-browser/#shouldUpdateScroll) function. Gatsby normally handles changing the scroll position of `window` when clicking on a `Link`. Because of our new grid layout on docs-nav, we no longer rely on scrolling `window`, but instead a nested div that acts as our scroll container for the page's content.


src/components/ContentTransition.js
```diff
@@ -1,15 +1,18 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { SwitchTransition, CSSTransition } from 'react-transition-group';
+import { CSSTransition, TransitionGroup } from 'react-transition-group';
```
mango-db, line 3: Non-blocking Q: Why did you change from SwitchTransition to TransitionGroup?

src/components/ContentTransition.js
```diff
@@ -1,15 +1,18 @@
...
 import { css, Global } from '@emotion/core';
+import { CONTENT_CONTAINER_CLASSNAME } from '../constants';
+
+const FADEOUT_DURATION = '100ms';
```
mango-db, line 7: Should `FADEOUT_DURATION` and `contentTransitionDuration` possibly be the same constant stored in `constants.js`?

gatsby-browser.js
```diff
@@ -1,9 +1,32 @@
 import React from 'react';
 import { ThemeProvider } from 'emotion-theming';
 import LeafyGreenProvider from '@leafygreen-ui/leafygreen-provider';
+import { CONTENT_CONTAINER_CLASSNAME } from './src/constants';
 import { theme } from './src/theme/docsTheme';
 import './src/styles/mongodb-docs.css';
 
+const contentTransitionDuration = 100;
+
+export const shouldUpdateScroll = ({ routerProps: { location } }) => {
```
mango-db, line 10: A small comment here explaining why we are implementing this plugin might be handy

mango-db: Esp. touching on the point of the "clicking on a new link" section of code + always returning false

src/components/ContentTransition.js
rayangler: Based on my understanding of the [React Transition Group docs](http://reactcommunity.org/react-transition-group/switch-transition), `SwitchTransition` is particularly good for changes in `state`, not necessarily content. From the page on `SwitchTransition`:

> Note: If you want the animation to happen simultaneously (that is, to have the old child removed and a new child inserted at the same time), you should use TransitionGroup instead.

Using `SwitchTransition` was also preventing Gatsby's `shouldUpdateScroll` function from being called and changing the scroll location, so I tried it with `TransitionGroup`.

gatsby-browser.js
```diff
@@ -1,9 +1,32 @@
...
+  const { hash } = location;
+  const scrollContainer = document.querySelector(`.${CONTENT_CONTAINER_CLASSNAME}`);
+
+  // Clicking on a new link
+  if (scrollContainer) {
+    if (hash) {
+      window.setTimeout(() => {
+        document.getElementById(decodeURI(hash).slice(1)).scrollIntoView(true);
```
mango-db, line 18: Would it be reasonable to catch a `URIError` from `decodeURI`? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI#exceptions

src/templates/blank.js
```diff
@@ -1,11 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import MainColumn from '../components/MainColumn';
-import { TEMPLATE_CLASSNAME } from '../constants';
 import landingStyles from '../styles/landing.module.css';
 
-const Blank = ({ children, className }) => (
-  <div className={`${TEMPLATE_CLASSNAME} ${className}`}>
```
mango-db, line 8: So were all these `${className}` s just useless the whole time? :o Nice catch

src/components/ContentTransition.js
```diff
@@ -19,7 +22,7 @@ const fadeIn = css`
   }
   .fade-enter-active {
     opacity: 1;
-    transition: opacity 200ms;
+    transition: opacity 200ms ${FADEOUT_DURATION};
```
mango-db, line 25: Still not sure why there are two durations (`200ms` and `${FADEOUT_DURATION}`) in this line! Everything I see online seems to have a single value

src/components/VersionDropdown.js
```diff
@@ -105,7 +105,6 @@ const VersionDropdown = ({
       onChange={navigate}
       placeholder={null}
       size={Size.Large}
-      usePortal={false}
```
mango-db, line 108: Why can we remove this?

mango-db (changes requested):
Wheeeeeew! The following LGTM:  
- [x] Staging LGTM on Firefox, Safari mobile/desktop
- [x] Link routing behaves appropriately with and without header links
- [x] Renames & pulls CONTENT_CONTAINER_CLASSNAME out of templates and into the 
- [x] Removes unnecessary React fragments from templates
- [x] Uses TransitionGroup instead of SwitchTransition
- [x] Uses shouldUpdateScroll plugin instead of onClick={handleLinkClick}
- [x] Footer moves to be included within Template to have content transition apply

Looking for:
- [] Removal of flashing hero image on lower half of page on content transition from landing site index to subpages like "Develop Applications"

src/components/VersionDropdown.js
rayangler: Due to both `master` and `docs-nav` adding the same `usePortal` prop to this component (I guess on two different lines), when `master` was merged to `docs-nav`, the same prop was being passed twice. This PR is just removing one of the two instances we're using this prop that I didn't catch before when merging

src/templates/blank.js
rayangler: Yup. I think we were just going for a different approach that once worked, and just didn't catch removing/updating it when we changed things around

src/theme/docsTheme.js
```diff
@@ -53,9 +53,15 @@ const header = {
   navbarHeight: '45px',
 };
 
+const transitionSpeed = {
+  contentFadeOut: '100ms',
+  contentFadeIn: '200ms',
+};
+
```
mango-db, line 60: Noice 👍

casthewiz (approved):
One code style nitpick w/ a suggestion, but nonblocking. 

Overall LGTM! 🚀

mango-db (approved):
🚗 💨💨💨  leaving us in the dust with your excellent work

gatsby-browser.js
```diff
@@ -1,9 +1,42 @@
...
+        } catch (e) {
+          console.error(e);
+        }
+      }, contentTransitionDuration);
+    } else {
+      window.setTimeout(() => {
+        scrollContainer.scrollTop = 0;
+      }, contentTransitionDuration);
+    }
+  }
```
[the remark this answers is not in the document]
rayangler: Thanks for this suggestion! Definitely looks a lot simpler to read and maintain. Re-built Realm staging to double-check and will merge in by end of day

PR #432, opened 2021-06-16
DOP-1436 Use LeafyGreen Modal in Lightbox component

carooobregon:
### Stories/Links:

DOP-1436

### Staging Links:
[Docs Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/docs/carolinao/DOP-1436/indexes/)

### Notes:
Changed the current `Lightbox` modal to the `LeafyGreen` modal, removed some legacy CSS to modern `Emotion` styled elements


rayangler (changes requested):
Changes look good so far! I left some comments that might be worth looking into before merging

src/components/Lightbox.js
```diff
@@ -1,63 +1,27 @@
...
-          <Image nodeData={nodeData} isLightboxOpen={false} />
+        <div className="lightbox__imageWrapper" onClick={() => setOpen((curr) => !curr)} role="button" tabIndex="-1">
+          <Image nodeData={nodeData} />
           <div className="lightbox__caption">{CAPTION_TEXT}</div>
         </div>
         <CaptionLegend {...rest} nodeData={nodeData} />
       </div>
-      {showModal && (
-        <div
-          className="lightbox__modal"
```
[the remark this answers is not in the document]
carooobregon: Deleted classes relating to lightbox that were only used in this component and applied styles using Emotion

src/styles/mongodb-docs.css
```diff
@@ -9728,8 +9728,7 @@ table.highlighttable td {
 }
 .deluge .rangestar:hover,
 .deluge .switch:hover,
-.deluge-vote-button:hover,
-.lightbox:hover {
+.deluge-vote-button:hover{
```
rayangler, line 9731: [nit] We can add a space between the selector and the curly brace here to keep it the same form

tests/unit/Lightbox.test.js
```diff
@@ -19,37 +23,34 @@ describe('Lightbox', () => {
...
+    const modalOpener = wrapper.find('Image');
     modalOpener.simulate('click');
+    expect(wrapper.exists('Modal')).toBe(true);
+    expect(wrapper.find('Modal').prop('open')).toBe(true);
   });
 
-  it('closes the modal', () => {
-    expect(wrapper.find('.lightbox__modal')).toHaveLength(0);
+  it('clicking anywhere outside of the modal', () => {
+    const lightboxwrapper = wrapper.find('LightboxCaption');
```
rayangler, line 52: [nit] We can use camelCase for this variable

src/components/Lightbox.js
```diff
@@ -1,63 +1,53 @@
...
-  };
-
-  const handleOnKeyDown = (e) => {
-    // Escape key
-    if (e.keyCode === 27) {
-      toggleShowModal();
+const StyledModal = styled(Modal)`
+  @media ${theme.screenSize.largeAndUp} {
+    div[role='dialog'] {
+      width: 40%;
```
rayangler, line 13: It seems like the media query works well for smaller images when the screen size is very large, but not as well on smaller screen sizes. [Example](https://docs-mongodbcom-staging.corp.mongodb.com/master/docs/carolinao/DOP-1436/tutorial/geospatial-tutorial/): an already large image gets significantly smaller.

Screen Shot 2021-06-21 at 5 20 08 PM

Is there another possible approach we could take for more varying screen sizes?

rayangler (commented):
Thanks for updating the css to use Emotion for the Lightbox component! Also, great work on getting the tests to pass.

I left a few nitpicky comments that may help to maintain the current style of our code. Also left a couple comments on the behavior of modals for different screen sizes

src/components/Lightbox.js
carooobregon: We could leave the modal size 80% for screen sizes large and up (the same way we had it in the last PR) and 100% for screen sizes smaller than that so they can take up the maximum amount of space (considering they don't have a lot of space to expand into), what do you think?

src/components/Lightbox.js
```diff
@@ -1,63 +1,60 @@
...
 
+const Lightbox = ({ nodeData, ...rest }) => {
+  const [open, setOpen] = useState(false);
   return (
     <React.Fragment>
-      <div className="figure lightbox" style={{ width: getNestedValue(['options', 'figwidth'], nodeData) || 'auto' }}>
-        <div className="lightbox__imageWrapper" onClick={toggleShowModal} role="button" tabIndex="-1">
-          <Image nodeData={nodeData} isLightboxOpen={false} />
-          <div className="lightbox__caption">{CAPTION_TEXT}</div>
+      <LightboxWrapper nodeData={nodeData}>
```
rayangler, line 48: We could pass in a prop for the figure width as `nodeData?.options?.figwidth` instead of passing the entire `nodeData` object here

src/components/Lightbox.js
```diff
@@ -1,63 +1,60 @@
...
+`;
 
-  // Hook to take effect with every re-render
-  useEffect(() => {
-    if (modal.current) {
-      modal.current.focus();
-    }
-  });
+const LightboxWrapper = styled('div')`
+  width: '${(props) => props.nodeData?.options?.figwidth || 'auto'}'
```
rayangler, line 38: It seems like the `width` isn't being applied here on staging. Adding the semicolon at the end may fix this

src/components/Lightbox.js
```diff
@@ -1,21 +1,28 @@
...
 import Image from './Image';
 import { theme } from '../theme/docsTheme.js';
-import styled from '@emotion/styled';
 
 const CAPTION_TEXT = 'click to enlarge';
 const StyledModal = styled(Modal)`
   @media ${theme.screenSize.largeAndUp} {
     div[role='dialog'] {
-      width: 40%;
+      width: 80%;
```
rayangler, line 13: This is a reasonable approach and it seems to work well!

tests/unit/Lightbox.test.js
```diff
@@ -1,11 +1,15 @@
...
 import { mount, shallow } from 'enzyme';
 import Lightbox from '../../src/components/Lightbox';
-
 // data for this component
 import mockData from './data/Figure.test.json';
 
-const mountLightbox = (nodeData) => mount(<Lightbox nodeData={nodeData} />);
+const mountLightbox = (nodeData) =>
+  mount(
+    <div className="lightbox_imagewrapper">
```
rayangler, line 9: Is this container still necessary since we're relying on clicking on `LightboxCaption` for clicking outside of the modal?

tests/unit/Lightbox.test.js
```diff
@@ -19,37 +23,34 @@ describe('Lightbox', () => {
...
   });
 
   it('shows a caption', () => {
-    expect(wrapper.find('.lightbox__caption')).toHaveLength(1);
-    expect(wrapper.find('.lightbox__caption').text()).toEqual('click to enlarge');
+    expect(wrapper.find('LightboxCaption')).toHaveLength(1);
+    expect(wrapper.find('LightboxCaption').text()).toEqual('click to enlarge');
   });
 
   it('clicking the photo', () => {
```
rayangler, line 44: Can we update this test name to be more specific? Maybe something like 'clicking the photo opens the modal'

rayangler (commented):
This looks good! Just a couple comments on the unit test

tests/unit/Lightbox.test.js
carooobregon: Good catch! Not necessary at all, just deleted it :)

tests/unit/Lightbox.test.js
carooobregon: Updated it!

PR #444, opened 2021-07-12
DOP-2316 - Make Realm SDK card clickable around tags

carooobregon:
### Stories/Links:

DOP-2316

### Staging Links:

[Realm PLP ](https://docs-mongodbcom-staging.corp.mongodb.com/master/realm/carolinao/DOP-2316/)

### Notes:
Sending cardRef prop to `RefRole` to determine if `RefRole` is coming from `Card` so that corresponding style is applied accordingly
Card is clickable however tags are not leading to their anchor links


src/components/RefRole.js
```diff
@@ -1,18 +1,40 @@
...
 import ComponentFactory from './ComponentFactory';
 import Link from './Link';
 import { normalizePath } from '../utils/normalize-path';
+import { theme } from '../theme/docsTheme';
 
-const RefRole = ({ nodeData: { children, domain, fileid, name, url }, slug }) => {
+const CardRef = styled(Link)`
+  background: ${uiColors.gray.light3};
+  border-radius: ${theme.size.tiny};
+  border: 1px solid rgba(184, 196, 194, 0.48);
```
mango-db, line 13: Can you check the LeafyGreen palette for this color `rgba(184, 196, 194, 0.48)`?

mango-db: [Yes, IK I wrote this literally a week ago 😆 ]

package.json
```diff
@@ -52,7 +52,7 @@
     "@emotion/core": "^10.1.1",
     "@emotion/styled": "^10.0.27",
     "@leafygreen-ui/badge": "^4.0.3",
-    "@leafygreen-ui/banner": "^3.0.6",
+    "@leafygreen-ui/banner": "^3.0.8",
```
mango-db, line 55: Let's not upgrade the banner / `package-lock.json` in this PR! [`git checkout origin/master <filename>` should do it]

mango-db (changes requested):
Woohoo! Good work, very speedy :) 

1. I think this is a fine temporary workaround for `card-ref` styling, though we would need to eventually come back in to fix the problem of other usages of other RefRoles on the Card (like in text) being styled as `card-ref`s. I suspect that fix would require parser intervention to distinguish between ref-role and card-ref on the frontend, so this is fine in the meantime, as long as we acknowledge that this isn't intended to be a permanent solution.  It'd be a good idea to add a `// TODO: <untangle card-ref and refrole, blah blah, this has the side effect of changing all refroles on cards into card-refs>` somewhere, as well as a comment above the passing of `cardRef={true}` into `ComponentFactory`, briefly explaining why that is necessary.

2. Do you get any console warnings about the `cardRef` prop on [http://localhost:8000](http://localhost:8000) in `npm run develop` mode?

3. I noticed that when I click on the `card-ref`s in the staging link, they don't seem to be routing appropriately - can you confirm?

src/components/RefRole.js
carooobregon: it's gray.light1, should i change it?

package.json
carooobregon: omg noo not again okok let me exclude it from the pr

carooobregon:
> Woohoo! Good work, very speedy :)
> 
> 1. I think this is a fine temporary workaround for `card-ref` styling, though we would need to eventually come back in to fix the problem of other usages of other RefRoles on the Card (like in text) being styled as `card-ref`s. I suspect that fix would require parser intervention to distinguish between ref-role and card-ref on the frontend, so this is fine in the meantime, as long as we acknowledge that this isn't intended to be a permanent solution.  It'd be a good idea to add a `// TODO: <untangle card-ref and refrole, blah blah, this has the side effect of changing all refroles on cards into card-refs>` somewhere, as well as a comment above the passing of `cardRef={true}` into `ComponentFactory`, briefly explaining why that is necessary.
> 2. Do you get any console warnings about the `cardRef` prop on http://localhost:8000 in `npm run develop` mode?
> 3. I noticed that when I click on the `card-ref`s in the staging link, they don't seem to be routing appropriately - can you confirm?

1. Added the comments! Good idea :) 
2. Nope no console warnings
3. Yeah it's not routing appropiately, I think it's because the route gets overwritten from the onClick=> event on the Card component? I've looked around online but haven't found a solution
4. for some reason it's now showing the links inline? when the only thing i changed was the comments? do you have an idea why this could be happening?

src/components/RefRole.js
carooobregon: is ur thumbs up a yes? @mango-db

carooobregon:
> > Woohoo! Good work, very speedy :)
> > 
> > 1. I think this is a fine temporary workaround for `card-ref` styling, though we would need to eventually come back in to fix the problem of other usages of other RefRoles on the Card (like in text) being styled as `card-ref`s. I suspect that fix would require parser intervention to distinguish between ref-role and card-ref on the frontend, so this is fine in the meantime, as long as we acknowledge that this isn't intended to be a permanent solution.  It'd be a good idea to add a `// TODO: <untangle card-ref and refrole, blah blah, this has the side effect of changing all refroles on cards into card-refs>` somewhere, as well as a comment above the passing of `cardRef={true}` into `ComponentFactory`, briefly explaining why that is necessary.
> > 2. Do you get any console warnings about the `cardRef` prop on http://localhost:8000 in `npm run develop` mode?
> > 3. I noticed that when I click on the `card-ref`s in the staging link, they don't seem to be routing appropriately - can you confirm?
> 
> 1. Added the comments! Good idea :)
> 2. Nope no console warnings
> 3. Yeah it's not routing appropiately, I think it's because the route gets overwritten from the onClick=> event on the Card component? I've looked around online but haven't found a solution
> 4. for some reason it's now showing the links inline? when the only thing i changed was the comments? do you have an idea why this could be happening?

ok you can ignore 4, i forgot a newline in the rST so it was parsed wrong and rendered wrong too but its fixed now! still trying to get routing right

carooobregon:
> > > Woohoo! Good work, very speedy :)
> > > 
> > > 1. I think this is a fine temporary workaround for `card-ref` styling, though we would need to eventually come back in to fix the problem of other usages of other RefRoles on the Card (like in text) being styled as `card-ref`s. I suspect that fix would require parser intervention to distinguish between ref-role and card-ref on the frontend, so this is fine in the meantime, as long as we acknowledge that this isn't intended to be a permanent solution.  It'd be a good idea to add a `// TODO: <untangle card-ref and refrole, blah blah, this has the side effect of changing all refroles on cards into card-refs>` somewhere, as well as a comment above the passing of `cardRef={true}` into `ComponentFactory`, briefly explaining why that is necessary.
> > > 2. Do you get any console warnings about the `cardRef` prop on http://localhost:8000 in `npm run develop` mode?
> > > 3. I noticed that when I click on the `card-ref`s in the staging link, they don't seem to be routing appropriately - can you confirm?
> > 
> > 
> > 
> > 1. Added the comments! Good idea :)
> > 2. Nope no console warnings
> > 3. Yeah it's not routing appropiately, I think it's because the route gets overwritten from the onClick=> event on the Card component? I've looked around online but haven't found a solution
> > 4. for some reason it's now showing the links inline? when the only thing i changed was the comments? do you have an idea why this could be happening?
> 
> ok you can ignore 4, i forgot a newline in the rST so it was parsed wrong and rendered wrong too but its fixed now! still trying to get routing right

actually- it routes properly in Safari but not in Chrome nor Firefox 🤔

mango-db:
From Slack, for posterity: the Card routing issue is likely related the [safer Card routing ticket](https://jira.mongodb.org/browse/DOP-2261), which has more information in the comments. This "small change" covers a lot more ground than we originally anticipated! 😸

src/components/RefRole.js
mango-db: Generally (and I can't speak for everyone), 👍 = yes / I agree, 👎 = no / I disagree

src/components/RefRole.js
```diff
@@ -1,18 +1,46 @@
...
+  margin-bottom: ${theme.size.tiny};
+  margin-right: ${theme.size.tiny};
+  padding: ${theme.size.tiny};
+
+  &:after {
+    content: ' ➔';
+  }
+`;
+
+const stopPropagation = function (e) {
```
mango-db, line 28: nice 😎

mango-db (approved):
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/realm/cgoecknerwald/DOP-2316/) LGTM on:

- [x] Chrome
- [x] Safari
- [x] Firefox

In order to avoid having side effects (namely, applying the `card-ref` styling to regular cards that use regular `ref`s), I did a quick search (`grep -r ".. card-group::" . `) of the other cards used in the documentation, and checked them for usage of `ref`s / any unexpected behavior with the propagation.

The current usages of `card-group`s (and therefore `card`s) are: 


### Product Landing Page cards
[./docs-realm/source/index.txt](https://docs-mongodbcom-staging.corp.mongodb.com/master/realm/cgoecknerwald/DOP-2316/)
[./docs-mcli/source/index.txt](https://github.com/10gen/docs-mongocli/blob/master/source/index.txt#L83)
[./docs-compass/source/index.txt](https://github.com/mongodb/docs-compass/blob/master/source/index.txt#L121)
[./docs-bi-connector/source/index.txt](https://docs-mongodbcom-staging.corp.mongodb.com/master/bi-connector/cgoecknerwald/DOP-2316/)

### Other sources
[./docs-realm/source/sdk.txt](https://docs-mongodbcom-staging.corp.mongodb.com/master/realm/cgoecknerwald/DOP-2316/sdk/)
[./docs-landing/source/index.txt](https://github.com/mongodb/docs-landing/blob/master/source/index.txt#L17)
[./docs-bi-connector/source/client-applications.txt](https://docs-mongodbcom-staging.corp.mongodb.com/master/bi-connector/cgoecknerwald/DOP-2316/client-applications/)

src/components/RefRole.js
carooobregon: hahahahah kk its fixed nowww

PR #449, opened 2021-07-20
DOP-2322 & DOP-2323 -  Implement QuizChoice component layout

carooobregon:
apologies for the long PR 😭  but this is a fun one 👍 😄 

### Stories/Links:

DOP-2322 
DOP-2323

### Staging Links:

[Random article I added a quiz widget to
](https://docs-mongodbcom-staging.corp.mongodb.com/master/cloud-docs/carolinao/dop-2322/mongodb-logs/)
### Notes:

Implemented QuizChoice with their respective styling depending on different states: Hovered, Selected, Submitted wrong question, Submitted right question
Sent setState callback functions as props to pass state between components


carooobregon:
github exposing all my commits 😭

src/components/QuizChoice.js
```diff
@@ -0,0 +1,83 @@
...
+  ${(props) =>
+    props.hasSubmitted
+      ? props.isCurrentChoiceCorrect
+        ? `border-color: ${uiColors.green.base}; border-width: 2px;`
+        : props.selectedThisChoice
+        ? `border-color: ${uiColors.black}; opacity: 0.5;`
+        : `opacity: 0.5;`
+      : `&:hover {
+        border-color: black;
+      }`}
```
casthewiz, line 25: I would look to unravel this nested ternary, or otherwise move some of the statements out of this block - it's a little hard to read and keep straight which styles are applied with which combo of boolean states

src/components/QuizChoice.js
```diff
@@ -0,0 +1,83 @@
...
+      {description.map((node, i) => (
+        <ComponentFactory nodeData={node} key={i} />
+      ))}
+    </DescriptionBody>
+  );
+};
+
+const QuizChoice = ({ nodeData: { argument, children, options }, selectedResponse, callback, idx, hasSubmitted }) => {
+  const description = children[0].children;
+  const isCurrentChoiceCorrect = options?.['is-true'] ? true : false;
```
casthewiz, line 59: total nitpick

but we can shorthand cast to a boolean with
`!!options?.['is-true']`

src/components/QuizWidget.js
```diff
@@ -61,17 +73,55 @@ const QuizCompleteSubtitle = ({ question }) => {
...
+
+  const footerWidget = hasSubmitted ? (
+    selectedResponse.isCurrentChoiceCorrect ? (
+      <ResultFooter />
+    ) : (
+      ''
+    )
+  ) : (
+    <SubmitButton setHasSubmitted={setHasSubmitted} />
+  );
```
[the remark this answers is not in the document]
carooobregon: nice i didn't know i could do that thank you!

src/components/QuizChoice.js
```diff
@@ -0,0 +1,83 @@
...
+import Card from '@leafygreen-ui/card';
+import { uiColors } from '@leafygreen-ui/palette';
+import ComponentFactory from './ComponentFactory';
+
+const StyledCard = styled(Card)`
+  box-shadow: none;
+  margin: auto auto 16px auto;
+  padding: 15px;
+  ${(props) => (props.selectedThisChoice ? `border-color: ${uiColors.black};` : ``)}
+  ${(props) => (props.hasSubmitted ? `pointer-events: none; ` : ``)}
```
rayangler, line 14: (Nit) Instead of returning an empty string, we could do something like:
```
${(props) => props.hasSubmitted && 'pointer-events: none;'}
```

src/components/QuizChoice.js
carooobregon: separated them into different statements! lmk if those are good enough

src/components/QuizChoice.js
carooobregon: looks so much better!

carooobregon:
fixed ternary ops and @allisonmui9 did some styling changes! [figma](https://www.figma.com/file/yYJuLvrlU1JGfiX1kvJrOW/Quiz-Widget?node-id=0%3A1)

allisonmui9 (changes requested):
Hi Caro! This is looking great. Thanks for making those changes! Some minor comments: 
- When an answer is correct, can we please remove the incorrect answer label at the bottom of the screen? We won't need that! 
- When choosing an answer and hitting submit, there's a slight pause where the stroke flickers. Do you know why that might be? 
- Similarly, when user hits submit, there's a brief lag / flicker and the x position of the text changes by maybe 1 / 2 pixels.  
- When the explanations are shown, can you make only the correct answer at full opacity? Looks like it's getting the same opacity treatment as the incorrect answers. 

Otherwise, looks amazing. :)

carooobregon:
> Hi Caro! This is looking great. Thanks for making those changes! Some minor comments:
> 
> * When an answer is correct, can we please remove the incorrect answer label at the bottom of the screen? We won't need that!
> * When choosing an answer and hitting submit, there's a slight pause where the stroke flickers. Do you know why that might be?
> * Similarly, when user hits submit, there's a brief lag / flicker and the x position of the text changes by maybe 1 / 2 pixels.
> * When the explanations are shown, can you make only the correct answer at full opacity? Looks like it's getting the same opacity treatment as the incorrect answers.
> 
> Otherwise, looks amazing. :)

thanks for your comments ! removed the incorrect answer label, changed opacity and was able to fix the border flicker w @rayangler 's help ! lmk if there's any additional styling changes to be done

rayangler (commented):
Just a couple of nitpicky comments and one question!

src/components/QuizWidget.js
```diff
@@ -61,17 +62,35 @@ const QuizCompleteSubtitle = ({ question }) => {
...
     question?.type === 'paragraph' && (
       <StyledCard>
         <QuizCompleteHeader />
         <QuizCompleteSubtitle question={question.children} />
         {choices.map((node, i) => (
-          <ComponentFactory nodeData={node} key={i} />
+          <ComponentFactory
+            nodeData={node}
+            key={i}
+            idx={i + 1}
```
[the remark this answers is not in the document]
carooobregon: Ahh it's `i+1` bc when I was doing `const selectedThisChoice = selectedResponse === idx;` on `QuizChoice.js` it was returning false when idx was 0, open to suggestions tho!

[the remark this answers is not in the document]
rayangler: Definitely weird! I would expect 0 === 0 to equal true 😅 . If this way works, I think it's fine then

src/components/QuizChoice.js
```diff
@@ -0,0 +1,97 @@
...
+      {hasSubmitted && <AnswerDescription description={description} />}
+    </StyledCard>
+  );
+};
+
+QuizChoice.propTypes = {
+  nodeData: PropTypes.shape({
+    children: PropTypes.arrayOf(PropTypes.object).isRequired,
+  }).isRequired,
+};
```
rayangler, line 95: Should we include the other props for `QuizChoice` here?

src/components/QuizWidget.js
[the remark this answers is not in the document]
carooobregon: You're right ! I think the issue was more when I was sending just the 0 in 
```
      onClick={() =>
        !selectedThisChoice ? callback({ index: idx, isCurrentChoiceCorrect: isCurrentChoiceCorrect }) : callback()
      }
```
but since I'm now sending an object it doesn't negate when it's 0. thanks for the catch :)

src/components/QuizChoice.js
```diff
@@ -0,0 +1,114 @@
...
+  setSelectedResponse,
+  idx,
+  hassubmitted,
+}) => {
+  const description = children[0].children;
+  const iscurrentchoicecorrect = !!options?.['is-true'];
+  const selectedthischoice = selectedResponseIdx === idx;
+  return (
+    <StyledCard
+      iscurrentchoicecorrect={iscurrentchoicecorrect ? 1 : 0}
```
carooobregon, line 82: I was receiving this warning when i just sent it as `iscurrentchoicecorrect={iscurrentchoicecorrect}` 
```
Warning: Received `false` for a non-boolean attribute `comingsoon`.

If you want to write it to the DOM, pass a string instead: 
comingsoon="false" or comingsoon={value.toString()}.
```
and i found this solution in [StackOverflow](https://stackoverflow.com/questions/49784294/warning-received-false-for-a-non-boolean-attribute-how-do-i-pass-a-boolean-f), does that sound right?

rayangler: Styling an existing React component and then passing props to conditionally style it can get kind of tricky because the props are still passed to the original component, and the prop can potentially make its way down to the html elements that make up the component (either intentionally or unintentionally).

To keep the camelCase for props and to prevent all of these errors and warnings of passing props and attributes to html elements, it might be worth just using emotion's `css` prop instead of `styled` for this case

carooobregon: great idea! using css instead of styled solved the problem :)

carooobregon:
Opened a new PR for this bc staging link was having some troubles. [Link to new PR](https://github.com/mongodb/snooty/pull/452)

PR #451, opened 2021-07-22
DOP-2338: Consistent Navigation PoC, 1.0.0 w/ feature flag

casthewiz:
### Stories/Links:

DOP-2338

### Staging Links:
Feature flag on:
https://docs-mongodbcom-staging.corp.mongodb.com/master/docs/cassidy.schaufele/con-nav-poc/

Feature flag off:
https://docs-mongodbcom-staging.corp.mongodb.com/master-flag-off/docs/cassidy.schaufele/con-nav-poc/

### Notes:
- Implements the consistent navigation component with a feature flag to toggle on/off behavior at build time of a docs property.
- Ridealong additional command at `npm build:clean` that runs `npm run clean && npm run build`


src/components/Navbar.js
```diff
@@ -140,37 +141,51 @@ const Navbar = () => {
   }, [isSearchbarDefaultExpanded]);
 
   return (
-    <ConditionalWrapper
-      condition={theme.bannerContent.enabled}
-      wrapper={(children) => (
+    <>
+      {process.env.FEATURE_FLAG_CONSISTENT_NAVIGATION ? (
```
casthewiz, line 145: This env var has to be set to `truthy` within the autobuilder's instances or within `.env` files during development in order for the consistent navigation to render.

If `undefined`, `falsey`, or simply not present in any configuration, our original nav will render.

package.json
```diff
@@ -48,6 +49,9 @@
       "npm run lint:fix"
     ]
   },
+  "publishConfig": {
+    "registry": "https://artifactory.corp.mongodb.com/artifactory/api/npm/npm-local/"
```
rayangler, line 53: Does this mean that we're deploying snooty to artifactory?

src/components/Navbar.js
rayangler: Great idea. Is there any more work that needs to be done on our end for the consistent nav? Or are we just waiting for the go signal to remove the flag?

src/components/Navbar.js
```diff
@@ -140,37 +141,51 @@ const Navbar = () => {
...
         <div
           css={css`
             position: fixed;
             top: 0;
+            width: 100%;
+            z-index: 2;
           `}
         >
-          <Banner {...theme.bannerContent} />
```
rayangler, line 152: Lint warnings on where we assign the banner! We probably won't have to set up another banner any time soon so I'm okay with removing its assignment as well

rayangler (commented):
Just a couple of questions for clarification!

package.json
casthewiz: Nope - I was instructed to add it when we first started integrating with the nav package, but I'm actually curious if it's needed. I'll look at dropping it. 🧐

casthewiz: I was able to safely drop it 🤷

rayangler:
I noticed that the "Try Free" anchor loads its green background with a delay (not sure why), and hovering over the link makes the text blue and underlines it. 

I'm convinced that the latter may be happening due to our global css file. Is there a way we can override the styling for that anchor tag specifically?

casthewiz:
> I noticed that the "Try Free" anchor loads its green background with a delay (not sure why), and hovering over the link makes the text blue and underlines it.
> 
> I'm convinced that the latter may be happening due to our global css file. Is there a way we can override the styling for that anchor tag specifically?

So the green background delay seems intentional - there's a transition effect on that part of the component. As far as overriding the styling, I've tried a few ways of resolving but none were really satisfactory. Either we risk leaking the overriding styles to other `a` tags in the shared component, or we risk having a "gotcha" style that overwrites any styles they might add in future versions.

I'm going to flag that `:hover` styles should be explicitly set on that button for future versions, to avoid this situation 👍

rayangler (approved):
LGTM! Thanks for looking into the styling

mango-db (approved):
LGTM!

Looks good on:
- [x] Safari
- [x] Chrome
- [x] Firefox

PR #454, opened 2021-07-26
DOP-2361 - Prepare Quiz Object to send to DB

carooobregon:
### Stories/Links:

DOP-2361
[Figma
](https://www.figma.com/file/yYJuLvrlU1JGfiX1kvJrOW/Quiz-Widget?node-id=0%3A1)
### Staging Links:

[Staging Link](https://docs-mongodbcom-staging.corp.mongodb.com/master/cloud-docs/carolinao/DOP-2361/mongodb-logs/)
### Notes:

There's like a hundred commits that are carried over from past branches so i'm sorry for that 😅   but for this PR I basically did a function called `get-nested-text.js` that gets nested plaintext from the AST so that we can the string of which question and which answer users select every time they click submit. Also prepared and organized the information for the document objects we'll be writing into the database in further PRs


src/components/QuizWidget.js
```diff
@@ -1,12 +1,14 @@
-import React, { useState } from 'react';
+import React, { useCallback, useState } from 'react';
 import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
 import Button from '@leafygreen-ui/button';
 import Card from '@leafygreen-ui/card';
 import { uiColors } from '@leafygreen-ui/palette';
 import Icon from '@leafygreen-ui/icon';
 import ComponentFactory from './ComponentFactory';
+import { getNestedText } from '../utils/get-nested-text';
```
rayangler, line 9: Just found out that we have a [getPlaintext](https://github.com/mongodb/snooty/blob/master/src/utils/get-plaintext.js) function! Would using this function work for you?

rayangler (commented):
Looks good so far! Just have a couple of comments

src/components/QuizWidget.js
carooobregon: Oh nice!! Tyty

rayangler (approved):
LGTM!

mango-db (commented):
### General comments/questions
Feel free to ask me if anything is unclear! I just took a brief scan, so I may be misunderstanding some things / not seeing the full context. My line numbers might also be off because of changes to the files.

1. Be sure to checkout docsTheme.js for [`size`/`fontSize` constants](https://github.com/mongodb/snooty/blob/master/src/theme/docsTheme.js#L14-L42)! I see a few places in the CSS where we can reuse certain sizes
2. General question: do we want to house the `QuizWidget` components within the components/widgets/ subdirectory, a la `FeedbackWidget`? [I'm personally ambivalent - this is for you to consider as an option]
3. [16.59](https://github.com/mongodb/snooty/blob/master/src/components/QuizWidget.js#L32) is a very specific line height! XD What's the reason for this?
4. Do [CorrectChoice and IncorrectChoice](https://github.com/mongodb/snooty/blob/master/src/components/QuizChoice.js#L57-L58) need to be functions?
5. How do you [handle `description` safely](https://github.com/mongodb/snooty/blob/master/src/components/QuizChoice.js#L75)? What happens if either of the two `children` are undefined? Also, why is the description so deeply nested?
6. What is the fail state of [`question?.type != paragraph`](https://github.com/mongodb/snooty/blob/master/src/components/QuizWidget.js#L81)? When do you expect that falsey condition to occur?
7. What do you think about changing booleans in `QuizChoice.js` from: [`selectedThisChoice` -> `isSelected`, `hasSubmitted` -> `isSubmitted`, `isCurrentChoiceCorrect` -> `isCorrect`]? Do you think that's appropriate, or too much information loss? 
8. Since you don't use options anywhere else, I think you can change [this](https://github.com/mongodb/snooty/blob/master/src/components/QuizWidget.js#L96) to retrieve `quizId` [like so](https://github.com/mongodb/snooty/blob/master/src/components/Twitter.js#L18).

PR #459, opened 2021-07-30
DOP-2325 - Set up Realm and write responses

carooobregon:
### Stories/Links:

DOP-2325
DOP-2381

### Staging Links:

[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/cloud-docs/carolinao/DOP-2325/mongodb-logs/)

### Notes:
Used the [realm-template-apps](https://github.com/mongodb-university/realm-template-apps/tree/main/generated/web-mql) to write the Provider and Context needed to set up Realm


src/components/Widgets/QuizWidget/RealmApp.js
```diff
@@ -0,0 +1,38 @@
...
+import { baseUrl } from './realm.json';
+
+function createRealmApp(id) {
+  return new Realm.App({ id, baseUrl });
+}
+const RealmAppContext = React.createContext(null);
+
+export function RealmAppProvider({ appId, children }) {
+  // Store Realm.App in React state. If appId changes, all children will rerender and use the new realmApp.
+  const [realmApp, setRealmApp] = React.useState(createRealmApp(appId));
```
rayangler, line 12: Is `setRealmApp` supposed to be used anywhere?

src/layouts/index.js
```diff
@@ -79,9 +81,11 @@ const DefaultLayout = (props) => {
       <Global styles={globalCSS} />
       <SiteMetadata siteTitle={title} />
       <TabProvider selectors={page?.options?.selectors}>
-        <ContentsProvider headingNodes={page?.options?.headings}>
-          <Template {...props}>{children}</Template>
-        </ContentsProvider>
+        <RealmAppProvider appId={appId}>
```
rayangler, line 84: What's the purpose of placing the quiz widget's context provider this high up vs. just wrapping the `QuizWidget` with this? Just wondering in case there was a specific reason!

Also, if this `RealmAppProvider` is specifically for the `QuizWidget`, do we need to pass the `appId` as a prop to the provider component instead of just keeping it inside of the component?

tests/unit/QuizWidget.test.js
```diff
@@ -8,10 +8,15 @@ import { completeQuiz, noQuestion } from './data/QuizWidget.test.json';
...
 const project = 'cloud-docs';
 
+const mockDate = new Date(1466424490000);
+const spy = jest.spyOn(global, 'Date').mockImplementation(() => mockDate);
+
 jest.mock('../../src/hooks/use-site-metadata', () => ({
   useSiteMetadata: () => ({ siteUrl, project }),
 }));
 
+beforeAll(() => {});
```
rayangler, line 18: Do we still need this here?

src/components/Widgets/QuizWidget/RealmApp.js
```diff
@@ -0,0 +1,38 @@
...
+  const logIn = React.useCallback(async () => {
+    await realmApp.logIn(Realm.Credentials.anonymous());
+    setCurrentUser(realmApp.currentUser);
+  }, [realmApp]);
+
+  useEffect(() => {
+    logIn();
+  }, [logIn]);
+
+  // Override the App's currentUser & logIn properties + include the app-level logout function
```
rayangler, line 24: Not sure if I'm misunderstanding this comment but what exactly is this "logout function"?

tests/unit/QuizWidget.test.js
```diff
@@ -21,3 +26,7 @@ it('doesnt render quiz widget without a specified question', () => {
   const tree = shallow(<QuizWidget nodeData={noQuestion} />);
   expect(tree).toMatchSnapshot();
 });
+
+afterAll(() => {
+  spy.mockRestore();
```
rayangler, line 31: This is a nice touch

rayangler (commented):
This is very cool. Just a few questions for now!

src/components/Widgets/QuizWidget/RealmApp.js
carooobregon: not right now! tbh i just thought every state had to have a `set` function, should i delete it?

src/layouts/index.js
carooobregon: 1. You're right! I just changed it so it's wrapped around `SubmitButton` since all the submit logic is inside there :-) 
2. In theory, `RealmAppProvider` can be reused in posteriority for other Realm apps we want to use in snooty, hence sending the `appId` as a prop, however as we're only using it for QuizWidget rn I could change it so that it gathers the `appId` from the `realm.json` as a constant inside of the component instead of sending it as a prop, what do you think is better?

tests/unit/QuizWidget.test.js
carooobregon: nope thx thx

src/components/Widgets/QuizWidget/RealmApp.js
carooobregon: oh it carried over from the [realm-template-apps](https://github.com/mongodb-university/realm-template-apps/tree/main/generated/web-mql) comments bc they had a logout function but i just deleted it !

src/layouts/index.js
rayangler: Ah, okay, thanks for the explanation! I think being able to reuse it would be cool, especially for the upcoming Feedback Widget, so sending in the id as a prop sounds good to me

src/components/Widgets/QuizWidget/RealmApp.js
rayangler: I think using it with `useEffect` like how they do [in the template](https://github.com/mongodb-university/realm-template-apps/blob/main/generated/web-mql/src/components/RealmApp.js#L13-L16) wouldn't be such a bad idea. Usually if we have a state, the expectation is that we have a corresponding `setState` function to update it

src/components/Widgets/QuizWidget/RealmFuncs.js
```diff
@@ -0,0 +1,17 @@
+import { useCollection } from './hooks/useCollection';
+import { dataSourceName } from './realm.json';
+
+export function useRealmFuncs() {
+  const responseCollection = useCollection({
+    cluster: dataSourceName,
+    db: 'quiz_test',
```
mango-db, line 7: Is this db supposed to be `quiz_test` in production?

carooobregon: ohh very true! just added `dbName` as a parameter in `useRealmFuncs` to support different databases to write to and also  added `const dbName = snootyEnv == "production" ? "quiz_prod" : "quiz_dev";` to decide which one to send, it seems as `snootyEnv` from `useSiteMetadata()` doesn't catch staging as a snootyEnv so I just left `quiz_dev` and `quiz_prod`

rayangler (approved):
LGTM, all of my comments/concerns have been resolved. Thank you!

src/components/Widgets/QuizWidget/QuizWidget.js
```diff
@@ -9,6 +9,9 @@ import ComponentFactory from '../../ComponentFactory';
 import { theme } from '../../../theme/docsTheme';
 import { useSiteMetadata } from '../../../hooks/use-site-metadata';
 import { getPlaintext } from '../../../utils/get-plaintext';
+import { useRealmFuncs } from './RealmFuncs';
```
mango-db, line 12: Nit: is there a more meaningful name we could give this function?

src/components/Widgets/QuizWidget/QuizWidget.js
```diff
@@ -78,23 +86,44 @@ const SubmitButton = ({ setIsSubmitted, selectedResponse, quizResponseObj }) =>
...
+};
+
+const getCorrectAnswerCount = (choices) => {
+  var correctAnswerCount = 0;
+  choices.forEach((node) => {
+    if (node.options?.['is-true']) {
+      correctAnswerCount += 1;
+    }
+  });
+  return correctAnswerCount;
```
mango-db, line 107: You may be able to one-liner this with something like:

`return choices.filter(c => !!c.options?.['is-true']).length === 1`

The filter can be used to reduce `choices` to an array of booleans. Then, you can take the array of that, and check that it is exactly equivalent to 1.

I haven't tested this so you may have to mess around with this ^^

mango-db (commented):
LGTM:

- [x] Correctly posts test data to quiz_dev in develop AND [staging](https://docs-mongodborg-staging.corp.mongodb.com/test-quiz-widget/cloud-docs/cgoecknerwald/realm-app/) mode, including correct and incorrect response
- [] Posts data to quiz_prod in production?
- [x] Functions on mobile

A thought: It would be interesting to see how this affects page load times + latency

src/components/Widgets/QuizWidget/QuizWidget.js
```diff
@@ -78,23 +86,44 @@ const SubmitButton = ({ setIsSubmitted, selectedResponse, quizResponseObj }) =>
   );
 };
 
-const createQuizResponseObj = (question, quizId, selectedResponse, project) => {
+const createQuizResponseObj = (question, quizId, selectedResponse, project, quizDate) => {
```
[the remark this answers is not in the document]
mango-db: Lol ^^ `:date:` field

src/components/Widgets/QuizWidget/QuizWidget.js
```diff
@@ -78,23 +86,44 @@ const SubmitButton = ({ setIsSubmitted, selectedResponse, quizResponseObj }) =>
...
     ...selectedResponse,
     questionText: getPlaintext(question.children),
     quizId: quizId,
     project: project,
+    timestamp: new Date(),
+    quizDate: quizDate,
+  };
+};
+
+const getCorrectAnswerCount = (choices) => {
```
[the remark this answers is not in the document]
rayangler: I agree that this should be enforced in the parser but I also think it might be fine to leave this logic here for now as well. It seems like there's a [ticket](https://jira.mongodb.org/browse/DOP-2386) that handles this check in the parser.

As a quick compromise, I'd say we can leave a comment here to let others know parser-level validation will be added later and we can check afterwards if this frontend check is still necessary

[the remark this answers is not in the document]
mango-db: Yeeee, I agree 🤠

carooobregon:
@mango-db thanks for all your comments!! responding here bc github won't let me reply to comments
**date issue**: good catch! added verification using regex
**What does the "-ftaro" in "quizwidget-ftaro" mean?**
its the unique realm ID for the quiz widget app that is generated by Realm
**re. useRealmFuncs**
so in theory realmFuncs can have insert functions, delete functions, etc, since it's essentially the middleware between a DB and the snooty app, not necessarily quizwidget specific so that’s why it’s pretty generic and most likely will be used in feedback too, that being said i changed some names to make it more generic and so it can be reused, lmk if thats okay

mango-db (approved):
💎 [ʕ•́ᴥ•̀ʔっ♡](https://docs-mongodbcom-staging.corp.mongodb.com/test-quiz-widget/cloud-docs/cgoecknerwald/realm-app/
) 💎

PR #482, opened 2021-08-27
DOP-2376: Treat sharedinclude as include

i80and:
### Stories/Links:

DOP-2376


src/components/ComponentFactory.js
```diff
@@ -148,6 +148,7 @@ const componentMap = {
   hlist: HorizontalList,
   image: Image,
   include: Include,
+  sharedinclude: Include,
```
rayangler, line 151: (Nit) We should alphabetize this based off of the key, not the value

rayangler (approved):
Tests are failing because of a current situation between our secrets, the CI, and forks. Tests and linting are fine locally, and this is a small change so I'm okay with approving.

casthewiz (approved):

PR #497, opened 2021-09-30
DOP-2495: Retain Realm Searchbar Context

casthewiz:
### Stories/Links:

DOP-2495

### Staging Links:

*Landing, with Realm searchProperty*
https://docs-mongodbcom-staging.corp.mongodb.com/master/landing/cassidy.schaufele/DOP-2495/search/?q=&searchProperty=realm-master

*Landing, default*
https://docs-mongodbcom-staging.corp.mongodb.com/master/landing/cassidy.schaufele/DOP-2495/search

### Notes:
We can be assured that for the version of the consistent navigation component that we utilize, the `searchProperty` will have a value of `realm-master` when a search originates from Realm docs.

This change ensures that we will retain Realm search as the default property for the navigation when originating search from said Realm pages.

### DIFF
--- a/src/components/Header.js
+++ b/src/components/Header.js
@@ -14,7 +14,10 @@ const StyledHeaderContainer = styled.header`
 
 const Header = ({ sidenav }) => {
   const { project } = useSiteMetadata();
-  const unifiedNavProperty = project === 'realm' ? 'REALM' : 'DOCS';
+
+  const searchProperty = new URL(window.location).searchParams.get('searchProperty');
+  const shouldSearchRealm = project === 'realm' || searchProperty === 'realm-master';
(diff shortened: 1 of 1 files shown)


mango-db (commented):
The code LGTM, but don't we want to implement this for all properties, not just Realm? I imagine the same flow occurs on Compass, etc...

casthewiz:
> The code LGTM, but don't we want to implement this for all properties, not just Realm? I imagine the same flow occurs on Compass, etc...

@mango-db This just sets the default selected property in the consistent navigation search - which is split between `All Documentation` and `Realm Documentation`.

mango-db (approved):
leafy green to me

PR #498, opened 2021-10-01
DOP-2136: Add LanguageSwitcher to code blocks

rayangler:
### Stories/Links:

DOP-2136

### Staging Links:

[Server Docs](https://docs-mongodbcom-staging.corp.mongodb.com/master/docs/raymundrodriguez/DOP-2136/indexes/)

### Notes:
* Leverages LG's language switcher for code blocks to work alongside our `TabsSelector` component. If there is no `tabs-selector` directive on the page, and if the code block isn't nested under a tab from the drivers tabset, then the language switcher will not be rendered.

### DIFF
--- a/src/components/Code.js
+++ b/src/components/Code.js
@@ -1,7 +1,9 @@
-import React, { useCallback } from 'react';
+import React, { useCallback, useContext } from 'react';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/core';
 import { default as CodeBlock, Language } from '@leafygreen-ui/code';
+import { CodeContext } from './code-context';
+import { TabContext } from './tab-context';
 import { theme } from '../theme/docsTheme';
 import { reportAnalytics } from '../utils/report-analytics';
 
@@ -19,7 +21,10 @@ const getLanguage = (lang) => {
 };
 
 const Code = ({ nodeData: { caption, copyable, emphasize_lines: emphasizeLines, lang, linenos, value } }) => {
+  const { setActiveTab } = useContext(TabContext);
+  const { languageOptions, codeBlockLanguage } = useContext(CodeContext);
   const code = value;
+  const language = (languageOptions?.length > 0 && codeBlockLanguage) || getLanguage(lang);
 
   const reportCodeCopied = useCallback(() => {
     reportAnalytics('CodeblockCopied', { code });
@@ -33,13 +38,29 @@ const Code = ({ nodeData: { caption, copyable, emphasize_lines: emphasizeLines, 
         min-width: 150px;
         table-layout: fixed;
         width: 100%;
+
+        // Inner div of LG component has a width set to 700px. Unset this as part of our
+        // override for docs when the language switcher is being used.
+        > div > div {
+          width: unset;
+        }
+
+        // Override default LG Code language switcher font size
+        button > div > div {
+          font-size: ${theme.fontSize.small};
+        }
       `}
     >
       <CodeBlock
         chromeTitle={caption}
         copyable={copyable}
         highlightLines={emphasizeLines}
-        language={getLanguage(lang)}
+        language={language}
+        languageOptions={languageOptions}
+        onChange={(selectedOption) => {
+          const tabsetName = 'drivers';
+          setActiveTab({ name: tabsetName, value: selectedOption.id });
+        }}
         onCopy={reportCodeCopied}
         showLineNumbers={linenos}
         showWindowChrome={caption ? true : false}
--- a/src/components/Code.js
+++ b/src/components/Code.js
@@ -47,7 +47,7 @@ const Code = ({ nodeData: { caption, copyable, emphasize_lines: emphasizeLines, 
 
         // Override default LG Code language switcher font size
         button > div > div {
-          font-size: ${theme.fontSize.small};
+          font-size: ${theme.fontSize.default};
         }
       `}
     >
--- a/src/components/code-context.js
+++ b/src/components/code-context.js
@@ -48,18 +48,15 @@ const getDriverImage = (driver, driverIconMap) => {
     return <DriverIcon />;
   }
 
-  // LG Code default file icon; not formally added in their code yet,
-  // but is mentioned in a source code comment
+  // Use LG File icon as our default placeholder for images. This overwrites
+  // LG's Language Switcher current default icon. See:
+  // https://github.com/mongodb/leafygreen-ui/blob/6041b89bf5f9dc1e5ea76018bc2cd84bc1fd6faf/packages/code/src/LanguageSwitcher.tsx#L135-L136
   return <LeafyIcon glyph="File" />;
 };
 
-// Generates language options for code block based on drivers tabset
-const generateLanguageOptions = (selectors, driverIconMap) => {
-  if (!(selectors && Object.keys(selectors).length > 0 && selectors['drivers'])) {
-    return null;
-  }
-
-  const drivers = selectors['drivers'];
+// Generates language options for code block based on drivers tabset on current page
+const generateLanguageOptions = (selectors = {}, driverIconMap) => {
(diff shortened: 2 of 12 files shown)


src/components/code-context.js
```diff
@@ -0,0 +1,101 @@
...
+const CodeContext = React.createContext(defaultContextValue);
+
+const LANGUAGES = new Set(Object.values(Language));
+
+// Custom mapping for unique drivers that don't have an exact 1:1 name to LG language mapping
+const DRIVER_LANGUAGE_MAPPING = {
+  clike: new Set(['c', 'cpp', 'c#']),
+  javascript: new Set(['compass', 'nodejs', 'shell']),
+  python: new Set(['motor']),
+};
```
mango-db, line 21: neat :0

src/components/code-context.js
```diff
@@ -0,0 +1,101 @@
...
+// shown on our TabSelector component
+const getDriverImage = (driver, driverIconMap) => {
+  const DriverIcon = driverIconMap?.[driver];
+  if (DriverIcon) {
+    return <DriverIcon />;
+  }
+
+  // LG Code default file icon; not formally added in their code yet,
+  // but is mentioned in a source code comment
+  return <LeafyIcon glyph="File" />;
```
mango-db, line 53: - What do you mean by "not formally added"? Does it exist at all?
- If isn't added, is there any chance of this being returned?
- We will need to update LG Code when this is released, right?

tests/unit/Code.test.js
```diff
@@ -1,25 +1,105 @@
...
-  expect(tree).toMatchSnapshot();
+  const wrapper = shallowCode({ data: mockData.testCode, activeTabs: { cloud: 'cloud' } });
+  expect(wrapper).toMatchSnapshot();
+  expect(wrapper.find('LanguageSwitcher').exists()).toBeFalsy();
 });
 
-it('renders with javascript disabled correctly', () => {
-  const tree = shallowCode({ data: mockData, activeTabs: {} });
-  expect(tree).toMatchSnapshot();
+describe('when rendering with selectors', () => {
```
mango-db, line 70: Any use for a "doesn't render when no selectors" test? Or is that handled by the language selector?

src/components/code-context.js
```diff
@@ -0,0 +1,101 @@
...
+      return language;
+    }
+  }
+
+  if (LANGUAGES.has(driver)) {
+    return driver;
+  }
+
+  // LG Code default language
+  return 'none';
```
mango-db, line 40: What happens when this is returned? / Why would this be returned?

mango-db (commented):
Leafy Green To Me 🍀 

- [x] Works as expected on Firefox, Safari, Chrome
- [x] Code blocks disappear when Compass selected
- [x] Refreshing page maintains language selection
- [x] Changing language from either code block or language switcher affects entire page

Note:
- Only weird thing I see is that when you select 'Compass' on the code block, the entire code block disappears, and one would have to scroll back up to the Language Selector to re-engage the code blocks. I feel this is probably not the slickest design design, and should be addressed in the future, but adequate for now

src/components/code-context.js
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  return languageOptions;
+};
+
+// Returns the display name of the current driver selected
+const getCurrentLanguageOption = (languageOptions, activeTabs) => {
+  const currentTab = activeTabs?.drivers;
+
+  if (!(languageOptions && languageOptions.length > 0 && currentTab)) {
```
casthewiz, line 81:
```suggestion
  if (!(languageOptions?.length > 0 && currentTab)) {
```

src/components/code-context.js
casthewiz: 🤔 I'm assuming we're referring to LG Code not having a default icon and this [reference](https://github.com/mongodb/leafygreen-ui/blob/6041b89bf5f9dc1e5ea76018bc2cd84bc1fd6faf/packages/code/src/LanguageSwitcher.tsx#L135-L136) in their source?

We might want to change the comment here to call out this explicitly as a placeholder, perhaps with a link to the above default that we want to override

src/components/code-context.js
```diff
@@ -0,0 +1,101 @@
...
+  // LG Code default file icon; not formally added in their code yet,
+  // but is mentioned in a source code comment
+  return <LeafyIcon glyph="File" />;
+};
+
+// Generates language options for code block based on drivers tabset
+const generateLanguageOptions = (selectors, driverIconMap) => {
+  if (!(selectors && Object.keys(selectors).length > 0 && selectors['drivers'])) {
+    return null;
+  }
```
casthewiz, line 60: Should we prefer a default param for `selectors` here instead of the null checks? Would `{}` work?

I think we could condense this check to `if (!selectors['drivers']) return;` if we use `{}` as a default param.

src/components/code-context.js
```diff
@@ -0,0 +1,101 @@
...
+const getCurrentLanguageOption = (languageOptions, activeTabs) => {
+  const currentTab = activeTabs?.drivers;
+
+  if (!(languageOptions && languageOptions.length > 0 && currentTab)) {
+    return null;
+  }
+
+  const currentLangOption = languageOptions.find((option) => option.id === currentTab);
+  // LG Code block's language switcher uses the display name to select the current "language"
+  return currentLangOption?.displayName || null;
```
casthewiz, line 87: I think we can just do `return currentLangOption?.displayName` here - unless we intend to explicitly check for ` === null` after this function's invocation

casthewiz (approved):
I just have some nitpicks, clarifications, and stylistic suggestions - otherwise I think we are 🟢 go 😄

src/components/code-context.js
rayangler: Yup, that's what I was referencing. Commenting that it's explicitly a placeholder sounds better. Thanks!

tests/unit/Code.test.js
rayangler: I want to say that our language selector always chooses a default driver/language by default if it is present. Rendering a code block without selectors present would technically be covered by the test above it (rendering with no `TabsProvider`)

src/components/code-context.js
rayangler: "none" is one of LG's supported languages for their Code component and it's explicitly included to prevent any syntax highlighting/coloring. I included it here as a default value for any drivers we may choose to add/include that don't have a language formally supported by LG.

src/components/Code.js
```diff
@@ -33,13 +38,29 @@ const Code = ({ nodeData: { caption, copyable, emphasize_lines: emphasizeLines,
...
+
+        // Inner div of LG component has a width set to 700px. Unset this as part of our
+        // override for docs when the language switcher is being used.
+        > div > div {
+          width: unset;
+        }
+
+        // Override default LG Code language switcher font size
+        button > div > div {
+          font-size: ${theme.fontSize.default};
```
rayangler, line 50: For whatever reason, I left the font size at small when making the PR but meant to make this default. 16px is the font size requested for the text in the language switcher.

rayangler (commented):
> Only weird thing I see is that when you select 'Compass' on the code block, the entire code block disappears, and one would have to scroll back up to the Language Selector to re-engage the code blocks. I feel this is probably not the slickest design design, and should be addressed in the future, but adequate for now

@mango-db Yeah, this is sort of a limitation of how our language selector and content work at the moment. It's never a guaranteed that the code block/example for one language will be in the same general area as another (nor does it even have to exist). 

I agree that it's something that can be improved later on as more docs actually use the `tabs-selector` component (there aren't actually that many pages that do)

src/components/code-context.js
rayangler: I went ahead and removed the check for `languageOptions.length` altogether here since it doesn't actually matter if the array itself is empty or not!

PR #508, opened 2021-10-28
DOP-2446: Implement Chapter component

rayangler:
### Stories/Links:

DOP-2446

### Staging Links:

[Guides Staging](https://docs-mongodbcom-staging.corp.mongodb.com/test-guides/guides/raymundrodriguez/DOP-2446/)

### Notes:
* The staging link above uses the default `document` template that regular docs pages use. The `guides-index` template will be added in a future PR.
* The `Chapters` component is pretty much just a container for now. More functionality like chapter vs. gallery view for guides will be implemented along with the `guides-index` template work.
* In the staging link, Chapter 3 intentionally does not have an image to show what it would look like if (for whatever reason) a chapter was included without an image.
* Most of the guides and their information are just placeholders for now. Please only take a look at what the chapter cards currently look like


allisonmui9 (approved):
hey Raymund! this looks great. How detailed should this review be? If super detailed, here are some minor comments. If not, feel free to disregard until later! 

- I noticed there are some slight grey backgrounds. Should I be reviewing the style of the card? 
- Could we remove the underline on link hover? 
- On card that don't have an image, could we wrap the paragraph at the same width as the cards with images? 

Everything else looks great. Thank you!

rayangler:
@allisonmui9 
> I noticed there are some slight grey backgrounds. Should I be reviewing the style of the card?

The gray background is coming from the component that the chapter cards are on. In the designs, this would be equivalent to the gray background of the chapter/gallery view. It doesn't look too good here because the guides' homepage template isn't done yet, but this should be addressed in a future PR!

> Could we remove the underline on link hover?

Done

> On card that don't have an image, could we wrap the paragraph at the same width as the cards with images?

Done. 200px has been reserved for the chapter image on chapters that don't actually have any images. This doesn't look as good on the staging link for smaller screen sizes, but should look better once the homepage template is ready.

Screen Shot 2021-11-01 at 2 16 11 PM

allisonmui9:
cool thank you!!

src/components/Chapter.js
```diff
@@ -0,0 +1,209 @@
...
+  @media ${theme.screenSize.mediumAndUp} {
+    grid-area: description;
+  }
+`;
+
+const ChapterNumberLabel = styled('div')`
+  background-color: ${uiColors.green.light3};
+  border-radius: ${theme.size.tiny};
+  color: ${uiColors.green.base};
+  font-size: 14px;
```
mango-db, line 78: obnoxious theme flag 🏴 🤓

src/components/Chapter.js
```diff
@@ -0,0 +1,209 @@
...
+  font-weight: bold;
+  height: ${theme.size.medium};
+  text-align: center;
+  width: 83px;
+`;
+
+const ChapterTitle = styled('div')`
+  font-size: 18px;
```
mango-db, line 86: obnoxious theme flag 🏴 🤓

src/components/Chapter.js
```diff
@@ -0,0 +1,209 @@
...
+  let guides = [];
+
+  if (guidePaths) {
+    guidePaths.forEach((guidePath) => {
+      const guide = metadata?.guides?.[guidePath];
+      guides.push({
+        path: guidePath,
+        time: guide?.completion_time,
+        title: guide?.title,
+      });
```
mango-db, line 145: Do we need a check for existence of `guide` / `guidePath` / `completion_time` / `title` anywhere before render?

src/components/Chapter.js
```diff
@@ -0,0 +1,209 @@
...
+        <EmptyImage />
+      )}
+      <DescriptionContainer>
+        <ChapterNumberLabel>Chapter {chapterNumber}</ChapterNumberLabel>
+        <ChapterTitle>{chapterTitle}</ChapterTitle>
+        <ChapterDescription>{description}</ChapterDescription>
+      </DescriptionContainer>
+      <GuidesList>
+        {guides.map((guide, i) => {
+          const time = guide.time ? `${guide.time} mins` : '';
```
mango-db, line 178: How do writers know that they shouldn't specify the `min` unit when creating the metadata?

mango-db (approved):
LGTM with miner nits 🪨 🔨 

- Implemented in the standard way
- Works as expected on mobile/desktop
- Has tests

src/components/Chapter.js
rayangler: `guidePath` will always exist as it's an element in the existing `guidePaths` array. 

We do check for the existence of `guide` while attempting to grab the guide's `completion_time` and `title`. If it doesn't exist, `completion_time` and `title` will both be `undefined` due to optional chaining.

If either `completion_time` or `title` don't exist, they'll be added to the object as `undefined`. These values are checked prior to rendering them in the form of `getPlaintext` for the `title`, and checking to see if `time` exists so we could add "mins" to the value.

src/components/Chapter.js
rayangler: The [rstspec](https://github.com/mongodb/snooty-parser/blob/70a6948c21fa19e032fe471e2640debc0868a09d/snooty/rstspec.toml#L801) may clue them in that the `time` directive they need to use accepts a number. Fwiw, writing just a number without the unit is [how it's currently done](https://github.com/10gen/docs-tutorials/blob/6870a958095211f298a454d45027e5e17d98880e/source/realm/react-web-quickstart.txt#L15) in guides!

PR #510, opened 2021-11-09
DOP-2555: The one where we copy media.mongodb.org

casthewiz:
### Stories/Links:

DOP-2555

### Staging Links:

https://docs-mongodbcom-staging.corp.mongodb.com/master/landing/cassidy.schaufele/DOP-2555/
### Notes:

- Moves assets in use by docs to `media`, to support any future deprecation of `media.mongodb.org` usage.
- Switches favicon resource to in-repo favicon, from `media.mongodb.org`.
- All assets pre-emptively populated in production S3 bucket.


rayangler (approved):
LGTM! I just have a couple of questions to help with my understanding/context.

What's the purpose of having the `static/media` folder? References to `media` in this repo still currently point to the media.mongodb.org url. 

Also, all of those same references except for the favicon are associated with css that we (most likely) don't use anymore. Will all of this be expected to go away as we sort through our legacy css?

casthewiz:
So we really just want to support moving off of the media.mongodb.org url where needed - since it's a `?` on who has access and there isn't a true owner for it. But at the same time, there's no real impetus or urgency to move off the url all at once - most of the references are either in legacy css or on older last gen repos. 

Thus most of the references in this repo retained, since it's pretty much all legacy css that we'll be cleaning up. Other references in docs repos can be cleaned up at the doc's team leisure now.

We're tracking the files in repo mostly so that they can be referenced as a source of truth in case something happens to the `docs.mongodb.com` S3 bucket - even if they're already up there, if the bucket was deleted they'd be lost to the ether.

rayangler:
Cool, thanks for the explanation!

PR #511, opened 2021-11-09
DOP-2447: Implement Chapters for guides homepage

rayangler:
### Stories/Links:

DOP-2447
[Designs](https://www.figma.com/file/Fjkr714qJ6a7uyyFQdzko7/Guides?node-id=467%3A4333) (see v4 and v5)
[Forked guides repo/branch](https://github.com/rayangler/docs-tutorials/tree/test-guides)

### Staging Links:

[Guides Staging](https://docs-mongodbcom-staging.corp.mongodb.com/test-guides/guides/raymundrodriguez/DOP-2447/)

To double-check that the changes don't negatively impact other PLPs:
[Realm Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/realm/raymundrodriguez/DOP-2447/)
[Compass Staging](https://docs-mongodborg-staging.corp.mongodb.com/master/compass/raymundrodriguez/DOP-2447/)

### Notes:
* Instead of making a new `guide-index` template, I decided to just reuse the `product-landing` for the guides homepage to prevent having more templates than needed. Alternatively, we could make the `product-landing` css reusable/sharable so that we have a separate `guide-index` template that includes anything specific to guides. Open to any thoughts/preferences on implementation.
* Updated LeafyGreen icons package to accommodate new icons from the design.


src/components/Chapters.js
```diff
@@ -1,19 +1,237 @@
...
+    options: {
+      columns: 3,
+    },
+  };
+};
+
+const Views = {
+  Chapter: {
+    name: 'Chapter',
+    icon: <BookIcon />,
```
rayangler, line 182: I didn't see a matching glyph for this in LG's Icon package so I just made it into a component that we could use.

allisonmui9:
This is amazing raymund. Great job! 
Few comments: 

- I wasn't able to track down a MongoDB in 60 seconds. So we could remove that link! 
- After scrolling past the hero, could we have the chapter and gallery view labels remain fixed? And then the right hand card fixed as well? And then the chapters to scroll underneath that? Let me know if that makes sense! 
- I noticed there's a white space after the last card? Can we extend the gray background to the footer? 
- The icons to the gallery view cards - is it possible for that to be changeable? Had a late request to have those correlate to the related chapters. 

Thank you!

src/components/Chapters/Chapters.js
```diff
@@ -0,0 +1,209 @@
...
+            <ChapterView>
+              <RightColumn />
+              <LeftColumn>
+                {children.map((child, i) => (
+                  <ComponentFactory key={i} metadata={metadata} nodeData={child} />
+                ))}
+              </LeftColumn>
+            </ChapterView>
+          )}
+          {view === Views.Gallery.name && <CardGroup css={guidesCardsStyle} nodeData={galleryViewData} />}
```
casthewiz, line 195: Should we add `galleryViewData` to the boolean check here, on the chance it's `null` from the logic on line 142?

src/components/Chapters/RightColumn.js
```diff
@@ -0,0 +1,139 @@
...
+        observer.observe(sentinel);
+      }
+    }
+
+    return () => {
+      if (sentinel && observer) {
+        observer.unobserve(sentinel);
+      }
+    };
+  }, [topLarge]);
```
casthewiz, line 121: nonblocking, but worth discussing:

This could probably become a more generic hook that lives in `src/hooks` - I can see a future where we might want a utility like this for other features, and it would be easier to test it in isolation then, as well.

src/components/CardGroup.js
```diff
@@ -61,6 +61,7 @@ const StyledGrid = styled('div')`
 `;
 
 const CardGroup = ({
+  className,
```
casthewiz, line 64: We should add this to the propTypes (just to make sure it's clear to expect a `string` and not `string[]`)

casthewiz (approved):
💯 This looks great! Added a few comments, but they're open suggestions and not blocking.

src/components/Chapters/RightColumn.js
rayangler: Good point. I refactored it into something that can hopefully be a little bit more reusable and included a basic test. We can adjust it accordingly in the future. I'll merge this in around tomorrow afternoon ET if everything looks okay.

PR #512, opened 2021-11-19
DOP-2533: Allow frontend to fetch OAS file from Realm

rayangler:
### Stories/Links:

DOP-2533

### Staging Links:

[Atlas Staging - OpenAPI](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/test-openapi/cloud-docs/raymundrodriguez/DOP-2533/reference/api/openapi/index.html)
[Realm Staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/realm/raymundrodriguez/DOP-2533/admin/api/v3/index.html#realm-administrator-apis) - To make sure that the rST/Snooty version of OpenAPI still renders okay

### Notes:
* This PR will provide support for fetching OpenAPI spec (OAS) files from Realm.
* For context: We want users to have a way to see somewhat up-to-date OpenAPI specs rendered on the docs. We have a new Realm trigger and function that takes the most recent version of the OAS file every 24 hours and stores it in our database. Our OpenAPI renderer will then fetch that spec on page load and render it on our frontend. For now, the logic on Realm is tuned for Cloud, but can be adjusted in the future to support other products' spec files.


src/components/OpenAPI.js
```diff
@@ -219,16 +246,38 @@ const MenuTitleContainer = ({ siteTitle, pageTitle }) => {
...
+        }
+      } catch (e) {
+        console.error(e);
+      }
+    };
+
+    if (usesRealm) {
+      fetchData(getPlaintext(argument));
+    }
+  }, [argument, database, usesRealm]);
```
rayangler, line 280: I think in the future, we may want to turn all of the `useEffect`s that use Realm functions to fetch data into their own hooks. I left the code like this for now to maintain the pattern of how we use these Realm functions. We can overhaul/refactor the Realm functions into hooks at a later time.

src/components/OpenAPI.js
```diff
@@ -241,49 +290,52 @@ const OpenAPI = ({ metadata, nodeData: { argument, children, options = {} }, pag
...
     <>
-      <Global styles={globalCSS} />
+      <Global styles={getGlobalCss(topValues)} />
       {/* Temporary loading widget to be removed once the Redoc component loads */}
-      <LoadingWidget className={tempLoadingDivClassName} />
       <RedocStandalone
         onLoaded={() => {
-          if (load) {
-            return;
-          }
```
mango-db, line 263: What's the reason for this deletion?

src/components/Sidenav/SidenavBackButton.js
```diff
@@ -70,10 +84,11 @@ const SidenavBackButton = ({ border, currentSlug, handleClick, project, target,
         as={Link}
         className={cx(sideNavItemBasePadding, backButtonStyling)}
         to={url}
-        glyph={<Icon glyph="ArrowLeft" size="small" />}
+        glyph={glyph}
         onClick={handleClick}
         {...props}
       >
+        {!enableGlyph && <span className={cx(htmlBackIcon)}>&#8592;</span>}
```
mango-db, line 91: Might be worth leaving a quick comment about the glyph workaround here as well

mango-db (commented):
Does this have any chance of interfering with our CDN caching, if there is a 24-hr refresh cycle? Or would it only update the caches if there is a spec state change? 🤔 Hopefully the latter...

staged master-branch [cloud-docs](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/cgoecknerwald/master/index.html)

src/components/OpenAPI.js
```diff
@@ -241,49 +290,52 @@ const OpenAPI = ({ metadata, nodeData: { argument, children, options = {} }, pag
     );
   }
 
-  // Check for JSON string spec first
-  const spec = children[0]?.value;
-  const specOrUrl = spec ? JSON.parse(spec) : argument[0]?.refuri;
```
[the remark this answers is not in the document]
rayangler: I have a [PR on the parser](https://github.com/mongodb/snooty-parser/pull/360) that updated the implementation of passing a url as the argument to the `openapi` directive. 

This change is mostly due to that fact that certain urls may not have the proper CORS configuration to allow fetching its content directly on the frontend. Redoc wouldn't be able to parse the spec file contents this way. The workaround is to have the parser pull the file contents instead and pass it as a JSON string.

src/components/OpenAPI.js
rayangler: The tl;dr is that I realized it wasn't really doing anything and keeping it around was resulting in an additional and unnecessary re-render.

mango-db (approved):
looks green to monkeys

PR #516, opened 2021-12-02
(DOP-2393) Fix permalinks to copy url and add tooltip when copied

gjchong25:
### Stories/Links:

[DOP-2393](https://jira.mongodb.org/browse/DOP-2393)

### Staging Links:

[Staging Link for Compass ](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/compass/grace.chong/DOP-2393/index.html)

### Notes:
This change automatically copies the permalink URL when a permalink link icon is clicked. It also shows a Tooltip that indicates that text has been copied to the clipboard.


src/components/Heading.js
```diff
@@ -23,6 +36,33 @@ const Heading = ({ sectionDepth, nodeData, page, ...rest }) => {
   const hasSelectors = selectors && Object.keys(selectors).length > 0;
   const shouldShowMobileHeader = isPageTitle && isTabletOrMobile && (hasSelectors || !hidefeedbackheader);
 
+  const [copied, setCopied] = useState(false);
+  const [headingNode, setHeadingNode] = useState(null);
+  const url = isBrowser ? window.location.href.split('#')[0] : null;
```
rayangler, line 41: Instead of making the alternative value `null`, should we turn this into an empty string? I'm afraid of `null` turning into the string `"null"` when it's concatenated with a string, as in:
```
null + "#" === "null#"
```

rayangler (commented):
Looks good so far and it seems to work fine on the staging link! Just a nitpick and a couple of things to consider

allisonmui9 (changes requested):
hello! looking good so far! I think the visual and style looks great! Only comment is that when the permalink is clicked, it jumps quite far into that section. Could we still jump down to that section but keep the header still in view? 

Let me know if that needs clarifying! Thanks :)

src/components/Heading.js
```diff
@@ -23,6 +33,16 @@ const Heading = ({ sectionDepth, nodeData, page, ...rest }) => {
   const hasSelectors = selectors && Object.keys(selectors).length > 0;
   const shouldShowMobileHeader = isPageTitle && isTabletOrMobile && (hasSelectors || !hidefeedbackheader);
 
+  const [copied, setCopied] = useState(false);
+  const [headingNode, setHeadingNode] = useState(null);
+  const url = isBrowser ? window.location.href.split('#')[0] : '';
+
+  useCopyClipboard(copied, setCopied, headingNode, url + '#' + id);
```
casthewiz, line 40: nitpick: since we don't use `url` in the file outside of these two lines, can we take the `url + '#' + id` statement and put it into the assignment on line 38? We could then pass just `url` as the last argument in `useCopyClipboard`.

src/hooks/useCopyClipboard.js
```diff
@@ -0,0 +1,26 @@
...
+      const timeoutId = setTimeout(() => {
+        setCopied(false);
+      }, 1500);
+
+      return () => clearTimeout(timeoutId);
+    }
+
+    return () => clipboard.destroy();
+  }, [component, contents, copied, setCopied]);
+};
```
casthewiz, line 24: It might be useful to indicate (either through variable name or comment) that `component` is expected to be a `ref` - at least, that's my understanding. If I'm off-base, feel free to ignore 😆

casthewiz (approved):
I think this looks structurally fine, I just have two minor nitpickish comments 😄

rayangler (approved):
LGTM!

allisonmui9 (approved):
LGTM thank you Grace!

PR #525, opened 2022-01-05
DOP-2504: YAML steps can be rendered as Procedure and Step directives

rayangler:
### Stories/Links:

DOP-2504

### Staging Links:

[Guides Staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/test-guides-steps/guides/raymundrodriguez/DOP-2504/cloud/account/index.html)
[Compass Staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/test-steps/compass/raymundrodriguez/DOP-2504/index.html)

### Notes:
* As part of guides revamp work, it was asked that the docs team have a way to create procedural steps without needing to write them out in a separate YAML file.
* The existing `Procedure` and `Step` components have been refactored to allow for PLP-based styling or normal/YAML steps styling. Currently, existing uses of the `procedure` and `step` directives that don't have a `style` option will default to the "connected" style. This will allow PLPs that use the directives to maintain styling without immediately breaking the design of the page. We can announce that uses of the `procedure` directive on PLPs should include the "connected" style option and update the PLP reference wiki accordingly.
* `StepYAML` and `StepsYAML` have been removed. Through the parser ([PR](https://github.com/mongodb/snooty-parser/pull/372)), steps written in YAML will be converted to `procedure` and `step` directives with "normal" styling.


src/components/Heading.js
```diff
@@ -105,6 +105,12 @@ const ChildContainer = styled.div(
   `
 );
 
+const LinkIcon = styled.img`
+  border-radius: 0 !important;
+  display: initial !important;
+  margin: initial !important;
```
rayangler, line 111: Added this since headings in PLP steps would have their link icons wrap to the next line when used in PLP-styled tabs

tests/unit/Procedure.test.js
```diff
@@ -8,4 +8,15 @@ import mockData from './data/Procedure.test.json';
 it('renders correctly', () => {
   const tree = shallow(<Procedure nodeData={mockData} />);
   expect(tree).toMatchSnapshot();
+  expect(tree.find('StyledProcedure').props()).toHaveProperty('procedureStyle', 'connected');
```
casthewiz, line 11: Instead of testing to see if the prop drills through, can we use `.toHaveStyleRule()` to make sure that the intended style is applied?

Alternatively, if we move the default logic to be the `Step`'s responsibility, we don't have to check for the style logic here, which seems cleaner and easier. We could just add a default case test to the Step test suite in that case.

src/components/Step.js
```diff
@@ -1,62 +1,105 @@
...
-const Step = ({ nodeData: { children, argument }, stepNumber, ...rest }) => {
+const Step = ({ nodeData: { children }, isLastStep, stepNumber, stepStyle, ...rest }) => {
   return (
-    <LandingStep>
-      <Circle>
-        <StepNumber>{stepNumber}</StepNumber>
-      </Circle>
-      <h2>
-        {argument.map((child, i) => (
+    <StyledStep css={landingStepStyles[stepStyle]}>
```
casthewiz, line 93: Maybe redundant, but I'd consider adding a default value for `stepStyle` here as well - or moving the default value logic to the Step and out of the `Procedure`. 

It seems a little off to have the default value housed in another component unless it's going to be specific to only when the components render in tandem, and I think we want to make sure that the `connected` styles render by default regardless?

casthewiz (commented):
Just a small comment on where default values to preserve backwards compatibility should be housed. I might not have full context on current state vs. intended state though.

Looks good! I love that we won't have to maintain `StepYAML` and  `StepsYAML`.

casthewiz (approved):
Looks good!

PR #527, opened 2022-01-19
(DOP-2651) Frontend components for input, output, and io code block, i/o toggle functionality, updated caption style

gjchong25:
### Stories/Links:

[DOP-2651](https://jira.mongodb.org/browse/DOP-2651)

### Staging Links:

[Atlas Staging Link](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/grace.chong/DOP-2651/driver-connection/index.html#driver-examples)

### Notes:

This tech spec related to this ticket can be found [here](https://docs.google.com/document/d/1mwoAsphbdLDbLJKo6ZCb2NfBGhpTzlT5oX_YEJlZ_sA/edit). This PR handles the frontend components of the new `io-code-block` directive and its corresponding required child directives, `input` and `output`. All of the code related to code snippets is now found in the `Code` directory. It also adds the new `caption` option UI, which now displays on top of the code snippet, instead of transforming the code snippet into a "chrome" view which limits functionality. 

 In the staging link, the following examples are shown: (1) New `io-code-block` with `input/output` taking in raw code content, (2) `io-code-block `with `input/output` reading from a specified file path, (3) a normal `code-block` with the new caption style, and (4) a normal `code-block` without the caption style, to show that previous UI and functionality of the code block is maintained.


rayangler (commented):
Nice work! Thank you for the examples on the page, super helpful! Just a couple of nitpicks and questions for now

src/components/Code/Output.js
```diff
@@ -0,0 +1,35 @@
...
+`;
+
+const Output = ({ nodeData: { children }, ...rest }) => {
+  const { emphasize_lines, value, linenos } = children[0];
+
+  return (
+    <CodeBlock
+      className={cx(outputCodeStyle)}
+      highlightLines={emphasize_lines}
+      language={'none'}
```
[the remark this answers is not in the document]
gjchong25: I would expect it to be, since output should be plain text? Language does get passed in from the parent node to output's Code child so if we wanted to display language here it shouldn't be hard.

src/components/Code/CodeIO.js
```diff
@@ -0,0 +1,102 @@
...
+  const [arrow, setArrow] = useState('ChevronDown');
+  let outputExists = showOutput === 'none' ? '4px' : '0px';
+
+  const handleClick = (e) => {
+    if (showOutput === 'inline') {
+      setShowOutput('none');
+      setButtonText('VIEW OUTPUT');
+      setArrow('ChevronDown');
+    } else {
+      setShowOutput('inline');
```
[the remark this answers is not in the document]
gjchong25: Yep, I thought I used this somewhere directly in the css so I didn't want to have to do the extra step of checking the boolean first before setting the display state, but it turns out I don't !

src/components/Code/CodeIO.js
```diff
@@ -0,0 +1,102 @@
...
+  };
+
+  return (
+    <div
+      css={css`
+        display: table;
+        margin: ${theme.size.default} 0;
+        min-width: 150px;
+        table-layout: fixed;
+        width: 100%;
```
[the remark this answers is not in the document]
gjchong25: Yep, this makes sense for the shared css!

src/components/Code/Output.js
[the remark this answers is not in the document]
rayangler: Makes sense to leave it as is then. Thanks!

src/components/Code/Code.js
```diff
@@ -1,11 +1,20 @@
...
-import { theme } from '../theme/docsTheme';
-import { reportAnalytics } from '../utils/report-analytics';
+import { CodeContext } from '../code-context';
+import { TabContext } from '../tab-context';
+import { reportAnalytics } from '../../utils/report-analytics';
+import { baseCodeStyle, borderCodeStyle } from './styles/codeStyle';
+
+const captionStyle = css`
+  padding: 10px;
+  color: #5d6c74;
```
rayangler, line 13: (Nit) This can be `uiColors.gray.dark1`

rayangler (commented):
Just one more nitpick but otherwise LGTM. It might be worth adding tests for the new components you have as well. It might be fine to do it after the React Testing Library PR has been merged so that you don't have to rewrite any of the tests

src/components/Code/CodeIO.js
```diff
@@ -0,0 +1,92 @@
...
+              onClick={handleClick}
+              leftGlyph={<Icon glyph={arrow} fill="#FF0000" />}
+            >
+              {buttonText}
+            </Button>
+          </IOToggle>
+        </>
+      )}
+      {showOutput && <ComponentFactory {...rest} nodeData={children[1]} />}
+      {children.length !== 2 && children.map((child, i) => <ComponentFactory {...rest} key={i} nodeData={child} />)}
```
casthewiz, line 76: I would assign `children.length === 2` to a `const` with a contextual name (maybe named `needsToggle`?) to add a little more context to the ternaries here.

tests/unit/CodeIO.test.js
```diff
@@ -0,0 +1,23 @@
...
+    const wrapper = render(<CodeIO nodeData={mockData} />);
+    expect(wrapper.asFragment()).toMatchSnapshot();
+  });
+
+  it('opens and closes output code snippet when io toggle button is clicked', () => {
+    const wrapper = render(<CodeIO nodeData={mockData} />);
+    userEvent.click(wrapper.getByRole('button'));
+    expect(wrapper.getByRole('outputShown')).toBeTruthy();
+    userEvent.click(wrapper.getByRole('button'));
+    const x = wrapper.queryAllByRole('outputShown');
```
rayangler, line 20: If `outputShown` is only going to be used for testing, we can either (1) check to see if the output is displayed by checking if the output text exists, or (2) use `data-testid` instead of `role`

src/components/Code/Output.js
```diff
@@ -0,0 +1,36 @@
...
+  border-bottom-right-radius: 4px;
+  border-bottom-left-radius: 4px;
+`;
+
+const Output = ({ nodeData: { children }, ...rest }) => {
+  const { emphasize_lines, value, linenos } = children[0];
+
+  return (
+    <CodeBlock
+      role="outputShown"
```
rayangler, line 16: `role` is typically used for accessibility and semantics but I'm not sure if it's necessary here if only using for testing

tests/unit/CodeIO.test.js
```diff
@@ -0,0 +1,22 @@
...
+    const wrapper = render(<CodeIO nodeData={mockData} />);
+    expect(wrapper.asFragment()).toMatchSnapshot();
+  });
+
+  it('opens and closes output code snippet when io toggle button is clicked', () => {
+    const wrapper = render(<CodeIO nodeData={mockData} />);
+    userEvent.click(wrapper.getByRole('button'));
+    expect(wrapper.getByText('hello world')).toBeTruthy();
+    userEvent.click(wrapper.getByRole('button'));
+    expect(wrapper.queryAllByText('hello world').length === 0);
```
rayangler, line 20: (Nit) This can be `expect(wrapper.queryAllByText('hello world')).toHaveLength(0)` but mostly a nitpick and I'm okay with leaving as is

rayangler (approved):
LGTM, just one optional nitpick. Feel free to merge after the parser PR is good to go

casthewiz (approved):
LGTM!

src/components/ComponentFactory.js
```diff
@@ -31,6 +32,7 @@ import Heading from './Heading';
 import HorizontalList from './HorizontalList';
 import Image from './Image';
 import Include from './Include';
+import Input from './Code/Input';
```
casthewiz, line 35: We might want to alias this as `import CodeInput from './Code/Input'` - since that file is using a default export, we don't have to use `as`.

This might help give us some context further down the page when we assign `Input` to `input` in the mapping - I can see a future where that context isn't obvious when tracing through.

(we can also apply this to `output` if agreeable 😄 )

src/components/Code/CodeIO.js
```diff
@@ -0,0 +1,95 @@
...
+              disabled={false}
+              onClick={handleClick}
+              leftGlyph={<Icon glyph={arrow} fill="#FF0000" />}
+            >
+              {buttonText}
+            </Button>
+          </IOToggle>
+        </>
+      )}
+      {showOutput && <ComponentFactory {...rest} nodeData={children[1]} />}
```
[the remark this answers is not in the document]
casthewiz: Also, is there ever the expectation that this won't be an `Output` component?

If we always expect an `Output` component, we can directly use it here instead of passing args through the ComponentFactory.

src/components/Code/CodeIO.js
```diff
@@ -0,0 +1,95 @@
...
+          border-top-left-radius: 0px;
+          border-bottom-right-radius: ${outputExists};
+          border-bottom-left-radius: ${outputExists};
+          margin: 0px;
+        }
+      `}
+    >
+      {needsIOToggle && (
+        <>
+          <ComponentFactory {...rest} nodeData={children[0]} />
```
casthewiz, line 63: Will this always be an `Input` component? If it is, we can directly create the `Input` here instead of passing through the `ComponentFactory`.

src/components/ComponentFactory.js
casthewiz: Also, do we ever expect to use the `Input` and `Output` outside of the `CodeIO` component? If we don't, we could directly invoke them in the `CodeIO` component and omit them from the component factory.

casthewiz (commented):
Functionality wise, this looks 💯  - just some quick questions about `Input` and `Output`, and whether we want to limit their scope to `CodeIO`.

src/components/Code/Code.js
```diff
@@ -25,6 +35,8 @@ const Code = ({ nodeData: { caption, copyable, emphasize_lines: emphasizeLines,
   const { languageOptions, codeBlockLanguage } = useContext(CodeContext);
   const code = value;
   const language = (languageOptions?.length > 0 && codeBlockLanguage) || getLanguage(lang);
+  const captionSpecified = !!caption;
+  const captionExists = captionSpecified ? '0px' : '4px';
```
rayangler, line 39: (Nit) This variable name kind of sounds like a boolean. Should we make it more specific to what its value represents? Something like `captionBorderRadius`?

src/components/Code/CodeIO.js
```diff
@@ -0,0 +1,95 @@
...
+  height: 24px;
+  margin: 8px;
+  font-weight: bold;
+`;
+
+const CodeIO = ({ nodeData: { children }, ...rest }) => {
+  const [showOutput, setShowOutput] = useState(false);
+  const [buttonText, setButtonText] = useState('VIEW OUTPUT');
+  const [arrow, setArrow] = useState('ChevronDown');
+  const outputExists = !showOutput ? '4px' : '0px';
```
rayangler, line 24: (Nit) `outputExists` also sounds like a boolean, similar to `captionExists`. Would it be possible to change this as well?

rayangler (commented):
Just a couple of nitpicks!

src/components/Code/CodeIO.js
```diff
@@ -0,0 +1,95 @@
...
+              onClick={handleClick}
+              leftGlyph={<Icon glyph={arrow} fill="#FF0000" />}
+            >
+              {buttonText}
+            </Button>
+          </IOToggle>
+        </>
+      )}
+      {showOutput && <ComponentFactory {...rest} nodeData={children[1]} />}
+      {children.length !== 2 && children.map((child, i) => <ComponentFactory {...rest} key={i} nodeData={child} />)}
```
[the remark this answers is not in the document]
gjchong25: Nice catch, sorry for overlooking this!

src/components/Code/CodeIO.js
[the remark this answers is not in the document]
gjchong25: No, it should always be an Output component.

src/components/ComponentFactory.js
gjchong25: You're right! Removed them from ComponentFactory. I think I included it as a safety measure just in case some other directives were passed into `io-code-block`, but that was before I added all the validity checks in the parser, making sure that only input/output can be children.

casthewiz (approved):

src/components/Code/CodeIO.js
```diff
@@ -37,47 +38,52 @@ const CodeIO = ({ nodeData: { children }, ...rest }) => {
   };
 
   return (
-    <div
-      css={css`
-        ${baseCodeStyle}
+    <>
+      {children.length === 0 && <></>}
```
rayangler, line 42: We may be able to move this logic into an if statement above the return like so:

```
if (children.length === 0) {
  return null;
}

return ( ... );
```

Not necessarily a blocker, but it would remove the need for wrapping the code below with the check for `children.length > 0`, and we can remove the need for wrapping everything within a fragment

rayangler (approved):

casthewiz (approved):

PR #529, opened 2022-01-21
DOP-2503: Expose ContentsList component for guides usage

rayangler:
### Stories/Links:

DOP-2503

### Staging Links:

[Guides Staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/test-guides/guides/raymundrodriguez/DOP-2503/index.html)
[MongoDB for VS Code Staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/mongodb-vscode/raymundrodriguez/DOP-2503/databases-collections/index.html) (to see the "On This Page"/Contents component in the docs)

### Notes:
* This PR exposes the internal components that `Contents` was using for the "On This Page" component. The styling and behavior should remain unchanged, especially for the docs pages that currently uses this component.


src/hooks/useActiveHeading.js
```diff
@@ -26,21 +26,11 @@ const useActiveHeading = (headingNodes) => {
...
-    // Callback is first performed upon page load. Ignore checking entries on first load to allow
-    // headings[0] to be counted first when headings[0] and headings[1] are both intersecting
-    let firstLoad = true;
-    let defaultActiveHeadingId = window.location.hash.slice(1) || headingNodes?.[0]?.id;
     const callback = (entries) => {
-      if (firstLoad) {
-        firstLoad = false;
-        setActiveHeadingId(defaultActiveHeadingId);
-        return;
-      }
```
rayangler, line 38: Deleted this because:

(1) Keeping this in was causing the active id to never change when the hook was used in a single component instead of within a context.

(2) I haven't been able to reproduce the behavior described by the comment. The behavior may be obsolete now but happy to add this back in if needed.

allisonmui9 (approved):
lgtm. Thanks Raymund!!! 🔥

src/components/Chapters/Chapter.js
```diff
@@ -139,11 +140,12 @@ const Chapter = ({ metadata, nodeData: { argument, options } }) => {
   const description = options?.description;
   const image = options?.image;
   const chapterTitle = getPlaintext(argument);
-  const chapterNumber = metadata?.chapters?.[chapterTitle]?.chapter_number;
-  const guides = useMemo(() => getGuidesData(metadata, chapterTitle), [metadata, chapterTitle]);
+  const currentChapter = metadata?.chapters?.[chapterTitle];
+  const chapterNumber = currentChapter.chapter_number;
```
gjchong25, line 144: Do we not want to define chapterNumber with `currentChapter?.chapter_number`? Since isn't it still possible that `currentChapter` is undefined?

tests/unit/Chapters.test.js
```diff
@@ -23,11 +24,19 @@ const mountChapters = () => {
 describe('Chapters', () => {
   jest.useFakeTimers();
 
+  beforeAll(() => {
+    jest.spyOn(useActiveHeading, 'default').mockImplementation(() => 'atlas');
```
gjchong25, line 28: Just curious, what is the reasoning behind this line and how is it being used for testing? And what is the function `() => 'atlas'` accomplishing?

src/hooks/useActiveHeading.js
```diff
@@ -26,21 +26,11 @@ const useActiveHeading = (headingNodes) => {
...
-
       for (let entry of entries) {
```
gjchong25, line 30: Since we're not changing the entry, could it be `const entry` instead of `let`?

gjchong25 (changes requested):
Great job, just a few questions and suggestions that you can feel free to use! It looks really nice UI-wise!

tests/unit/Chapters.test.js
```diff
@@ -7,6 +7,7 @@ import Chapters from '../../src/components/Chapters/Chapters';
 import { tick } from '../utils';
 import { SidenavContext } from '../../src/components/Sidenav';
 import { theme } from '../../src/theme/docsTheme';
+import * as useActiveHeading from '../../src/hooks/useActiveHeading';
```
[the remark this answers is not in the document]
rayangler: While attempting to mock the hook, the tests were still trying to use the function's original implementation. This way of importing the hook allows us to import it as an object, and then we spy on the object's default function to mock the hook. We perform similar patterns in some of our other tests to mock things

tests/unit/Chapters.test.js
rayangler: The `useActiveHeading` hook uses the `IntersectionObserver` API, which is implemented for a browser. Since we don't have access to a browser that implements this API while testing, we mock the hook here to bypass making calls to the non-existent `IntersectionObserver`. This will prevent errors from the hook being called by the component we're testing.

In this case, the mock function will just return the string "atlas" immediately because the original hook is supposed to return the id of the active heading/node being observed by the `IntersectionObserver`. Realistically, the mock function here could return anything since we're only using it here to prevent errors from occurring.

tests/unit/Chapters.test.js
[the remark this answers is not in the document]
gjchong25: Oh gotcha, thanks for clarifying!

tests/unit/Chapters.test.js
gjchong25: Makes sense!

gjchong25 (approved):
LGTM once Cassidy takes a second look!

src/components/Chapters/RightColumn.js
```diff
@@ -71,30 +78,59 @@ const StyledLink = styled(Link)`
...
   margin-bottom: ${theme.size.small};
 `;
 
-const RightColumn = () => {
+const RightColumn = ({ chapters }) => {
   const { isCollapsed } = useContext(SidenavContext);
   // Have children of the RightColumn appear as user scrolls past hero image on large screen sizes
   const isVisible = useVisibleOnScroll('.hero-img');
+  const chapterEntries = Object.entries(chapters);
+  const chapterValues = Object.values(chapters);
```
casthewiz, line 100: I would default `chapters` to `{ }` here - if `chapters` is nullish, then this will throw `Uncaught TypeError: Cannot convert undefined or null to object`

casthewiz (approved):
lgtm! 🚀

PR #532, opened 2022-01-27
DOP-2605: Fix styling for permalink and apply it to Target

gjchong25:
This should have been dealt with previously for [DOP-2393](https://jira.mongodb.org/browse/DOP-2393) but it seems I missed a spot! Included a cleaner style for the Tooltip (as per Allison's original designs) and consolidated the permalink code in one file. I also changed the link file to an svg in hopes that the [strange Google search issues](https://jira.mongodb.org/browse/DOP-2640?focusedCommentId=4286771&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4286771) will go away. 

### Stories/Links:

[DOP-2605](https://jira.mongodb.org/browse/DOP-2605)

### Staging Links:

[Docs Staging Link](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/grace.chong/DOP-2605/reference/command/serverStatus/index.html#syntax)


allisonmui9 (commented):
Styling looks good! 
Note: When you click on the link icon, the icon disappears before the tooltip. Could we keep the link showing for a few more seconds or align it with the tooltip disappearing?

Thank you!

src/components/Target.js
```diff
@@ -13,18 +17,32 @@ const partition = (array, isValid) => {
...
-    <dt id={html_id}>
+    <dt>
       {children.map((child, j) => (
         <ComponentFactory key={j} {...rest} nodeData={child} />
       ))}
-      <a href={`#${html_id}`} className="headerlink" title="Permalink to this definition">
-        ¶
-      </a>
+      <Permalink copied={copied} setHeadingNode={setHeadingNode} id={html_id} handleClick={handleClick} />
+      <HeaderBuffer id={html_id}></HeaderBuffer>
```
rayangler, line 36: Should this be a self-closing tag since there's no content nested inside?

src/components/Target.js
```diff
@@ -13,18 +17,32 @@ const partition = (array, isValid) => {
...
 const DescriptionTerm = ({ children, html_id, ...rest }) => {
+  const [copied, setCopied] = useState(false);
+  const [headingNode, setHeadingNode] = useState(null);
+  const url = isBrowser ? window.location.href.split('#')[0] + '#' + html_id : '';
+
+  useCopyClipboard(copied, setCopied, headingNode, url);
+
+  const handleClick = (e) => {
+    setCopied(true);
+  };
```
rayangler, line 28: This logic is more or less the same as what's on the `Heading` component. Should we move this logic into the new `Permalink` component instead, for both `Target` and `Heading`? Or do we want this to be separate?

src/components/Permalink.js
```diff
@@ -0,0 +1,54 @@
...
+  <>
+    <a
+      className="headerlink"
+      ref={setHeadingNode}
+      css={headingStyle}
+      href={`#${id}`}
+      title="Permalink to this definition"
+      onClick={handleClick}
+    >
+      <LinkIcon src={withPrefix('assets/link.svg')} />
```
rayangler, line 39: We can probably remove/delete `assets/link.png` if we're sticking with `assets/link.svg`

src/components/Permalink.js
```diff
@@ -0,0 +1,54 @@
...
+`;
+
+const LinkIcon = styled.img`
+  border-radius: 0 !important;
+  display: initial !important;
+  margin: initial !important;
+`;
+
+const Permalink = ({ copied, setHeadingNode, id, handleClick }) => (
+  <>
```
rayangler, line 30: We can remove this fragment since we can use the `a` tag as our parent for this component

src/components/Target.js
gjchong25: I kept this separate because I thought that the states for the two components need to be defined within each separate component. But I guess if they're consolidated in Permalink, and then Permalink is used in Heading, then the state would be specific to the given component its found in and not `Target`? I can try this out!

src/components/Permalink.js
```diff
@@ -0,0 +1,83 @@
...
+  ${!!copied && 'visibility: visible !important;'}
+  align-self: center;
+  padding: 0 10px;
+`;
+
+const Permalink = ({ id, description, buffer }) => {
+  const [copied, setCopied] = useState(false);
+  const [headingNode, setHeadingNode] = useState(null);
+  const url = isBrowser ? window.location.href.split('#')[0] + '#' + id : '';
+  const bufferSpace = buffer ? buffer : '-175px';
```
rayangler, line 37: We may be able to simplify this logic like so:
```
const bufferSpace = buffer || '-175px';
```

rayangler (commented):
One last suggestion!

rayangler (approved):
LGTM!

PR #533, opened 2022-01-28
DOP-2650: The great url accommodation roundup

casthewiz:
### Stories/Links:

DOP-2650

### Staging Links:

[Landing](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/cassidy.schaufele/DOP-2650/index.html)

[Realm](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/realm/cassidy.schaufele/DOP-2650/index.html)

### Notes:

This changeset adds a series of helpers for determining the `base url` that a front end website is deployed with.
It also utilizes these helpers throughout the frontend to ensure compatibility with tandem `dotcom` url scheme deployments, e.g. properties following `www.mongodb.com/docs` url patterns instead of `docs.*.mongodb.com` patterns.

This is implemented by first preferring the presence of a `GATSBY_BASE_URL` environment variable, then otherwise falling back on logic to determine if the site uses `www` or `docs` subdomain patterns.

#### How do we test this?
image

Well, at least we don't manually test this with the `www` logic, on account of that logic being intrinsically coupled to being hosted on a `www` site. The conceit of this change is that it's backwards and forwards compatible, and works with both current and dotcom instanced sites. Smoke testing in staging links should give us enough confidence that everything 'works' as intended, combined with unit tests present.

### DIFF
--- a/src/components/Breadcrumbs.js
+++ b/src/components/Breadcrumbs.js
@@ -5,6 +5,7 @@ import { uiColors } from '@leafygreen-ui/palette';
 import BreadcrumbSchema from './BreadcrumbSchema';
 import { theme } from '../theme/docsTheme';
 import BreadcrumbContainer from './BreadcrumbContainer';
+import { baseUrl } from '../utils/dotcom';
 
 const Wrapper = styled('nav')`
   font-size: ${theme.fontSize.small};
@@ -23,7 +24,7 @@ const Wrapper = styled('nav')`
 const Breadcrumbs = ({ homeUrl = null, pageTitle = null, parentPaths, siteTitle, slug }) => {
   const homeCrumb = {
     title: 'Docs Home',
-    url: homeUrl || 'https://docs.mongodb.com/',
+    url: homeUrl || baseUrl(true),
   };
   // If a pageTitle prop is passed, use that as the last breadcrumb instead
   const lastCrumb = {
--- a/src/components/Breadcrumbs.js
+++ b/src/components/Breadcrumbs.js
@@ -6,6 +6,7 @@ import BreadcrumbSchema from './BreadcrumbSchema';
 import { theme } from '../theme/docsTheme';
 import BreadcrumbContainer from './BreadcrumbContainer';
 import { baseUrl } from '../utils/dotcom';
+import { assertTrailingSlash } from '../utils/assert-trailing-slash';
 
 const Wrapper = styled('nav')`
   font-size: ${theme.fontSize.small};
@@ -24,7 +25,7 @@ const Wrapper = styled('nav')`
 const Breadcrumbs = ({ homeUrl = null, pageTitle = null, parentPaths, siteTitle, slug }) => {
   const homeCrumb = {
     title: 'Docs Home',
-    url: homeUrl || baseUrl(true),
+    url: homeUrl || assertTrailingSlash(baseUrl(true)),
   };
   // If a pageTitle prop is passed, use that as the last breadcrumb instead
   const lastCrumb = {
--- a/src/components/SEO.js
+++ b/src/components/SEO.js
@@ -1,8 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Helmet } from 'react-helmet';
+import { baseUrl } from '../utils/dotcom';
 
 const DEFAULT_TWITTER_SITE = '@mongodb';
+const metaUrl = `${baseUrl(true)}/assets/meta_generic.png`;
 
 const SEO = ({ pageTitle, siteTitle }) => (
   <Helmet>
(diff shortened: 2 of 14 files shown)


src/components/Breadcrumbs.js
```diff
@@ -23,7 +24,7 @@ const Wrapper = styled('nav')`
 const Breadcrumbs = ({ homeUrl = null, pageTitle = null, parentPaths, siteTitle, slug }) => {
   const homeCrumb = {
     title: 'Docs Home',
-    url: homeUrl || 'https://docs.mongodb.com/',
+    url: homeUrl || baseUrl(true),
```
rayangler, line 27: Should we assert trailing slash here as well or is it not necessary?

src/html.js
```diff
@@ -1,5 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { baseUrl } from './utils/dotcom';
+
+const metaUrl = `${baseUrl(true)}/assets/meta_generic.png`;
+const metaSecureUrl = `http://${baseUrl()}/assets/meta_generic.png`;
```
rayangler, line 6: Does this need to maintain `http` over `https`? Or are we keeping it `http` to prevent any unintended changes?

src/templates/landing.js
```diff
@@ -58,7 +59,7 @@ const Landing = ({ children }) => {
             '@context': 'http://schema.org',
             '@type': 'WebSite',
             name: 'MongoDB Documentation',
-            url: 'https://docs.mongodb.com/',
+            url: baseUrl(true),
```
rayangler, line 62: Also wondering if leaving out trailing slash here is okay

src/utils/dotcom.js
```diff
@@ -0,0 +1,35 @@
+import { isBrowser } from './is-browser';
+
+// Used to convert a 'docs.mongodb.com' or 'docs.<product>.mongodb.com' url to the new dotcom format
+// this *should* be removed post consolidation, or alternatively, made to use `new URL(url)` and polyfilled for safari
+// with relevant string split-slice-join logic turned into URL.pathname, and etc.
```
rayangler, line 5: What's the benefit of keeping this function post-consolidation? Is it in case of any instances of urls that we missed?

src/components/SEO.js
```diff
@@ -14,11 +16,8 @@ const SEO = ({ pageTitle, siteTitle }) => (
     <meta name="twitter:card" content="summary_large_image" />
     <meta name="twitter:site" content={DEFAULT_TWITTER_SITE} />
     <meta property="twitter:title" content={pageTitle} />
-    <meta name="twitter:image" content="http://docs.mongodb.com/assets/meta_generic.png" />
-    <meta
-      name="twitter:image:alt"
-      content="MongoDB logo featuring a green leaf on a dark gray background. Slogan reads 'For giant ideas'."
-    />
+    <meta name="twitter:image" content={metaUrl} />
+    <meta name="twitter:image:alt" content="MongoDB logo featuring a green leaf on a dark green background." />
```
casthewiz, line 20: Took some liberty here, but the old description doesn't match the new logo.

tests/unit/__snapshots__/Breadcrumbs.test.js.snap
```diff
@@ -110,7 +110,7 @@ exports[`renders correctly with siteTitle 1`] = `
     <p>
       <a
         class="emotion-0 emotion-1"
-        href="https://docs.mongodb.com/"
```
casthewiz, line 113: `assertTrailingSlash(baseUrl(true))` implies to me that this shouldn't have changed.... and yet it did. 🤔

rayangler (commented):
Everything looks good! Just some questions for either discussion/my understanding

src/html.js
casthewiz: Oh my goodness the names are swapped

...correcting that 🤦‍♂️

src/components/Breadcrumbs.js
casthewiz: Annnnd that's why the snapshot changed. Great catch 🙏

src/templates/landing.js
casthewiz: Adding an assert here for parity purposes

src/utils/dotcom.js
casthewiz: The benefit's really in having a dynamic base url function - since it would let us differentiate between staging/pre-prod/production base urls, and enable us to have true click through when testing.

Which doesn't really matter for us all that much, honestly, but would be pretty useful for non-engineers smoke testing (product, SEO, etc.)

rayangler: That sounds like a good idea! Thanks for the explanation

rayangler (approved):
LGTM! I really like that we can dotcomify urls based on the env variable or the hostname

PR #538, opened 2022-02-01
Dop 2511

crisatmongo:
### Stories/Links:

DOP-2511

### Staging Links:
https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/compass/cris.newsome/DOP-2511/index.html


src/components/Widgets/FeedbackWidget/FeedbackHeading.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import StarRating, { StarRatingLabel } from './components/StarRating';
 import { useFeedbackState } from './context';
 
-export default function FeedbackHeading({ isVisible = true }) {
+export default function FeedbackHeading({ isVisible = false }) {
```
rayangler, line 5: This change successfully handles removing the feedback heading on mobile, but also removes the feedback heading on medium/tablet sized screens. Do we want to intentionally remove this from both screen sizes?

rayangler (commented):
Looks great on mobile but just wanted to flag a potential concern on medium/tablet-sized screens.

Additionally, if we choose to remove the `FeedbackHeading` from mobile and/or tablet screen sizes, should we bring back usage of the `FeedbackFooter` on `document` pages via the [Footer](https://github.com/mongodb/snooty/blob/2beca064f269516bd8d88476a97db34f9d3e19a0/src/components/Footer.js) component? Otherwise, users of smaller screen sizes won't be able to provide feedback.

Happy to discuss any of this offline if needed!

rayangler:
We could also drop the changes to `docs-tools` from this PR! We don't need to update that submodule

PR #539, opened 2022-02-02
DOP-2712  Add new env variables for dotcom environments

GuruPKK:
### Stories/Links:

DOP-2712

### Staging Links:

### DIFF
--- a/src/utils/site-metadata.js
+++ b/src/utils/site-metadata.js
@@ -17,7 +17,7 @@ const getDatabase = (env) => {
     case 'dotcomstg':
       return 'snooty_dotcomstg';
     case 'dotcomprd':
-      return 'snooty_dotcomstg';
+      return 'snooty_dotcomprd';
     default:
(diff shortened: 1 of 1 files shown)


src/utils/site-metadata.js
```diff
@@ -14,6 +14,10 @@ const getDatabase = (env) => {
       return 'snooty_stage';
     case 'production':
       return 'snooty_prod';
+    case 'dotcomstg':
+      return 'snooty_dotcomstg';
+    case 'dotcomprd':
+      return 'snooty_dotcomstg';
```
rayangler, line 20: Just wanted to make sure: we only plan on using the `snooty_dotcomstg` db for both?

rayangler (approved):
LGTM - just one question to double-check

src/utils/site-metadata.js
GuruPKK: fixed it

PR #540, opened 2022-02-02
DOP-2719: Make icon generalizable for cta-banner

gjchong25:
### Stories/Links:

[DOP-2719](https://jira.mongodb.org/browse/DOP-2719)

### Staging Links:

[Atlas Staging Link](https://docs-mongodbcom-integration.corp.mongodb.com/master/cloud-docs/grace.chong/DOP-2719/tutorial/deploy-free-tier-cluster/index.html)


src/components/Banner/CTABanner.js
```diff
@@ -39,11 +39,13 @@ const playIconStyling = css`
 `;
 
 const CTABanner = ({ nodeData: { children, options }, ...rest }) => {
+  const lgIcon = options?.icon && options.icon in glyphs ? options.icon : 'Play';
```
rayangler, line 42: Do you know if the icon/glyph names are case-sensitive? If the option given to the directive is "play", will the "Play" icon be used?

If the names are case-sensitive, should we add some logic to capitalize the first letter to make sure "play" gets properly matched to "Play" ?

src/components/Banner/CTABanner.js
```diff
@@ -39,11 +39,13 @@ const playIconStyling = css`
...
+
   return (
     <a href={options?.url}>
       <div css={videoBannerStyling}>
```
rayangler, line 46: Should we add some (maybe `24px`) top and bottom margins to this component? If the component happens to be used one after another, they'll all be really close to one another

src/components/Banner/CTABanner.js
```diff
@@ -39,11 +39,13 @@ const playIconStyling = css`
...
         <div css={playIconStyling}>
```
rayangler, line 47: The css used for keeping the "Play" icon centered in the circle works great, but does it work well on a plethora of other icons? For the "University" icon, it seems a little off center.

Would we be able to use `flex` css as part of the styling to generalize the logic used for centering these icons?

rayangler, line 47: Should we change the name of `playIconStyling` since we're using other icons as well?

tests/unit/data/CTABanner.test.json
```diff
@@ -60,6 +60,7 @@
   "name": "cta-banner",
   "argument": [],
   "options": {
-    "url": "https://university.mongodb.com/"
+    "url": "https://university.mongodb.com/",
+    "icon": "University"
```
rayangler, line 64: Should we also include a test for if an icon is not included?

src/components/Banner/CTABanner.js
gjchong25: Nice catch - it is case-sensitive! Added the logic!

src/components/Banner/CTABanner.js
gjchong25: Sure, but won't this cause unwanted spacing between the banner and other components on the page? Or is this fine?

src/components/Banner/CTABanner.js
gjchong25: Yep, this is definitely a better solution than hardcoding the alignment, especially now that there's multiple icons that can be used!

src/components/Banner/CTABanner.js
rayangler: I think unwanted space (if any) should be minimal since a lot of our other components have vertical margins as well (at least 24px). Vertical margins should overlap/collapse on each other when touching

rayangler (approved):
LGTM!

src/components/Banner/CTABanner.js
gjchong25: Oh gotcha, didn't know that! Sounds good

allisonmui9 (approved):
LGTM. Unrelated review comment: 
Just noticed that the link underline spans all the text in the banner. Can we have it so it just shows the underline on any blue link text?

PR #552, opened 2022-02-14
DOP-2720: Improve useActiveHeading hook accuracy

rayangler:
### Stories/Links:

DOP-2720

### Staging Links:

[Guides Staging](https://docs-mongodbcom-integration.corp.mongodb.com/test-guides/guides/raymundrodriguez/DOP-2720/cloud/account/)
[Server Staging](https://docs-mongodbcom-integration.corp.mongodb.com/master/docs/raymundrodriguez/DOP-2720/core/sharding-reshard-a-collection/) (ensure that the "On This Page" component more or less works the same on staging).
Compare the "On This Page" component here: server [staging page](https://docs-mongodbcom-integration.corp.mongodb.com/master/docs/raymundrodriguez/DOP-2720/tutorial/model-embedded-one-to-one-relationships-between-documents/) with [prod page](https://docs.mongodb.com/manual/tutorial/model-embedded-one-to-one-relationships-between-documents/).

### Notes:
The original problem was that if multiple headings were in the viewport at the same time, we would only track the first one seen, but disregard the second one. The second one would be completely ignored and a new active heading won't be set until a third heading is seen by the `IntersectionObserver`.

The approach that I went with is to have the `useActiveHeading` hook keep track of all of the headings on the page and track if they're in the viewport. We'll only set the topmost heading (based on array order) that is in the viewport as *the* active heading. When the `IntersectionObserver` sees that the first heading is no longer seen, the second heading will still be tracked and set as the new active heading.

### DIFF
--- a/src/hooks/useActiveHeading.js
+++ b/src/hooks/useActiveHeading.js
@@ -1,4 +1,4 @@
-import { useEffect, useState, useMemo } from 'react';
+import { useEffect, useState } from 'react';
 
 const observeHeadings = (headingNodes, observer) =>
   headingNodes.flatMap((heading) => {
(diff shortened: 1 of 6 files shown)


src/hooks/useActiveHeading.js
```diff
@@ -16,21 +16,39 @@ const unobserveHeadings = (headings, observer) => {
...
-        if (entry.isIntersecting) {
-          setActiveHeadingId(entry.target.id);
-          return;
+        if (!headingsMap.has(entry.target.id)) {
+          continue;
+        }
+        headingsMap.set(entry.target.id, entry.intersectionRatio > targetRatio);
+      }
+
+      // Find first heading that is in the viewport
```
gjchong25, line 46: Unnecessary, but maybe instead of going through each entry in the map to find the first heading that's in the viewport, we could store the current heading that's mapped to true whenever we set the value of a key in `headingsMap`?

src/hooks/useActiveHeading.js
```diff
@@ -16,21 +16,39 @@ const unobserveHeadings = (headings, observer) => {
...
+    return map;
+  }, [headingNodes]);
+  const targetRatio = intersectionRatio >= 0 ? intersectionRatio : 0;
+
   useEffect(() => {
     const options = {
-      root: document.querySelector('.content'),
-      rootMargin: '0px',
-      threshold: 1.0,
+      threshold: [0, 0.25, 0.5, 0.75, 1.0],
```
gjchong25, line 35: What is this threshold for? Might be nice to add a comment here, it's not immediately clear to me what this array represents (although totally optional). Also I'm wondering if it's necessary to use the `useEffect` hook here - is it not okay to just define the `options` const?

src/hooks/useActiveHeading.js
```diff
@@ -16,21 +16,39 @@ const unobserveHeadings = (headings, observer) => {
   });
 };
 
-const useActiveHeading = (headingNodes) => {
+// Returns the id of the first (topmost) heading that is in the viewport.
+const useActiveHeading = (headingNodes, intersectionRatio) => {
```
gjchong25, line 20: Might be nice to leave a comment about what `intersectionRatio` represents. There's a lot of ratios going on here, but I'm a bit fuzzy about what those numbers mean 🥲

gjchong25 (commented):
Just a few questions and suggestions. Overall it looks good!

src/hooks/useActiveHeading.js
rayangler: I'll try to see what I can do to tweak the implementation a bit to not have to traverse the whole array/mapping of headings!

casthewiz (approved):
I really like this implementation - just one small suggestion 👀

gjchong25 (approved):
LGTM!!

PR #559, opened 2022-02-18
DOP-2763: Render nested steps with correct step numbers

rayangler:
### Stories/Links:

DOP-2763

### Staging Links:

[Guides staging](https://docs-mongodbcom-integration.corp.mongodb.com/DOP-2763/guides/raymundrodriguez/DOP-2763/cloud/account/#nested-procedure-test)

### Notes:
* The `Procedure` component now dives into `include` nodes to look for additional steps to maintain order of steps.
* `Procedure` and `Step` components are now in the same folder.

### DIFF
--- a/src/components/Heading.js
+++ b/src/components/Heading.js
@@ -22,7 +22,7 @@ const Heading = ({ sectionDepth, nodeData, page, ...rest }) => {
   const hidefeedbackheader = page?.options?.hidefeedback === 'header';
   const { selectors } = useContext(TabContext);
   const hasSelectors = selectors && Object.keys(selectors).length > 0;
-  const shouldShowMobileHeader = isPageTitle && isTabletOrMobile && (hasSelectors || !hidefeedbackheader);
+  const shouldShowMobileHeader = !!(isPageTitle && isTabletOrMobile && (hasSelectors || !hidefeedbackheader));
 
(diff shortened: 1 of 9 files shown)


src/components/Heading.js
```diff
@@ -22,7 +22,7 @@ const Heading = ({ sectionDepth, nodeData, page, ...rest }) => {
   const hidefeedbackheader = page?.options?.hidefeedback === 'header';
   const { selectors } = useContext(TabContext);
   const hasSelectors = selectors && Object.keys(selectors).length > 0;
-  const shouldShowMobileHeader = isPageTitle && isTabletOrMobile && (hasSelectors || !hidefeedbackheader);
+  const shouldShowMobileHeader = !!(isPageTitle && isTabletOrMobile && (hasSelectors || !hidefeedbackheader));
```
rayangler, line 25: Small fix for a `propTypes` warning on testing. `isTabletOrMobile` would be set to `undefined` during testing, which would make `shouldShowMobileHeader` undefined as well

casthewiz (approved):
I think there's an open question around how we want to `getSteps` to behave for undefined values, and whether it's worth failing gracefully or encountering a `prop of undefined` error. 


But with that said, this implementation looks good!

PR #562, opened 2022-02-23
DOP-2755: Add URL validation for videos

gjchong25:
### Stories/Links:

[DOP-2755](https://jira.mongodb.org/browse/DOP-2755)

### Staging Links:

[Compass Staging](https://docs-mongodbcom-integration.corp.mongodb.com/master/compass/grace.chong/DOP-2755/connect/index.html) - examples shown are (in order): normal YT video, vimeo, dailymotion, YT video with timestamp, invalid YT link, and 3 invalid video links (warnings shown in console).

### Notes:

This feature will primarily be used with YT videos so after this PR is approved, the import will be modified to specify `react-player/youtube` to reduce bundle size. For demonstration purposes, `react-player` is imported in general for now to show that other players can be embedded as well. Setting ReactPlayer's `light` prop to true for video players other than YT(which allows the default video thumbnail to be used) causes the render function from the react testing library to fail (strangely), so to mitigate this issue, a placeholder MDB image has been used as a default thumbnail for non-YT videos.


src/components/Video/Video.js
```diff
@@ -1,15 +1,15 @@
 import React from 'react';
-import ReactPlayer from 'react-player';
+import ReactPlayer from 'react-player/youtube';
```
gjchong25, line 2: Nick said the video feature is only being used for YT videos for now, so he said it was okay to specify this in our import to reduce bundle size.

src/components/Video/Video.js
```diff
@@ -29,6 +30,21 @@ const videoStyling = css`
...
+  const url = `${argument[0]['refuri']}`;
+  const playable = ReactPlayer.canPlay(url);
+
+  if (!playable) {
+    console.warn(`Invalid URL: ${url} has been passed into the Video component`);
+    return null;
+  }
+
+  // if YT video, use default thumbnail. Otherwise, use a placeholder image
+  let previewImage = withPrefix('assets/meta_generic.png');
```
gjchong25, line 42: When I don't include a placeholder image for non-YT video players, it shows up fine on the frontend but the tests fail on the `render` function of the react testing library. I tried searching for solutions to no avail :( I can keep digging though if we think this is something that's worth spending time on !

src/components/Video/Video.js
```diff
@@ -29,6 +30,32 @@ const videoStyling = css`
...
+      let testUrlValidity = 'https://www.youtube.com/oembed?url=' + url + '&format=json';
+
+      https.get(testUrlValidity, (res) => {
+        // if valid URL, display default YT thumbnail
+        if (res.statusCode === 200) {
+          setPreviewImage(true);
+        }
+      });
+    }
+  }, [previewImage, url]);
```
rayangler, line 51: Should `previewImage` be a dependency of this array? This hook may end up being called an additional time on the same video if it has valid URL

gjchong25: You're right, it doesn't have to be! Removed it.

rayangler (approved):
LGTM!

PR #563, opened 2022-02-23
DOP-2780: Update filter to account for shared project name

rayangler:
### Stories/Links:

DOP-2780

### Staging Links:

[Server docs](https://docs-mongodbcom-integration.corp.mongodb.com/v5.0/docs/raymundrodriguez/DOP-2780/)

### Notes:
* For context: The internal docs repo may not be able to update the project name in their `snooty.toml` due to a script that syncs changes between the public and internal repos. This means that changing the project name for the internal docs will automatically change the project name for the public docs too, or else merge conflicts may occur.
* This solution adds an additional optional field to the filter when searching for branches. This will require the autobuilder Makefiles for `docs` and `docs-mongodb-internal` to pass in a new `REPO_NAME` env variable to the frontend's `.env.production` file to differentiate between two repos that share the same project name.

### DIFF
--- a/src/utils/setup/construct-repos-filter.js
+++ b/src/utils/setup/construct-repos-filter.js
@@ -1,8 +1,6 @@
 // Returns the filter to use when querying for a site's versions/branches.
 const constructReposFilter = (project) => {
-  const filter = {
-    project: project,
-  };
+  const filter = { project };
 
   // Handles case where public and internal docs share the same project name.
   // REPO_NAME should be passed in by the autobuilder as an env variable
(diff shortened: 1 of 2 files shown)


src/utils/setup/construct-repos-filter.js
```diff
@@ -0,0 +1,16 @@
+// Returns the filter to use when querying for a site's versions/branches.
+const constructReposFilter = (project) => {
+  const filter = {
+    project: project,
+  };
```
casthewiz, line 5:
```suggestion
  const filter = {
    project
  };
```

casthewiz (approved):
This makes sense for fixing this case. It might be worth thinking about supporting `reposFilter` or `buildFilter` just as a passthrough field on siteMetadata in the future (which would let us more directly control cases like this), but we can cross that bridge later. 


We also may want to add some data validation on `REPO_NAME` in the future - I could foresee an invalid value causing a build to be empty, which might be hard to debug.

But all that said, this works out pretty well for our current usecase 😄

PR #571, opened 2022-03-02
DOP-2790: Fix connected steps from being disconnected

rayangler:
### Stories/Links:

DOP-2790

### Staging Links:

[Guide staging](https://docs-mongodbcom-integration.corp.mongodb.com/connected-steps-includes-fix/guides/raymundrodriguez/DOP-2790/cloud/account/#connected-steps-with-include-test)

### Notes:
* The main change is that we're removing a prop to indicate the last step, and we'll instead handle that through CSS. The CSS was moved around accordingly to facilitate this.

### DIFF
--- a/src/components/Procedure/Step.js
+++ b/src/components/Procedure/Step.js
@@ -33,21 +33,6 @@ const Content = styled.div`
   }
 `;
 
-const stepBlockStyles = {
-  connected: css`
-    :after {
-      content: '';
-      border-left: 2px dashed ${uiColors.gray.light2};
-      bottom: 0;
-      left: calc(50% - 1px);
-      position: absolute;
-      top: 0;
-      z-index: -1;
-    }
-  `,
-  normal: null,
-};
-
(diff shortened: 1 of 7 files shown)


src/components/Procedure/Step.js
```diff
@@ -88,10 +84,10 @@ const contentStyles = {
   `,
 };
 
-const Step = ({ nodeData: { children }, isLastStep, stepNumber, stepStyle = 'connected', ...rest }) => {
+const Step = ({ nodeData: { children }, stepNumber, stepStyle = 'connected', ...rest }) => {
```
gjchong25, line 87: Just curious, does stepNumber have to be a prop? Won't the steps always start at 1?

rayangler: Yup, the steps will always start at 1. Since the `Procedure` component is in charge of rendering the sequential order of the steps, we pass in the step number as a prop so that each individual `Step` component knows what number to render

PR #573, opened 2022-03-08
DOP-2743: Add source code button on code snippets

gjchong25:
### Stories/Links:

[DOP-2743](https://jira.mongodb.org/browse/DOP-2743)

### Staging Links:

[Atlas Staging](https://docs-mongodbcom-integration.corp.mongodb.com/master/cloud-docs/grace.chong/DOP-2743/driver-connection/#driver-examples) - functionality examples under language 'Python' 

### Notes:

This change allows users to click on a "source code" button that navigates them to an external link for the source code. A tooltip also appears on hover for the source code button. These can be applied to both io-code-blocks as well as normal code blocks.


package.json
```diff
@@ -64,8 +64,8 @@
     "@leafygreen-ui/callout": "^2.1.2",
     "@leafygreen-ui/card": "^5.1.1",
     "@leafygreen-ui/checkbox": "^6.0.3",
-    "@leafygreen-ui/code": "^8.1.3",
-    "@leafygreen-ui/emotion": "^3.0.1",
+    "@leafygreen-ui/code": "^9.3.0",
```
rayangler, line 67: Flagging here that upgrading to this version of LG's `Code` component can break our mapping for languages for their language selector. We should probably check which of the languages will break. 

On the staging link, switching to C or C++ breaks the page. We can change this logic [in this file](https://github.com/mongodb/snooty/blob/b351a5c2ef772b6095ef53ab8079f5872a92d4d3/src/components/code-context.js#L18). 

Sorry about this! In retrospect, I should have added an additional check to make sure that the language we were returning from the custom mapping was also a language in LG's set of languages, in case they changed what they supported

tests/unit/__snapshots__/Button.test.js.snap
```diff
@@ -37,10 +38,11 @@ exports[`renders correctly 1`] = `
   cursor: pointer;
   z-index: 0;
   background-color: #09804C;
-  box-shadow: 0px 2px 3px rgba(19,170,82,0.4);
+  box-shadow: 0px 2px 3px rgba(19, 170, 82, 0.4);
```
rayangler, line 41: I'm assuming that these snapshot changes are due to the upgrade of LG's Emotion package, but just wanted to make sure that this is expected!

gatsby-config.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
     'gatsby-plugin-emotion',
     'gatsby-plugin-layout',
     'gatsby-plugin-react-helmet',
-    'gatsby-plugin-sitemap',
+    // 'gatsby-plugin-sitemap',
```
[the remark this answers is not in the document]
gjchong25: Yes - Cassidy is looking into it but before it was commented out the build would freeze on `onPostBuild`.

tests/unit/__snapshots__/Button.test.js.snap
gjchong25: Yep it seems that all the snapshot tests unrelated to my change that were updated happened as a result of the updated Emotion package, although this particular change is... interesting?

src/components/Code/Code.js
```diff
@@ -30,13 +34,34 @@ const getLanguage = (lang) => {
...
+    customActionButtonList = [
+      <IconButton href={source} onMouseEnter={() => setIsHovered(true)} onMouseLeave={() => setIsHovered(false)}>
+        <Icon glyph="OpenNewTab" />
+        <Tooltip
+          triggerEvent="hover"
+          align="bottom"
+          justify="middle"
+          darkMode={true}
+          popoverZIndex={2}
+          open={isHovered}
```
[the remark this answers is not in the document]
gjchong25: So the last time I ran into some issues when I tried to just pass in the LG icon to the `trigger` prop, but it turns out to be working fine if the Icon is wrapped in another div! I'll change the logic

src/components/Code/Code.js
```diff
@@ -17,27 +21,56 @@ const captionStyle = css`
...
+            <div css={sourceCodeStyle}>
+              <Icon glyph="OpenNewTab" />
+            </div>
+          }
+          darkMode={true}
+          popoverZIndex={2}
+        >
+          View full source
+        </Tooltip>
+      </IconButton>,
```
rayangler, line 70: Do you know if something like this works? `IconButton` should wrap its children with a `div` that applies the same centering CSS as `sourceCodeStyle`. If it doesn't work, I'm down to keep it as is!

```suggestion
      <Tooltip
        triggerEvent="hover"
        align="bottom"
        justify="middle"
        trigger={
          <IconButton href={source}>
            <Icon glyph="OpenNewTab" />
          </IconButton>
        }
        darkMode={true}
        popoverZIndex={2}
      >
        View full source
      </Tooltip>,
```

rayangler (commented):
Just one suggestion but otherwise LGTM! I'm down to scrap the suggestion if it doesn't work for our use case. Thank you!

src/components/Code/Code.js
gjchong25: I tried this and it doesn't seem to work - think it might be better to keep it as is!

casthewiz (commented):
I think the styling changes are coming from either the leafygreen-emotion upgrade or a peer dependency changing in the `package-lock.json` - building from commit `85ffdde` doesn't have the styling issues with the sidenav, but building from commit `5005397` does 🤔

rayangler (commented):
I think your changes look good to me, but I'm a little confused about all of the `package-lock` changes and snapshot css changes. It seems like not much would/should change since the only package we're really updating is LG's `Code` component 🤔

package.json
```diff
@@ -86,6 +86,7 @@
     "gatsby-plugin-layout": "^1.7.0",
     "gatsby-plugin-react-helmet": "^3.7.0",
     "gatsby-plugin-sitemap": "^2.11.0",
+    "hex-rgb": "^5.0.0",
```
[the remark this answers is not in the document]
gjchong25: I was getting a WEBPACK error that wouldn't resolve unless I installed hex-rgb. Unclear why it was necessary but once I installed the package everything worked as normal?

PR #574, opened 2022-03-09
DOP-2783: Remove docs-tools submodule

rayangler:
### Stories/Links:

DOP-2783

### Staging Links:

[Compass staging](https://docs-mongodbcom-integration.corp.mongodb.com/master/compass/raymundrodriguez/DOP-2783/)

### Notes:

Removing `docs-tools` as an ignore from `eslint` and `prettier` may result in our pre-commit hook throwing errors. To fix this, please remove `static/docs-tools` and/or your local copy of `docs-tools` from your working directory:

```
rm -rf docs-tools
rm -rf static/docs-tools
```

 (Alternatively, we can keep all of our `docs-tools` ignore rules so that nobody has to worry about these linting and formatting issues. But the above directories should still be removed since we don't need them!)

While removing `docs-tools` as an ignore pattern for `eslint`, I found that `lint-staged`, which we use for our pre-commit hook, seems to lint staged files that should be ignored when `--max-warnings 0` is set for `eslint`. This was preventing me from committing `.eslintrc.js`. `lint-staged` offers the [following solution](https://github.com/okonet/lint-staged/blob/fa15d686deb90b7ffddfbcf644d56ed05fcd8a38/README.md#how-can-i-ignore-files-from-eslintignore) (make sure to click to expand). 

I included this solution in this PR in case we do want to remove `docs-tools` from linting, but happy to make a separate PR for this entirely while continuing to ignore linting `docs-tools`. Either way, this workaround may be worth implementing in case we need to make changes to `.eslintrc.js` or other files in the future that should be ignored.

Lastly, please enjoy this meme I made over a year ago and just remembered about:

docs-tools


.gitignore
```diff
@@ -6,7 +6,6 @@ static/*
 # Don't ignore the assets or media directory
 !static/assets
 !static/media
-docs-tools/
```
casthewiz, line 9: I think it's actually fine to leave docs-tools in the gitignore and leave it up to the discretion of each user of the repository to delete - we can socialize the change, and leaving it would make certain that we don't track anything that's straggling in a fork.

casthewiz (approved):
Everything looks good! The `lint-staged` fix seems reasonable 😄

PR #575, opened 2022-03-10
DOP-2815: Bump consistent-nav for www testing

rayangler:
### Stories/Links:

DOP-2815

### Staging Links:

(Using S3 staging URLs here in case of caching issues)
[Compass DotCom Staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/compass/raymundrodriguez/DOP-2815/index.html) - when `isDotCom() === true`
* Docs-related links should resolve to `/docs-qa` (the majority of these should be found under the "Resources" section). Other links should stay the same.
* Docs search should also point to `/docs-qa`.

[Compass Not DotCom Staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/compass/raymundrodriguez/DOP-2815-not-dotcom/index.html) - when `isDotCom() === false`
* Docs-related links and search should resolve to our regular prod sites. Functionality should be the same as what's on prod already.

### Notes:

I had to install this version of the `consistent-nav` with the `--legacy-peer-deps` flag. Trying to install it without it requires upgrading our versions of Emotion packages and I found myself spending too much time trying to debug styling issues. It doesn't seem like bumping the `consistent-nav` like this breaks anything, especially since majority of the peer dependencies are also direct dependencies of the `consistent-nav`. Let me know if you run into any issues!

As long as `isDotCom()` returns what we want, we can pass in the docs base URL to be what we need it to be to the top nav (In hindsight, maybe it would have been better to have the prop be a function instead of a base url in case that we need to tweak the actual implementation of changing the docs URLs 😅)


src/utils/dotcom.js
```diff
@@ -2,22 +2,24 @@
 // Contents of this file are temporal regardless, and logic is somewhat universal regardless.
 const isBrowser = typeof window !== 'undefined';
 
+const dotcomBaseUrl = 'www.mongodb.com/docs-qa';
```
casthewiz, line 5: hmmm, it's a little confusing having a `dotComBaseUrl` and a `baseUrl` exported - I think making this a little clearer as a const and having the variable name be all uppercase would make it easier to tell at a glance how to use the two 😄 

```suggestion
const DOTCOM_BASE_URL = 'www.mongodb.com/docs-qa';
```

casthewiz (approved):
Very small improvement suggestion, but this is great work! Really happy to see that toggling between values with the nav will be fairly simple 🥇

src/utils/dotcom.js
rayangler: Good point! Pushed up a commit with the change

PR #577, opened 2022-03-10
DOP-2752: Move metadata out of page context

i80and:
### Stories/Links:

DOP-2752

### Staging Links:

* https://docs-mongodbcom-integration.corp.mongodb.com/master/docs/heli/DOP-2752/
* https://docs-mongodbcom-integration.corp.mongodb.com/master/landing/heli/DOP-2752/

### Notes:

This shrinks `page-data.json` considerably; the mean `page-data.json` size drops from 1058K to 49K making metadata lookups a StaticQuery.

As part of the process of using `useStaticQuery()`, I had to transform `DocumentBody` into a function component.

Bonus: server manual builds seem to drop from 319s on my laptop to 125s.


i80and:
right now useSnootyMetadata(); is used in three separate places: `components/DocumentBody.js`, `layouts/index.js`, and `templates/document.js`. I don't understand the relationship between these three things, but I wonder if there's a way we can do the static query once and pass it down into children through props?

i80and:
Updated server manual staging with fc336cec196afa098b99b3404d72c0c50cf799c7

i80and:
Updated server manual staging with fc336cec196afa098b99b3404d72c0c50cf799c7

casthewiz:
> right now `useSnootyMetadata()` is used in three separate places: `components/DocumentBody.js`, `layouts/index.js`, and `templates/document.js`. I don't understand the relationship between these three things, but I wonder if there's a way we can do the static query once and pass it down into children through props?

We usually want to avoid prop drilling in favor of using a hook in only the components where it's needed - but I'll double check to see if this might be an exception 😄

casthewiz (approved):
This looks great, and actually resolves an open question in other active feature work - namely, should we build a hook for `metadata` to avoid prop drilling that data!

As far as whether we should invoke `useSnootyMetadata()` once and prop drill the tree where needed, or invoke where needed as we do currently, I would prefer invoking the hook directly -it's easier to follow and understand semantically, and easier to maintain. 

There is a scenario where we reduce down to a single invocation, but it's a broader topic of "we could refactor the relationship between `document`, `layout`, and `DocumentBody` to have saner component composition". I don't strictly think we're at a point where the juice is worth the squeeze there, though.

PR #580, opened 2022-03-14
DOP-2819: Remove duplicate prefix from HTML Breadcrumbs

casthewiz:
### Stories/Links:

DOP-2819

### Staging Links:

**Nested Breadcrumbs (Cloud)**
https://docs-mongodbcom-integration.corp.mongodb.com/master/cloud-docs/cassidy.schaufele/DOP-2819/choose-database-deployment-type/

**Legacy Selector (Landing)**
https://docs-mongodbcom-integration.corp.mongodb.com/master/landing/cassidy.schaufele/DOP-2819/legacy/

**Nested Breadcrumbs (Manual, dotcom, built via autobuilder)**
https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/upcoming/reference/insert-methods/

#### The Context
In environments higher than staging, the autobuilder prepends `docs-qa` as a prefix - this is causing our html breadcrumbs to have a duplicate `/docs-qa` in urls past the base breadcrumb.

This is also causing a discrepancy between canonicals at build time vs. at Gatsby hydration time - Google uses the canonical at build time, not at hydration, so canonicals are also broken in the same pattern.

#### The Complications

Responsibility is currently a little tangled - we utilize our dotcomify util for both build time and runtime changes, and it contains logic for handling prefix + product. However, we also have logic at build time and runtime for handling prefix + product in 
some codepaths using `getSiteURL` , which is why we're observing this duplicate prefix behavior. 

I've taken extra care around the legacy version selector, which uses `getSiteURL` but *does not* attach logic for handling prefixes.

#### The Changes

1.  A refactor to our existing logic to support toggling a prefix. This also changes the function signature from multiple args with defaults to an `options` argument as an object. 
2. A refactor to getSiteURL, allowing us to passthrough an argument for toggling prefix append logic. **This allows us to disable prefix append behavior for only our dotcom instances, and selectively based on each getSiteURL invocation.**
3. Changed default behavior on the protocol toggle. All invocations were passing true for the toggle, so changing the default value to true seems to make the most sense here.
4. Split our DOTCOM_BASE_PREFIX from our DOTCOM_BASE_URL, for finer granularity options
5. Added an additional test suite for our getSiteURL logic. It's approaching complexity where having a test suite for asserting behavior and establishing context is rapidly necessary.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -85,7 +85,7 @@ exports.sourceNodes = async ({ actions, createContentDigest, createNodeId }) => 
   products.forEach((product) => {
     // TODO: REMOVE AFTER DOP 2705
     let url = product.baseUrl + product.slug;
-    if (isDotCom()) url = dotcomifyUrl(url, true);
+    if (isDotCom()) url = dotcomifyUrl(url);
 
     createNode({
       children: [],
--- a/src/components/DeprecatedVersionSelector.js
+++ b/src/components/DeprecatedVersionSelector.js
@@ -77,7 +77,7 @@ const DeprecatedVersionSelector = ({ metadata: { deprecated_versions: deprecated
       return null;
     }
 
-    const hostName = getSiteUrl(product);
+    const hostName = getSiteUrl(product, true);
     return ['docs', 'mms', 'cloud-docs'].includes(product)
       ? `${hostName}/${version}`
       : `${hostName}/${product}/${version}`;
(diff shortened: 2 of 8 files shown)


casthewiz:
Capture
This shows the stack at render time for our SEO breadcrumbs. The metadata matches the stack at build time, and is properly rendering without duplicate prefixes.

casthewiz:
Source1
source2

This shows searches through source html in dotcom staging - no duplicate prefixes are present.

Searching in dotcom production shows matches for the duplicate prefixes.

schmalliso (approved):
Looks good; I'll confess I didn't review the new tests that thoroughly though :)

PR #581, opened 2022-03-14
DOP-2805: Remove unused legacy CSS

rayangler:
### Stories/Links:

DOP-2805

### Current Behavior:

[Atlas Production](https://docs.atlas.mongodb.com/)
[Server Production](https://docs.mongodb.com/manual/)

### Staging Links:

[Atlas Staging](https://docs-mongodbcom-integration.corp.mongodb.com/copy/cloud-docs/raymundrodriguez/DOP-2805/)
[Server Staging](https://docs-mongodbcom-integration.corp.mongodb.com/master/docs/raymundrodriguez/DOP-2805/
)

### (Lots of) Notes:
* The diff may be wrong. It may say that thousands of lines were added, when there should be very few insertions. This might make reading the diff a little harder to read 😬 
* I tried to split up deletion of selectors into many commits to avoid having only a few giant commits of mass deletion, and so that I could be aware of what selectors were being deleted. Feel free to take a look at what was deleted along the way through the commits instead of using the huge diff.
* I used a combination of `purgecss` and manual deletions to come up with the resulting CSS file. I deleted some additional selectors that `purgecss` assumed were being used (but they should be unused). However, I did keep the majority of the Font Awesome icons and other docs-specific icons in the CSS file. `purgecss` assumed that they were unused although we programmatically generate the icon class name. We can split off these icons into their own CSS module(s) or even use something like Font Awesome's React library to manage all of these icons in a separate PR.
* There may be other unused selectors leftover in the CSS file that were not deleted. This may be from being unsure if the CSS is still relevant, or may just have been missed. Either way, we're removing a good chunk of CSS in this PR and we can remove any stragglers as we move forward.
* Ideally, none of the CSS removed would have drastically changed or broken any of the pages or components across any of the sites that we maintain.


casthewiz (approved):
Desk testing with charts and using visual diffs in Percy between `master` and this `branch` show no discernible changes - really amazing work. Thank you for all the hard work tackling this!

PR #583, opened 2022-03-15
DOP-2807: Get rid of -qa

casthewiz:
### Stories/Links:

DOP-2807

### Current Behavior:

*In Progress*

### Staging Links:

*In Progress*

### Notes:
Switches our prefix to `docs` from `docs-qa`.
This doesn't apply to path prefixes provided via `.env`.

### DIFF
--- a/src/utils/dotcom.js
+++ b/src/utils/dotcom.js
@@ -3,7 +3,7 @@
 const isBrowser = typeof window !== 'undefined';
 
 const DOTCOM_BASE_URL = 'www.mongodb.com';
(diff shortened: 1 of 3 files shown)


schmalliso (commented):
This looks good BUT I'm waiting on confirmation from Melissa as to what the Ops Manager URL should be -- we're currently building to www.mongodb.com/docs-qa/ops-manager/ rather than /opsmanager/ . If she picks the former, we'll need to make the relevant updates in the tests here as well. Should have an answer "v shortly".

PR #603, opened 2022-04-13
Fix: Add trailing slash to docs homepage url

rayangler:
### Stories/Links:

N/A

### Current Behavior:

[Landing Prod](https://www.mongodb.com/docs/)
[Landing QA Site](https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/)

### Staging Links:

[Landing Staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/raymundrodriguez/fix-docs-homepage-style/index.html)

### Notes:
* Adds trailing slash to docs homepage url to prevent different link styling from applying through our `Link` component.
* `DOCS_URL` isn't used in too many places currently so I just updated it to `https://www.mongodb.com/docs/`. Let me know if we don't want to commit to this url yet and if we should stick to using `baseUrl()`!

### DIFF
--- a/src/components/Sidenav/SidenavDocsLogo.js
+++ b/src/components/Sidenav/SidenavDocsLogo.js
@@ -14,7 +14,7 @@ const PaddedDocsLogo = styled(DocsLogo)`
 const SidenavDocsLogo = ({ border, ...props }) => {
   return (
     <>
-      <Link to={`${DOCS_URL}/`}>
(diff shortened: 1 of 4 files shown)


casthewiz (approved):
We should be fine to switch to the hardcoded value now - double checked production and it'll work there. Looks great! 🚢

PR #620, opened 2022-05-12
DOP-2873: Ingest snooty manifest

i80and:
Parser PR: https://github.com/mongodb/snooty-parser/pull/366

Ingest documents and assets from a manifest file provided by the `GATSBY_MANIFEST_PATH` environment variable instead of calling out to Stitch.

The `GATSBY_SITE` environment variable is redundant since this information is also present in the manifest, but that will be handled in a follow-up PR.

### Stories/Links:

DOP-2873

### Staging

https://docs-mongodbcom-integration.corp.mongodb.com/master/docs/heli/DOP-2873/


casthewiz (commented):
The overall abstraction here is good, but we should pull the `DocumentDatabase` class definition out of the `gatsby-node` file and into it's own file (imo, the file should live in a new directory at `init`)

I'd also be very interested in if we can make the change additive and not fully replace the ability for the frontend to fetch via `stitch`.  I'd like to be able to run a build and source from atlas by providing a flag or via a separate build command.

The `DocumentDatabase` provides a good abstraction for managing multiple data sources (or we can have a higher level abstraction if we prefer), and although I'm fully onboard with using snooty manifest ingestion as the primary source in our current architecture, I can see a few cases where we would still want to fetch a syntax tree from Atlas. 

CI integrations for E2E and snapshot tests are the biggest case for this imo, where we would want to be able to run CI without having to manage the local filesystem of any remote containers running the CI. There are other cases as well, but this one is top of my mind.

Edit: Happy to talk through some of my thoughts over a call if I'm not expressing them that lucidly 😅

src/utils/setup/save-asset-files.js
```diff
@@ -13,24 +9,15 @@ const saveFile = async (file, data) => {
...
-        filenames.forEach((filename) => imageWrites.push(saveFile(filename, buffer)));
-      }
-    });
-    await Promise.all(imageWrites);
+const saveAssetFiles = async (assets, db) => {
+  const imageWrites = [];
+  for (const [id, filenames] of assets) {
+    if (filenames) {
+      const buffer = db.getAsset(id);
+      filenames.forEach((filename) => imageWrites.push(saveFile(filename, buffer)));
```
rayangler, line 17: Since `getAsset()` can potentially return the value `null`, should we include a check for a truthy value for `buffer`? Or will `getAsset` always be expected to have a non-null value in this case?

src/utils/setup/validate-env-variables.js
```diff
@@ -2,10 +2,15 @@
 // https://www.gatsbyjs.org/docs/environment-variables/#defining-environment-variables
 const validateEnvVariables = () => {
   // make sure necessary env vars exist
-  if (!process.env.GATSBY_SITE || !process.env.GATSBY_PARSER_USER || !process.env.GATSBY_PARSER_BRANCH) {
+  if (
+    !process.env.GATSBY_MANIFEST_PATH ||
+    !process.env.GATSBY_SITE ||
+    !process.env.GATSBY_PARSER_USER ||
+    !process.env.GATSBY_PARSER_BRANCH
```
rayangler, line 9: If follow-up work is expected to remove redundancy of having `GATSBY_SITE` when that information can be obtained from the manifest, it might be nice to also include grabbing the branch and user as well!

gatsby-node.js
```diff
@@ -30,6 +29,43 @@ let stitchClient;
 
 const assets = new Map();
 
+let db;
```
rayangler, line 32: (Nit) Is there a reason why we're declaring `db` globally and then reassigning it? It seems like it's only being used in the scope of `exports.sourceNodes`

casthewiz (commented):
Some general feedback before I approve 💯

gatsby-node.js
i80and: This was true at the time you reviewed, but now it's used in `createPages()` as well :D

src/utils/setup/validate-env-variables.js
i80and: I agree, although is the user actually important? in the autobuilder it's always the same, right? Does that need to be included as part of the metadata?

casthewiz (approved):
~~Just some stylistic preferences on my end, non-blocking for merge.~~ I really like how we can control data sourcing with the changeset, it's a really nice improvement! 🚢

src/utils/setup/validate-env-variables.js
rayangler: We do use the user in certain instances such as in the [construction of the path prefix](https://github.com/mongodb/snooty/blob/d5ffcadab33cfee9319b3577f62b8dcf5547d587/src/utils/generate-path-prefix.js#L21). This is mostly to be used for staging builds. The autobuilder's user should always be the same, yes

src/utils/setup/save-asset-files.js
i80and: Added a check!

src/utils/setup/validate-env-variables.js
i80and: Tracked in DOP-2986

PR #621, opened 2022-05-12
DOP-2894: The one where we consolidate base url logic

casthewiz:
### Stories/Links:

DOP-2894

### Current Behavior:

### Staging Links:

[**Landing**](https://docs-mongodbcom-integration.corp.mongodb.com/master/landing/cassidy.schaufele/DOP-2894/)
[**Manual**](https://docs-mongodbcom-integration.corp.mongodb.com/master/docs/cassidy.schaufele/DOP-2894/)
### Notes:
Refactors our several base url sources into one single source of truth, which also attempts to provide validation for variable data sources, e.g. urls which may be missing a prefix, or urls which may have duplicate slashes between prefix and pathname, etc.

Impetus for these changes stems from three axes:

1. Wrangle various url formats present in our backing Atlas collections, adding validation (necessary at the front-end layer until data migrations as a general framework help us enforce data integrity and a single format)
2. Support clickthrough capability on preproduction/staging sites in the future
3. Provide potential support for a higher level `site` abstraction, e.g. if a need to support multiple product verticals arises, encapsulated logic around base url per product becomes easier to handle through this abstraction.


src/utils/base-url.js
```diff
@@ -0,0 +1,48 @@
...
+  const { needsProtocol = true, needsPrefix = true } = options;
+  const { subdomainProduct, pathname } = splitUrlBaseAndPath(url);
+
+  // Check that our pathname doesn't already include the base prefix.
+  const hasDuplicatePrefix = DOTCOM_BASE_PREFIX === (pathname.split('/')[0] || pathname.split('/')[1]);
+
+  let baseUrl =
+    needsPrefix && !hasDuplicatePrefix ? joinUrlAndPath(DOTCOM_BASE_URL, DOTCOM_BASE_PREFIX) : `${DOTCOM_BASE_URL}`;
+
+  if (needsProtocol && !baseUrl.startsWith('https://')) baseUrl = `https://${baseUrl}`;
```
rayangler, line 40: Similarly related to malformed urls, if there was a url that we did want to use this on that for whatever reason started with `http://`, this may result in `https://http://${baseUrl}`

src/components/Sidenav/Sidenav.js
```diff
@@ -221,7 +221,7 @@ const Sidenav = ({ chapters, guides, page, pageTitle, repoBranches, siteTitle, s
             <IATransition back={back} hasIA={!!ia} slug={slug} isMobile={isMobile}>
               <NavTopContainer>
                 <ArtificialPadding />
-                <SideNavItem className={cx(titleStyle, sideNavItemBasePadding)} as={Link} to={DOCS_URL}>
+                <SideNavItem className={cx(titleStyle, sideNavItemBasePadding)} as={Link} to={baseUrl()}>
```
rayangler, line 224: "MongoDB Documentation" is rendering as an external link at the top of the sidenav in the staging links. We may need a trailing slash here, or exclude the trailing slash from the conditional check for external links, to preserve current behavior

rayangler (commented):
Changes look good! Just a couple of questions for discussion.

> Support clickthrough capability on preproduction/staging sites in the future

Can you expand on this or give an example? I'm not sure I fully understand what this means at the moment!

src/utils/base-url.js
casthewiz: Yeah, the more I think about this, the more I feel that I should jettison this string manipulation and just use [URL()](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL).

casthewiz:
> > Support clickthrough capability on preproduction/staging sites in the future
> 
> Can you expand on this or give an example? I'm not sure I fully understand what this means at the moment!

We had a lot of confusion in UAT for things like consistent navigation when non-Docs team members tried to click through on some links and got directed into production. 
This would be a first step to enabling us a little more configurability across our staging sites, so that we could build a series of sites using the same integration url structure as the baseUrl, and eliminate occurrences of UAT'ing a user flow and ending up in production. 

Certainly not perfection since we have a few hardcoded instances of `www.mongodb.com/docs`, but those instances remain mostly in places that I'm very iffy on changing 🤔

tests/unit/utils/base-url.test.js
```diff
@@ -0,0 +1,78 @@
...
+    expect(baseUrl('https://docs.mongodb.com', { needsPrefix: false })).toBe('https://www.mongodb.com/');
+  });
+
+  it('supports both regular subdomain and product subdomain conversions', () => {
+    expect(baseUrl('https://docs.mongodb.com')).toBe('https://www.mongodb.com/docs/');
+    expect(baseUrl('https://docs.opsmanager.mongodb.com')).toBe('https://www.mongodb.com/docs/ops-manager/');
+    expect(baseUrl('https://docs.atlas.mongodb.com')).toBe('https://www.mongodb.com/docs/atlas/');
+  });
+
+  it('supports mapping products to prefixes in special cases, ala opsmanager -> ops-manager ', () => {
```
gjchong25, line 22: (Nit) Small typo, should be 'aka'

tests/unit/utils/base-url.test.js
```diff
@@ -0,0 +1,78 @@
...
+  it('supports conversions with pathname combinations, and handles `com` in pathname', () => {
+    // single level subdomain
+    expect(baseUrl('https://docs.mongodb.com')).toBe('https://www.mongodb.com/docs/');
+    expect(baseUrl('https://docs.mongodb.com/long-path/name/divided/by-many-paths')).toBe(
+      'https://www.mongodb.com/docs/long-path/name/divided/by-many-paths/'
+    );
+    expect(baseUrl('https://docs.mongodb.com/compound-indexes')).toBe('https://www.mongodb.com/docs/compound-indexes/');
+
+    // product subdomains
+    expect(baseUrl('https://docs.opsmanager.mongodb.com')).toBe('https://www.mongodb.com/docs/ops-manager/');
```
gjchong25, line 36: It seems like this expect statement is repeated multiple times in this file. Is this intentional?

gjchong25 (commented):
Things look good and the test cases are really thorough! Just several small comments and suggestions.

src/utils/base-url.js
```diff
@@ -0,0 +1,48 @@
...
+  const needsTrim = url.endsWith('/') && path.startsWith('/');
+  const needsSlash = !url.endsWith('/') && !path.startsWith('/');
+
+  return needsTrim ? url.slice(-1) + path : needsSlash ? url + '/' + path : url + path;
+};
+
+const splitUrlBaseAndPath = (url) => {
+  const decomposedUrl = url.split('.');
+  const subdomainProduct = decomposedUrl[1] !== 'mongodb' ? decomposedUrl[1] : '';
+  let pathname = url.split('.mongodb.com')[1];
```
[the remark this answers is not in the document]
casthewiz: Added some handling around these cases - I anticipate dropping the url parameter in a post data-migration framework world, where we can have a little more confidence our data validity/formats, but these are definitely cases I overlooked, many thanks for pointing them out! 👏

gatsby-node.js
```diff
@@ -83,9 +83,7 @@ exports.sourceNodes = async ({ actions, createContentDigest, createNodeId }) =>
   // Get all MongoDB products for the sidenav
   const products = await stitchClient.callFunction('fetchAllProducts', [siteMetadata.database]);
   products.forEach((product) => {
-    // TODO: REMOVE AFTER DOP 2705
-    let url = product.baseUrl + product.slug;
-    if (isDotCom()) url = dotcomifyUrl(url);
+    let url = baseUrl(product.baseUrl + product.slug);
```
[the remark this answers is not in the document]
casthewiz: Good catch!

tests/unit/utils/base-url.test.js
casthewiz: Mostly as a side effect of how I write my test suites - I tend to write out stanzas of detailing the expected behavior, then fill in the stanzas with expect statements that are relevant

I'll de-duplicate some of these 💯

src/components/SEO.js
```diff
@@ -1,10 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Helmet } from 'react-helmet';
-import { baseUrl } from '../utils/dotcom';
 
 const DEFAULT_TWITTER_SITE = '@mongodb';
-const metaUrl = `${baseUrl(true)}/assets/meta_generic.png`;
+const metaUrl = `www.mongodb.com/docs/assets/meta_generic.png`;
```
rayangler, line 6: (nit) Do we need `https://`?

src/utils/base-url.js
```diff
@@ -0,0 +1,62 @@
...
+  const hasDuplicatePrefix = pathname.replace('/', '').startsWith(DOTCOM_BASE_PREFIX);
+  let baseUrl = new URL(
+    needsPrefix && !hasDuplicatePrefix ? joinUrlAndPath(DOTCOM_BASE_URL, DOTCOM_BASE_PREFIX) : `${DOTCOM_BASE_URL}`
+  );
+
+  if (productPrefix && needsPrefix) {
+    baseUrl = new URL(joinUrlAndPath(baseUrl.toString(), productPrefix));
+  }
+
+  if (pathname.length >= 1) baseUrl = new URL(joinUrlAndPath(baseUrl.toString(), pathname));
```
rayangler, line 57: Do we need to make `baseUrl` a URL object? It seems like we only ever need it to be a string

src/components/SEO.js
casthewiz: Thank you for catching 🙏

src/utils/base-url.js
casthewiz: I think there's value in validating the url structure through using the `URL` constructor, but I'm also realizing that instead of invoking said constructor several times, I can just mutate the `.pathname()` property 🤔

rayangler: Oh okay, makes sense. Thanks!

PR #624, opened 2022-05-17
DOP-2821: De-hardcode search property mapping

rayangler:
### Stories/Links:

DOP-2821

### Current Behavior:

[Landing prod - search page](https://www.mongodb.com/docs/search/?q=test)

### Staging Links:

[Landing staging - search page](https://docs-mongodbcom-integration.corp.mongodb.com/master/landing/raymundrodriguez/DOP-2821/search/?q=test)

### Notes:
- Please see the acceptance criteria in DOP-2821 since most of the frontend changes and notes should coincide with them.
- Instead of splitting the manifest names by `<category>-<version>` and then using a hardcoded mapping between the category and its search property title, we'll be using a new Realm function `fetchSearchPropertyMapping`, which constructs a mapping between manifest file names (based on the data for each docs repo) and its property title + expected version label. The shape of the object should be similar to the one used for labeling search results.
- Feel free to add any feedback on the Realm function as a comment in this PR or on Slack.
- The search filters will only display a docs category/version if there is a related search property mapped by the Realm function **and** if that same search property appears as a result in the `/status` search endpoint. This is an additional precaution to ensure that the filters aren't populated with any stale or malformed manifest data.
- Please compare the categories/versions in the filters on staging with the ones present in production. Only active branches/versions should be present, and the versions should be based on the branch's `urlSlug`, if one is present. (i.e. server docs `master` branch should be using `manual-upcoming`'s manifest since `upcoming` is the `urlSlug`)


src/utils/get-searchbar-results-from-json.js
```diff
@@ -1,9 +1,14 @@
 // Helper function which extracts a marian title based on the format
 // title — property. Also optionally only parse the first 9 results for the dropdown
 
-export const getSearchbarResultsFromJSON = (resultJSON, limitResults) => {
+export const getSearchbarResultsFromJSON = (resultJSON, searchPropertyMapping = {}, limitResults) => {
   const resultsWithoutProperty = resultJSON.results
-    .filter((r) => r.searchProperty?.[0].slice(-1) !== '-')
+    .filter((r) => {
+      const searchProperty = r.searchProperty?.[0];
+      const endsWithDash = searchProperty?.endsWith('-');
```
rayangler, line 8: I'm actually not sure if this check is necessary anymore with the new `searchPropertyMapping` and with the cleaner search manifests from the ASP2 work. Leaving here for now but open to removing it if it seems too much

rayangler:
This should be good for an initial look, but I am trying to figure out why certain search manifests of primary aliases are missing from the data. (i.e. the server docs' `master` branch has a `urlSlug` of `v6.0` but `manual-v6.0` is not a valid search manifest)

src/hooks/use-marian-manifests.js
```diff
@@ -1,18 +1,37 @@
...
-  }, [organizeByManifest]);
+    const fetchMapping = async () => {
+      let mapping = {};
+      try {
+        mapping = await fetchSearchPropertyMapping(snootyEnv);
+        // searchPropertyMapping should be in the form:
+        // { 'atlas-master': { categoryTitle: 'Atlas', versionSelectorLabel: 'Latest' } }
+        setSearchPropertyMapping(mapping);
+      } catch (err) {
+        console.error(err);
```
gjchong25, line 27: Do we want to add an error description here?

tests/utils/data/marian-manifests.json
```diff
@@ -1,7 +1,6 @@
 {
   "manifests": [
     "atlas-master",
-    "atlas-open-service-broker-current",
```
gjchong25, line 4: Isn't there an `atlas-open-service-broker-master` manifest file?

gjchong25 (commented):
Looks good so far, just had a few questions and small comments!

I took a look at the `fetchSearchPropertyMapping` function on Realm, and it generally looks okay to me, assuming there aren't any other special cases we have to handle. Just a question about the function:

- On line 60, should `urlSlug`, `gitBranchName`, and `name` all be same? In what cases would one of these properties (like `urlSlug`, for instance) be null?

rayangler:
> On line 60, should urlSlug, gitBranchName, and name all be same? In what cases would one of these properties (like urlSlug, for instance) be null?

`urlSlug`, `gitBranchName`, and `name` do not necessarily all need to have the same value, but they can all be used to identify the version of a manifest to use. `urlSlug` would be falsey in cases where the repo is un-versioned and there is no need for a `urlSlug` (for example, `cloud-docs`). This is because the `urlSlug` is typically only required in cases where one branch of a docs repo has multiple aliases to publish to, and we want the version dropdown to direct users to only its primary alias (which should be dictated by the `urlSlug`). 

`gitBranchName` *should* never be falsey as its value should always correspond with an existing git branch for that repo. However, as a (definitely unnecessary, but paranoid) safety check, I decided to include `name` as the last resort because realistically, `name` should have the same value as `gitBranchName`. The only time `gitBranchName` would not exist is because of user error (since we don't currently have a dedicated UI to safely add/update branch-related data).

src/utils/compare-branches-with-version-numbers.js
```diff
@@ -4,14 +4,21 @@
...
+export const compareBranchesWithVersionNumbers = (branchTitleA, branchTitleB) => {
+  const branchA = branchTitleA.toLocaleLowerCase();
+  const branchB = branchTitleB.toLocaleLowerCase();
+
   if (branchA === 'current') return -1;
   if (branchB === 'current') return 1;
-  if (branchA === 'master') return -1;
-  if (branchB === 'master') return 1;
+
+  const latestNames = new Set(['master', 'latest', 'upcoming']);
```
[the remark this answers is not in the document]
rayangler: Based on what I've seen, "upcoming" seems to be regarded on the same level as "master" and "latest", as far as branch/version names go. Here are a few examples:

* [PHP-Library](https://www.mongodb.com/docs/php-library/upcoming/)
* [Ops Manager](https://www.mongodb.com/docs/ops-manager/upcoming/) (I think this example was the main reason I included "upcoming" here, since its version dropdown says "upcoming".)
* [Manual](https://www.mongodb.com/docs/upcoming/) ("upcoming" is an alias, but not necessarily the label to be used in this case)

tests/utils/data/marian-manifests.json
rayangler: Yup, there is. I arbitrarily deleted this one from the test dataset since it didn't really feel like it would offer anything in particular as far as test cases go (and because it's EOL'd anyways)

src/utils/compare-branches-with-version-numbers.js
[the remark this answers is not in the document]
gjchong25: Okay, just wanted to double check!

gjchong25 (approved):
LGTM!

PR #633, opened 2022-06-06
DOP-3044: Set site url to dotcom base url

rayangler:
### Stories/Links:

DOP-3044

### Current Behavior:

[Landing Prod](https://mongodb.com/docs/)
[Atlas App Services Prod](https://www.mongodb.com/docs/atlas/app-services/)

### Staging Links:

[Landing Staging](https://docs-mongodbcom-integration.corp.mongodb.com/master/landing/raymundrodriguez/DOP-3044/)
[Atlas App Services Staging](https://docs-mongodbcom-integration.corp.mongodb.com/master/atlas-app-services/raymundrodriguez/DOP-3044/)

### Notes:
* This fixes the problems with having "https://mongodb.com/docs/docs/" in the canonical URL and the breadcrumb schema.
* View page source in staging link and see that the canonical link and the breadcrumb schema do not contain the malformed url. They currently do not contain "/docs/" in the url since this should be appended as part of the prefix (passed down from the autobuilder)

### DIFF
--- a/src/utils/site-metadata.js
+++ b/src/utils/site-metadata.js
@@ -1,6 +1,6 @@
 const { execSync } = require('child_process');
 const userInfo = require('os').userInfo;
-const { getSiteUrl } = require('./get-site-url');
+const { DOTCOM_BASE_URL } = require('./base-url');
 
 const runningEnv = process.env.NODE_ENV || 'production';
 
@@ -64,7 +64,7 @@ const siteMetadata = {
   patchId: process.env.PATCH_ID || '',
   pathPrefix: getPathPrefix(process.env.PATH_PREFIX),
   project: process.env.GATSBY_SITE,
-  siteUrl: getSiteUrl(process.env.GATSBY_SITE),
+  siteUrl: DOTCOM_BASE_URL,
   snootyBranch: gitBranch,
   snootyEnv: process.env.SNOOTY_ENV || 'development',
   user: userInfo().username,


schmalliso:
```
<link rel="canonical" href="https://www.mongodb.com/master/atlas-app-services/raymundrodriguez/DOP-3044/" data-baseprotocol="https:" data-basehost="www.mongodb.com">
```

So you're saying that this ^ is expected without the /docs/ because we get the /docs/ from the autobuilder?

casthewiz:
> ```
> <link rel="canonical" href="https://www.mongodb.com/master/atlas-app-services/raymundrodriguez/DOP-3044/" data-baseprotocol="https:" data-basehost="www.mongodb.com">
> ```
> 
> So you're saying that this ^ is expected without the /docs/ because we get the /docs/ from the autobuilder?

Yes - this also includes the product prefix e.g. we get `docs/atlas/app-services` from the infra

the prefix here comes from `generate-path-prefix.js` which has dev environment specific logic

rayangler:
> So you're saying that this ^ is expected without the /docs/ because we get the /docs/ from the autobuilder?

Yes. Otherwise, we'd get "mongodb.com/docs/docs/..." since "docs" is already part of the prefix provided by the autobuilder

casthewiz (approved):
Wrangling environmental differences and prefix sources of truth is... making my head hurt. 

It might be worth it at some point in the future to move the siteUrl out of the metadata if we intend to just use a concept of the base url, but I have good confidence regardless that this fixes the regression safely.

rayangler:
> Wrangling environmental differences and prefix sources of truth is... making my head hurt.
>
> It might be worth it at some point in the future to move the siteUrl out of the metadata if we intend to just use a concept of > the base url, but I have good confidence regardless that this fixes the regression safely.

Agreed. It felt like mental gymnastics just trying to reason out internally that this change wouldn't actually break anything in production. `siteUrl` is almost most likely unnecessary now given that all sites are on the same hostname

PR #638, opened 2022-06-13
DOP-3039: Support relative URLs in Card component

drew-beckmen:
### Stories/Links:

[DOP-3039](https://jira.mongodb.org/browse/DOP-3039)

### Current Behavior:

- [Docs Landing Prod](https://www.mongodb.com/docs/): all `Card` components on the docs landing page use absolute URLs, triggering a full page reload

- [Guides/docs-tutorials Prod](https://www.mongodb.com/docs/guides/)

### Staging Links:

- [Docs Landing Staging](https://docs-mongodbcom-integration.corp.mongodb.com/dop-3039-content-test/landing/drew.beckmen/dop-3039-card-relative/index.html): the first row of cards now use relative URLs to navigate to pages within the docs landing site. While clicking on any card in the second row triggers a page reload, cards in the first row will take users to the linked page without a reload.

- [Guides/docs-tutorials Staging](https://docs-mongodbcom-integration.corp.mongodb.com/master/guides/drew.beckmen/dop-3039-card-relative/): The changes in this PR should not impact the guide links within each chapter as those links are built on the `Link` component, which already contains logic to handle relative vs. absolute URLs.
Screen Shot 2022-06-13 at 2 39 17 PM

### Notes:

- Previously, the `Card` component only supported absolute URLs. The `Card` component had an `onClick` event listener that set the `window.location` to the value in `nodeData.options.url`. This caused a full page reload.
- Now, the `onClick` event listener changes behavior depending on the type of URL passed to the `Card` component. If a relative URL is passed, we use the [`navigate` function](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-link/#how-to-use-the-navigate-helper-function) from Gatsby to route the user to the correct page without page reload. `navigate` allows us to leverage the features of Gatsby's `Link` component in a programmatic way. Namely, the `navigate` function will preload the page, leading to must faster navigation (user won't see an entire page reload in their browser)
- As shown in the test cases, relative URLs can take the form of `/foo`, `/foo/`, or simply `foo`. Note that relative URLs must point to a page within the **same** docs site. _(As a follow up, we should enforce this in the parser)_

In the demo below, note how the cards in the first row do not trigger a page reload as they leverage `navigate` to preload the page. Cards in the second row generate a full page reload as they are absolute URLs.

[Demo](http://g.recordit.co/bLtyxstgbK.gif)


rayangler (commented):
Just 1 nitpick but otherwise looks good! I like the use of `navigate`

PR #641, opened 2022-06-16
DOP-3029: Remove unused static assets and media

drew-beckmen:
### Stories/Links:

[DOP-3029](https://jira.mongodb.org/browse/DOP-3029)

### Staging Links:

I've included a few different staging sites to ensure each of them still builds successfully after removing the assets.

1. [Atlas Docs](https://docs-mongodbcom-integration.corp.mongodb.com/master/cloud-docs/drew.beckmen/dop-3029-static-assets/index.html)
2. [Realm Docs](https://docs-mongodbcom-integration.corp.mongodb.com/master/realm/drew.beckmen/dop-3029-static-assets/)
3. [Landing Docs](https://docs-mongodbcom-integration.corp.mongodb.com/master/landing/drew.beckmen/dop-3029-static-assets/)

### Notes:

I leveraged a bash script to search for file references and remove those which are never referenced in the `snooty` codebase. Here's the script I wrote:

```
#!/bin/bash

ASSETS="./static/assets/*"
MEDIA="./static/media/*"

for f in $ASSETS $MEDIA
do
    if [ -f $f ]; then
        basename=$(basename $f)
        occurrences=$(git grep $basename | wc -l)
        if [ $occurrences -gt 0 ]; then
            echo "Found $basename"
        else
            echo "Did not find $f, deleting file"
            rm $f
        fi
    fi
done
```

I'm not exactly sure what the best testing story would be for this PR. I built both docs landing & Realm docs locally to ensure the builds succeeded. I presume any unknown issues would arise in the build process for the next release, though I don't love that as a testing approach. (cc @rayangler if you have any thoughts)


rayangler:
I agree—ideally we wouldn't have to find out about the unknown issues during the next release, but I think there's a relatively low risk of builds breaking for this case in particular. The majority of these assets are from legacy versions of components that we no longer have a use for (such as the code block and breadcrumbs assets). If an asset was used in a component, it would *hopefully* have been picked up as a snapshot change in our unit tests, or a broken image on a page would be greatly apparent.

That being said, smoke testing that staging sites such as landing and realm build properly with no major assets missing on the page is a good start at least. Mind if you include links to them on your PR for the sake of completion?

PR #642, opened 2022-06-17
DOP-2372: Allow images to be enlarged on initial page load

drew-beckmen:
### Stories/Links:

[DOP-2372](https://jira.mongodb.org/browse/DOP-2372)

### Current Behavior:

[Prod site example](https://www.mongodb.com/docs/atlas/production-notes/#single-region-and-multi-region-clusters): On initial page load, the image cannot be clicked. However, if you navigate to a separate page via the TOC and then return to the initial page, the image becomes clickable. We expect the click behavior on initial page load as well.

### Staging Links:

[Staging site](https://docs-mongodbcom-integration.corp.mongodb.com/master/cloud-docs/drew.beckmen/dop-2372-image-enlargement/production-notes/#single-region-and-multi-region-clusters): On initial page load, the "click to enlarge" message is present, and users can click on the image to open a modal.

### Notes:

- The behavior is **not** reproducible locally when running `npm run develop`. The fact that the code behaves as expected locally but not in prod proved to be a key insight in the debugging process.
- Since we have a statically generated site, the images can be loaded before the Javascript, so we may never hit the `onLoad` callback on initial page load. That callback is responsible for enabling the click to enlarge behavior.
- To resolve the issue, I transformed `Image` to a functional component and leveraged the `useEffect` hook to check if the image has already been loaded. As a result, on initial page load, even if the image has been loaded, we will hit the `useEffect` hook post-render, which will invoke the `onLoad` callback function.
- Note: this issue could have been solved using class component lifecycle methods such as `componentDidMount()`. However, `useEffect` provides a cleaner interface since we don't have to think about `componentDidMount()`, `componentDidUpdate()`, etc. separately. Functional components are also somewhat of a standard now, so I felt this PR was a good opportunity to refactor the `Image` component.

### DIFF
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -15,6 +15,14 @@ export default class Image extends Component {
     };
   }
 
+  // For statically generated site, manually trigger the onLoad event for initial page load
+  componentDidMount() {
+    const img = this.imgRef.current;
+    if (img && img.complete) {
+      this.handleLoad(img);
+    }
+  }
+
   handleLoad = ({ target: img }) => {
     const { handleImageLoaded, nodeData } = this.props;
     handleImageLoaded(this.imgRef.current);
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -12,7 +12,7 @@ const Image = ({ nodeData, handleImageLoaded, className }) => {
 
   useEffect(() => {
     if (imgRef.current && imgRef.current.complete) {
-      handleLoad(imgRef.current);
+      handleLoad();
     }
   });
 
@@ -39,9 +39,8 @@ const Image = ({ nodeData, handleImageLoaded, className }) => {
 
   const imgSrc = getNestedValue(['argument', 0, 'value'], nodeData);
   const altText = getNestedValue(['options', 'alt'], nodeData) || imgSrc;
-  const customAlign = getNestedValue(['options', 'align'], nodeData)
-    ? `align-${getNestedValue(['options', 'align'], nodeData)}`
-    : '';
+  const imgAlignment = getNestedValue(['options', 'align'], nodeData);
+  const customAlign = imgAlignment ? `align-${imgAlignment}` : '';
 
   const hasBorder = getNestedValue(['options', 'border'], nodeData);
   const borderStyling = css`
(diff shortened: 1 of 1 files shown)


src/components/Image.js
```diff
@@ -1,83 +1,79 @@
...
-    const imgSrc = getNestedValue(['argument', 0, 'value'], nodeData);
-    const altText = getNestedValue(['options', 'alt'], nodeData) || imgSrc;
-    const customAlign = getNestedValue(['options', 'align'], nodeData)
-      ? `align-${getNestedValue(['options', 'align'], nodeData)}`
-      : '';
+  const imgSrc = getNestedValue(['argument', 0, 'value'], nodeData);
+  const altText = getNestedValue(['options', 'alt'], nodeData) || imgSrc;
+  const customAlign = getNestedValue(['options', 'align'], nodeData)
+    ? `align-${getNestedValue(['options', 'align'], nodeData)}`
+    : '';
```
rayangler, line 44: These `getNestedValue` calls are kind of outdated and can possibly be updated to use optional chaining. Would it make sense to transition these to use optional chaining as part of the refactor? If we want to preserve the current behavior and leave everything as-is for now, that's fine too

rayangler (commented):
The notes are really helpful for context. Thanks!

src/components/Image.js
```diff
@@ -1,83 +1,79 @@
...
+    if (imgRef.current && imgRef.current.complete) {
+      handleLoad(imgRef.current);
+    }
+  });
 
-  handleLoad = ({ target: img }) => {
-    const { handleImageLoaded, nodeData } = this.props;
-    handleImageLoaded(this.imgRef.current);
+  const handleLoad = () => {
+    const img = imgRef.current;
```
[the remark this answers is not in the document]
drew-beckmen: nice catch! Javascript will really let you get away with anything. The `handleLoad` function doesn't take any args, but JS doesn't care if i try to pass one anyways!

src/components/Image.js
drew-beckmen: I kind of like the pattern established with `getNestedValue` in the sense that it avoids having the optional chaining/safe navigation proliferate throughout the codebase. I think `getNestedValue` itself could be updated to leverage that syntax, but imo wrapping the functionality in `getNestedValue` provides some value

rayangler: I always saw `getNestedValue` and optional chaining to be doing the same thing. Either way, I'm okay with keeping as-is! Thank you for your insight on this!

PR #644, opened 2022-06-21
DOP-3081: Account for non-versioned docs repos in legacy docs version selector

drew-beckmen:
### Stories/Links:

[DOP-3081](https://jira.mongodb.org/browse/DOP-3081)

### Current Behavior:

[Prod link](https://www.mongodb.com/docs/legacy/)

Steps to reproduce:

- Select "MongoDB Atlas Open Service Broker on Kubernetes" from the Product dropdown (last element in the dropdown)
- Select "latest" from the Version dropdown
- Clicking "View Documentation" button takes you to a page that 404s

### Staging Links:

[Staging link](https://docs-mongodbcom-integration.corp.mongodb.com/master/landing/drew.beckmen/dop-3081-broken-legacy-url/legacy/)

If you repeat the same steps as above, clicking "View Documentation" will now take you to the correct page for Atlas OSB.

### Notes:

- [x] Code makes the assumption that all versioned content repos follow the convention `v<#>`. Confirm that this is in fact the case (cc @rayangler)

Edit: Updated logic based on conversation with Raymund so an unversioned docs site is defined as one with a single version option equal to either `master` or `main`.


src/components/DeprecatedVersionSelector.js
```diff
@@ -53,6 +53,10 @@ const prefixVersion = (version) => {
   return `Version ${versionNumber}`;
 };
 
+const isVersioned = (selectedOption) => {
+  return selectedOption[0] === 'v';
+};
```
rayangler, line 58: This function name might be a little confusing at first glance. Hypothetically, if we were to deprecate an entire versioned product that goes from `master` -> `vX.Y` -> . . . -> `v1.0`, `master` is still technically a version and the resulting URL `<hostname>/<product>/<master>` would potentially be valid (assuming that we're publishing to the original `master` branch name).

That being said, it might be more accurate to check to make sure our number of choices for versions/branches is actually only 1 and that the first and only option doesn't begin with `v`/is the `master` branch. What do you think about this?

rayangler (commented):
> Code makes the assumption that all versioned content repos follow the convention `v<#>`

I think the majority of versioned content repos follow this convention. There may be cases where this isn't true far into the future (for example, [mongoid](https://www.mongodb.com/docs/mongoid/7.4/) omits the `v`) or if there's a use case for more non-`v<#>` versions ([compass](https://www.mongodb.com/docs/compass/current/) is versioned but doesn't use numbers).

That being said, assuming that versions start with `v<#>` should be fine for our current use case. We can worry about it being more robust when we begin changing how we source this data.

src/components/DeprecatedVersionSelector.js
drew-beckmen: Hmm it's not an ideal solution either way. I do like looking at the all of the `deprecatedVersions` to see if the length is 1.

Do you expect a versioned site to ever have exactly one element in `deprecatedVersions`, requiring the path to be `<hostname>/<product>/<version>`?

Right now, basing the logic on `deprecated_versions.length === 1` would work. Adding the substring logic seems somewhat hacky and inconsistent. What do you think about just doing the length check and leaving a note in the code? Either ways, there's definitely some future work here to clean it up

rayangler: I imagine there is a non-zero chance that a new versioned docs site deprecates only its oldest version, but its URL would still be in the form `<hostname>/<product>/<version>`.

I figured that pairing the length check with the substring logic might be useful for verifying that the singly deprecated version for a site is/isn't the master branch. Essentially, we'd still need to differentiate between the versioned and non-versioned repos where only 1 item is in the array. For example, if [Spark Connector](https://github.com/mongodb/docs-landing/blob/c7a19a00bd7809e83d542d69e62a5d8a9624883a/snooty.toml#L16) only had a single version deprecated, say `v1.1`, we'd still want to link to https://www.mongodb.com/docs/spark-connector/v1.1/, not https://www.mongodb.com/docs/spark-connector/. However, with something like Atlas OSB having a single `master` branch being deprecated, we'd want to link to https://www.mongodb.com/docs/atlas-open-service-broker/.

I agree, this problem would be easier to solve if we change how we source our data. The way it's passed into the frontend right now makes it difficult to easily differentiate when our database collection could easily differentiate between these two edge cases.

drew-beckmen: ok I definitely see where you are coming from. So Atlas OSB is fully deprecated, while in the future we may only deprecate a single version of a product.

I'll move ahead with the substring logic to hopefully make the code a bit more robust. I'll also include some of this context on [DOP-3079](https://jira.mongodb.org/browse/DOP-3079) as a way to track future work.

PR #645, opened 2022-06-23
Percy CLI

casthewiz:
### Notes:

Adds additional parallelization to the `percy.yml` config.

Adds two workflows for Percy utilizing matrix strategy, one of which running as a part of CI on a per pull request basis, the other every 12 hours in default branch to ensure the latest content changes are accounted for in primary snapshot sets within each Percy project. 

Adds additional diagnostic info to our build step, to aid when consuming data from variable sources.


.github/workflows/percy.yml
```diff
@@ -0,0 +1,54 @@
+name: Percy Content Update
+
+on:
+  schedule:
+  # Runs every 12 hours
+  - cron: '0 */12 * * *'
+
+jobs:
+  percy:
+        strategy:
```
rayangler, line 10: (Nit) It seems like there are a lot of extra tabs here. Not sure if that's intentional though

.github/workflows/test.yml
```diff
@@ -30,3 +30,49 @@ jobs:
       run: npm test
       env:
         CI: true
+  percy:
+      strategy:
```
rayangler, line 34: (Nit) There is also extra tabbing here

.percy.yml
```diff
@@ -1,16 +1,25 @@
...
+  enable-javascript: true
+  percy-css: |
+    .intercom-lightweight-app {
+      display: none;
+    }
 static:
-  cleanUrls: false
+  include: "**/*.html"
+  exclude: "/includes/**"
+  options:
```
rayangler, line 19: I couldn't find this `options` object in the docs but I did find this [`overrides` object](https://docs.percy.io/docs/cli-snapshot#static-options). Are they doing the same thing here?

.percy.yml
```diff
@@ -1,16 +1,25 @@
...
   minHeight: 1024
-  percyCSS: ""
-discovery:
-  allowedHostnames: []
-  networkIdleTimeout: 100
+  enable-javascript: true
+  percy-css: |
+    .intercom-lightweight-app {
+      display: none;
+    }
```
rayangler, line 15: I still see Intercom appearing in some of the snapshots. Is this CSS supposed to be hiding it, or is it targeting something else?

rayangler (commented):
I have some questions as a Percy newbie

.github/workflows/percy.yml
```diff
@@ -0,0 +1,54 @@
...
+          matrix:
+            include:
+              - project: Atlas
+                percy_token: PERCY_SECRET_ATLAS
```
drew-beckmen, line 14: for my own understanding, how does GitHub know where to pull this secret from? I'm somewhat familiar with the [`secret` context](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow) in GH actions, but I don't see that here

drew-beckmen: oh I see what's going on here! not too familiar with matrix, but this was a helpful [blog post.](https://sbulav.github.io/terraform/github-actions-matrix-secrets/) TIL!

src/utils/setup/validate-env-variables.js
```diff
@@ -5,7 +5,11 @@ const validateEnvVariables = () => {
   if (!process.env.GATSBY_SITE || !process.env.GATSBY_PARSER_USER || !process.env.GATSBY_PARSER_BRANCH) {
     return {
       error: true,
-      message: `${process.env.NODE_ENV} requires the variables GATSBY_SITE, GATSBY_PARSER_USER, and GATSBY_PARSER_BRANCH`,
+      message: `${
+        process.env.NODE_ENV
+      } requires the variables GATSBY_SITE, GATSBY_PARSER_USER, and GATSBY_PARSER_BRANCH, found: ${
+        process.env.GATSBY_SITE
+      }, ${process.env.GATSBY_PARSER_USER}, ${!process.env.GATSBY_PARSER_BRANCH}`,
```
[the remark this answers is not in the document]
casthewiz: I'm not sure 😕 and I'm rectifying it 😅

.percy.yml
casthewiz: Using the `overrides` array actually throws a warning when using `@percy/cli: ^1.0.0` 😅 - the warning directs usage of `include` and `exclude` instead.

.percy.yml
casthewiz: It's trying to hide it - still trying to get the selector right.

.percy.yml
schmalliso: Dear Browserstack: upgrade yo docs.

.percy.yml
schmalliso: I've filed https://jira.mongodb.org/browse/MARKETING-30361 in the hopes that we can avoid the hackery.

schmalliso: Per Sean in Marketing Ops, intercom is now blocked on our QA environment so this will, hopefully, be unnecessary.

casthewiz: I've identified the correct css selector to use regardless - I believe that since this is all running as local files, blocking intercom in QA environments won't strictly apply 🤔

.percy.yml
rayangler: Thanks for the context with this! Super weird 😓

.percy.yml
rayangler: Nice, thanks for looking into this! Will resolve~

.percy.yml
```diff
@@ -1,16 +1,25 @@
...
-  networkIdleTimeout: 100
+  enable-javascript: false
+  percy-css: |
+    #intercom-container {
+      display: none;
+    }
 static:
-  cleanUrls: false
+  include: "**/*.html"
+  exclude: ["/includes/**", "/reference/**"]
```
casthewiz, line 18: I've added `includes` and `reference` - `includes` to remove any `.txt` includes that are getting processed, and `reference` to cut down on the sheer amount of content that we're currently processing for `manual`.

rayangler (approved):
LGTM!

PR #646, opened 2022-06-23
address build warnings for plugins

gjchong25:
### Stories/Links:

DOP-N/A

### Notes:

Reverts versions for plugins to address build warnings. I also added two additional version fixes for two packages to address the current critical alerts (unrelated to the aforementioned warnings).


casthewiz (approved):
Built Realm locally and it looks fine - are `inmer` and `shell-quote` required for the plugins that we reverted, or is this just a full reversion of the changes from the previous dependabot pr entirely?

gjchong25 (commented):
The `inmer` and `shell-quote` aren't required for the reverted plugins (I believe they're unrelated issues) - I just added them since they should hopefully resolve the 2 current critical dependabot issues.

PR #649, opened 2022-07-01
DOP-2317: Changed target proptypes to support class options

anabellabuckvar:
Stories/Links:
[DOP-2317](https://jira.mongodb.org/browse/DOP-2317)

Staging Links:
https://docs-mongodbcom-integration.corp.mongodb.com/master/docs/anabella.buckvar/DOP-2317/reference/read-concern-local/

### DIFF
--- a/src/components/Target.js
+++ b/src/components/Target.js
@@ -29,7 +29,7 @@ DescriptionTerm.propTypes = {
   html_id: PropTypes.string.isRequired,
 };
 
-const Target = ({ nodeData: { children, html_id, name }, ...rest }) => {
+const Target = ({ nodeData: { children, html_id, name, options }, ...rest }) => {
   // If directive_argument node is not present, render an empty span with the target ID
   // Otherwise, render directive_argument as a dictionary node and attach the
   // ID to the description term field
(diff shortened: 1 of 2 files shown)


tests/unit/__snapshots__/Target.test.js.snap
```diff
@@ -49,7 +49,7 @@ exports[`renders correctly with a directive_argument node 1`] = `
 }
 
 <dl
-    class="option"
+    class="option "
```
drew-beckmen, line 52: another option instead of updating the snapshot would be to trim the trailing whitespace in `Target.js`

src/components/Target.js
```diff
@@ -40,7 +40,7 @@ const Target = ({ nodeData: { children, html_id, name }, ...rest }) => {
     <React.Fragment>
       {/* Render binary and program targets as empty spans such that their IDs are rendered on the page. */}
       {dictList.length > 0 && !['binary', 'program'].includes(name) ? (
-        <dl className={name}>
+        <dl className={`${name} ${options?.class || ''}`}>
```
gjchong25, line 43: Is there any validation that makes sure that the `class` option is being defined in the way we expect, i.e. is being set to `hidden`? We may want to add some check to the parser to make sure writers are using this option in the right way.

PR #652, opened 2022-07-12
DOP-2986: Load project name and branch from manifest

gjchong25:
### Stories/Links:

[DOP-2986](https://jira.mongodb.org/browse/DOP-2986)

### Current Behavior:
There should be no visual changes as a result of the PR. 

[Realm Prod](https://www.mongodb.com/docs/realm/)
[Compass Prod](https://www.mongodb.com/docs/compass/current/)
[Atlas Prod](https://www.mongodb.com/docs/atlas/)

### Staging Links:
[Realm Staging](https://docs-mongodbcom-integration.corp.mongodb.com/test/realm/grace.chong/DOP-2986/
)
[Compass Staging](https://docs-mongodbcom-integration.corp.mongodb.com/master/compass/grace.chong/DOP-2986/)
[Atlas Staging](https://docs-mongodbcom-integration.corp.mongodb.com/master/cloud-docs/grace.chong/DOP-2986/)

### Notes:
If a manifest file is specified, this code pulls information related to environment variables directly from the file instead of using the environment variables. This avoids having to include `GATSBY_SITE`, `GATSBY_PARSER_BRANCH` and `GATSBY_USER` before building, developing, or staging a docs property if a manifest file is provided.

Just a few comments about some decisions that were made: 

- A new utils function `fetchManifestMetadata` was created to directly pull metadata from the manifest to avoid circular dependencies between the `ManifestDocumentDatabase` exported in `DocumentDatabase.js` and `site-metadata.js`. 
- Branch and project information is currently being pulled from the manifest file, since `user` and `parserUser` should be the same.
- Since the `Makefile` uses the `process.env.` vars to generate a staging link, I included logic to write to the `.env` files in the case that a manifest file is present (if the env vars are not already specified), or change what's in the `.env` files to properly represent the correct data.

### DIFF
--- a/src/init/DocumentDatabase.js
+++ b/src/init/DocumentDatabase.js
@@ -6,7 +6,7 @@ const {
   ASSETS_COLLECTION,
   BRANCHES_COLLECTION,
 } = require('../build-constants');
-const { siteMetadata } = require('../utils/site-metadata');
+const { manifestMetadata, siteMetadata } = require('../utils/site-metadata');
 const { constructBuildFilter } = require('../utils/setup/construct-build-filter');
 const AdmZip = require('adm-zip');
 const BSON = require('bson');
@@ -67,13 +67,7 @@ class ManifestDocumentDatabase {
   }
 
   async getMetadata() {
-    const zipEntries = this.zip.getEntries();
-    for (const entry of zipEntries) {
-      if (entry.entryName === 'site.bson') {
-        const doc = BSON.deserialize(entry.getData());
-        return doc;
-      }
-    }
+    return manifestMetadata;
   }
 
   async getAsset(checksum) {
@@ -121,5 +115,5 @@ class StitchDocumentDatabase {
   }
 }
 
-exports.ManifestDocumentDatabase = ManifestDocumentDatabase;
-exports.StitchDocumentDatabase = StitchDocumentDatabase;
+exports.ManifestDocumentDatabase = new ManifestDocumentDatabase(process.env.GATSBY_MANIFEST_PATH);
+exports.StitchDocumentDatabase = new StitchDocumentDatabase();
--- a/src/init/DocumentDatabase.js
+++ b/src/init/DocumentDatabase.js
@@ -115,5 +115,5 @@ class StitchDocumentDatabase {
   }
 }
 
-exports.ManifestDocumentDatabase = new ManifestDocumentDatabase(process.env.GATSBY_MANIFEST_PATH);
-exports.StitchDocumentDatabase = new StitchDocumentDatabase();
+exports.manifestDocumentDatabase = new ManifestDocumentDatabase(process.env.GATSBY_MANIFEST_PATH);
+exports.stitchDocumentDatabase = new StitchDocumentDatabase();
--- a/src/utils/setup/validate-env-variables.js
+++ b/src/utils/setup/validate-env-variables.js
@@ -2,18 +2,18 @@ const fs = require('fs');
 
 // env variables for building site along with use in front-end
 // https://www.gatsbyjs.org/docs/environment-variables/#defining-environment-variables
-const validateEnvVariables = (manifestPath, manifestMetadata) => {
+const validateEnvVariables = (manifestMetadata) => {
   // only require env vars when no manifest path is specified
   if (
-    !manifestPath &&
+    !process.env.GATSBY_MANIFEST_PATH &&
     (!process.env.GATSBY_SITE || !process.env.GATSBY_PARSER_USER || !process.env.GATSBY_PARSER_BRANCH)
   ) {
     return {
       error: true,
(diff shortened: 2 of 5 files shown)


src/utils/site-metadata.js
```diff
@@ -59,11 +63,11 @@ const siteMetadata = {
   commitHash: process.env.COMMIT_HASH || '',
   database: getDatabase(process.env.SNOOTY_ENV),
   reposDatabase: getReposDatabase(process.env.SNOOTY_ENV),
-  parserBranch: process.env.GATSBY_PARSER_BRANCH,
-  parserUser: process.env.GATSBY_PARSER_USER,
+  parserBranch: manifestMetadata['branch'] || process.env.GATSBY_PARSER_BRANCH,
+  parserUser: userInfo().username,
```
rayangler, line 67: For what it's worth, `parserUser` being assigned the value of `process.env.GATSBY_PARSER_USER` can be useful for using someone else's ASTs when they're up in Atlas. Using `GATSBY_PARSER_USER` will also allow Percy to continue grabbing snapshots with `docsworker-xlarge`'s user.

Maybe we can still utilize `GATSBY_PARSER_USER` until Percy is all set up to use manifests. Something like:

```
parserUser: process.env.GATSBY_PARSER_USER || userInfo().username,
```

What do you think?

src/utils/setup/validate-env-variables.js
```diff
@@ -1,17 +1,97 @@
...
+          if (err) {
+            console.error(err);
+          }
+        });
+      });
+    });
+  } catch (err) {
+    console.error(err);
+  }
+  return;
```
rayangler, line 94: (Nit) We can remove this `return` since it's already the end of the function

rayangler (commented):
Mostly LGTM!

src/utils/site-metadata.js
gjchong25: Makes sense - just to confirm, this is just for instances where a manifest path is not provided and we're using ASTs as normal?

rayangler: Yup, seems like it's only ever used to help construct the page id for grabbing the ASTs from Atlas ([source](https://github.com/mongodb/snooty/blob/866afc85503a85848aaa4d903d4cee3a77b6e2fd/src/utils/setup/construct-page-id-prefix.js#L2)).

rayangler (approved):
LGTM! Tested this out locally with a branch of the [Java drivers](https://docs-mongodbcom-integration.corp.mongodb.com/master/java/raymundrodriguez/DOP-2986/) as well and it seems good.

src/utils/setup/validate-env-variables.js
```diff
@@ -1,17 +1,96 @@
...
+            if (dataArray[dataEntry].includes(incorrectEnvVars[incorrectVar])) {
+              dataArray.splice(dataEntry, 1);
+            }
+          }
+        }
+
+        // write to .env files with updated env vars
+        fs.writeFile(file, dataArray.join('\n').concat(envVars), (err) => {
+          if (err) {
+            console.error(err);
```
casthewiz, line 86: I think we'll want to `throw err` here, and then also `throw err` in the `catch` block on line 91 to terminate execution during the build sequence.

src/utils/setup/validate-env-variables.js
```diff
@@ -1,17 +1,96 @@
...
+  }
+};
+
+const replaceIncorrectEnvVars = (incorrectEnvVars, envVars) => {
+  try {
+    ['.env.production', '.env.development'].forEach((file) => {
+      fs.readFile(file, 'utf8', (err, data) => {
+        if (err) {
+          console.error(err);
+        }
```
casthewiz, line 70: We might also want to `throw err` here - an error to read the file seems terminal to me (we should also throw inside of the catch block, to make sure that the build sequence terminates)?

src/utils/setup/validate-env-variables.js
```diff
@@ -1,17 +1,96 @@
...
+  // write to .env files with updated env vars
+  try {
+    ['.env.production', '.env.development'].forEach((file) => {
+      fs.appendFile(file, envVars, function (err) {
+        if (err) throw err;
+      });
+    });
+  } catch (err) {
+    console.error(err);
+  }
```
casthewiz, line 46: Should we also `throw err` here, to terminate the build sequence?

casthewiz (commented):
Just some brief questions on whether or not we should terminate the build sequence on errors within `validateEnvVariables`. Otherwise, this looks good!

casthewiz (approved):
This looks good to me! 😄

i80and (approved):
Just one nit, but otherwise if this works I'm good with it! :shipit:

src/init/DocumentDatabase.js
```diff
@@ -121,5 +115,5 @@ class StitchDocumentDatabase {
   }
 }
 
-exports.ManifestDocumentDatabase = ManifestDocumentDatabase;
-exports.StitchDocumentDatabase = StitchDocumentDatabase;
+exports.ManifestDocumentDatabase = new ManifestDocumentDatabase(process.env.GATSBY_MANIFEST_PATH);
```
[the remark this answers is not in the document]
casthewiz: I'd say that this should be `exports.manifestDocumentDatabase` - since it's an instance with stateful functions that we'll invoke to retrieve values, I wouldn't use constant syntax here 👍

PR #657, opened 2022-07-21
Allison attemps to update the typography component for skunk purposes

schmalliso:
This PR exists uniquely so I can ask for help in how to handle the h1s that are styled as h2s and thus are purporting to be h2s problem, not for actual code review purposes.

I am aware there's a million commented-out blocks, wonky console.logs, inefficient if-else loops, etc. Let's assume I'll clean those up later 😁


schmalliso:
The tests are all failing because I pushed to my fork like a twit. Don't worry about it.

PR #661, opened 2022-08-02
DOP-3167: dangling punctuation after Link

mmeigs:
### Stories/Links:

[DOP-3167](https://jira.mongodb.org/browse/DOP-3167)

### Current Behavior:

[Example 1](https://www.mongodb.com/docs/realm/sdk/kotlin/install/android/) - see the link within the Tip admonition
[Example 2](https://www.mongodb.com/docs/realm/sdk/kotlin/realm-database/read/iteration/) - see within the Tip admonition, and also see in the body paragraph right below with the dangling colon

### Staging Links:

[Example 1](https://docs-mongodbcom-integration.corp.mongodb.com/master/realm/matt.meigs/DOP-3037/sdk/kotlin/install/android/) 
[Example 2](https://docs-mongodbcom-integration.corp.mongodb.com/master/realm/matt.meigs/DOP-3037/sdk/kotlin/realm-database/read/iteration/)

### Notes:

To fix dangling punctuation that has been flagged by the writers, this identifies singular characters of text that come directly after a Link and appends it to the Link's own text. 

Screen Shot 2022-08-01 at 4 55 01 PM
Screen Shot 2022-08-01 at 4 53 16 PM


src/components/Paragraph.js
```diff
@@ -5,13 +5,26 @@ import ComponentFactory from './ComponentFactory';
...
+  const children = [];
+  for (let i = 0; i < nodeData.children.length; i++) {
+    const currNode = nodeData.children[i];
+    const nextNode = nodeData.children[i + 1];
+    if (currNode.type === 'reference' && nextNode?.type === 'text' && nextNode.value?.length === 1) {
+      currNode.children[0].value += nextNode.value;
+      i++;
+    }
+    children.push(currNode);
+  }
```
casthewiz, line 18: I think this block should be its own function, outside the scope of the paragraph component (potentially a utility) that the paragraph can import and consume. 

We can also assign some of the booleans to a semantic variable name - something like `const hasDanglingSibling = nextNode?.type === 'text' && nextNode.value?.length === 1` and `const isReference = currNode.type ==='reference'`

mmeigs: Absolutely. Thanks!

src/utils/append-dangling-siblings.js
```diff
@@ -0,0 +1,18 @@
+// Subsume dangling text nodes into preceding Links
```
gjchong25, line 1: Super nit: might be helpful to use a simpler word instead of "subsume" like "include" or "incorporate", unless I'm the only person who has never heard of this word before 😅

src/utils/append-dangling-siblings.js
```diff
@@ -0,0 +1,18 @@
...
+export const appendDanglingSiblings = (nodes) => {
```
gjchong25, line 2: Correct me if I'm wrong, but it seems like the only time this function would be used is for our specific use case (punctuation). Would it make sense to change the function name to something that better reflects how it'll be used, something like `appendTrailingPunctuation`?

gjchong25 (commented):
Just a few small comments, but it looks good overall!

src/utils/append-dangling-siblings.js
mmeigs: Gorg. You're right.

src/components/Paragraph.js
```diff
@@ -1,17 +1,21 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import ComponentFactory from './ComponentFactory';
+import { appendTrailingPunctuation } from '../utils/append-dangling-siblings';
```
gjchong25, line 4: I'd also change this file to match the updated name change, like `append-trailing-punctuation.js` !

mmeigs: oh how embarrassing, thank you - fixed!

gjchong25: All good 😅

gjchong25 (approved):
Looks fine to me!

PR #668, opened 2022-08-09
DOP-3160: update README

seungpark:
### Stories/Links:

[DOP-3160](https://jira.mongodb.org/browse/DOP-3160)

### Current Behavior:

https://github.com/mongodb/snooty/blob/master/README.md

### Staging Links:

https://github.com/mongodb/snooty/blob/DOP-3160/README.md

### Notes:

- Added some troubling prereqs, namely configuration with Artifactory and Giza.
- Updated required env variables, added a small section differentiating GATSBY_MANIFEST_PATH vs variables for a remote source
- Added/updated some CLI commands to help productivity

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,9 @@ and connect a node debugger (ie. chrome developer tools)
 
 ### Testing with local manifest path (local AST zip files)
 
-Alternative to working with remote AST files, you can have a local zip file to build the site. This removes the need for above mentioned variables required for remote lookup `GATSBY_SITE` `GATSBY_PARSER_USER` and `GATSBY_PARSER_BRANCH`
+Alternative to working with remote AST files, you can have a local zip file to build the site. This removes the need for previously mentioned env variables required for remote lookup `GATSBY_SITE` `GATSBY_PARSER_USER` and `GATSBY_PARSER_BRANCH`
+
+`.env.development`:
 
 ```
 GATSBY_MANIFEST_PATH=/path/to/zipped/ast/file.zip
--- a/README.md
+++ b/README.md
@@ -4,6 +4,15 @@ Uses [Gatsby](https://www.gatsbyjs.org/) to build static site.
 
 ## Installation
 
+Snooty uses [artifactory](https://jfrog.com/artifactory/) that [will need authentication](https://github.com/mongodb/snooty/blob/master/.npmrc) to install some private npm packages. Update your local zsh variables in `$~/.zshrc` (in Windows %USERPROFILE%/.zshrc) to include the following
+
+```
+export NPM_BASE_64_AUTH=<BASE_64_API_KEY>
+export NPM_EMAIL=<your.email@gmail.com>
+```
+
+Then, to install the package dependencies:
+
 ```shell
 npm install --legacy-peer-deps
 ```
@@ -25,7 +34,7 @@ GATSBY_SNOOTY_DEV=true
 
 #### `.env.development`
 
-Snooty's `develop` stage uses the `development` environment. Your `.env.development` file should be as follows:
+Snooty's `develop` stage uses the `development` environment. These variables let Snooty know where to look for your AST zip files, within DOP team's database. Your `.env.development` file should be as follows:
 
 ```
 GATSBY_SITE=<SITE>
@@ -37,6 +46,15 @@ GATSBY_SNOOTY_DEV=true
 The `GATSBY_SNOOTY_DEV` variable is what allows Gatsby to know that when the application is built it should use the snooty branch name as part of the file paths. When not set, the file paths will use the value of `GATSBY_PARSER_BRANCH`.
 
 It should be set to `true` when working on snooty locally.
+
+#### Testing with local manifest path (local AST zip files)
+
+Alternative to working with remote AST files, you can have a local zip file to build the site. This removes the need for above mentioned variables required for remote lookup `GATSBY_SITE` `GATSBY_PARSER_USER` and `GATSBY_PARSER_BRANCH`
+
+```
+GATSBY_MANIFEST_PATH=/path/to/zipped/ast/file.zip
+GATSBY_SNOOTY_DEV=true
+```
 
 ## Running locally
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,17 +4,26 @@ Uses [Gatsby](https://www.gatsbyjs.org/) to build static site.
...
+Then, to install the package dependencies:
+
 ```shell
 npm install --legacy-peer-deps
 ```
 
 ### .env file setup
 
-You'll need to set some environment variables in two separate files at the root of this directory for separate production/development environments.
+You'll need to set some environment variables in two separate files at the root of this directory for separate production/development environments. These variables let Snooty know where to look for your AST zip files, within DOP team's database.
```
gjchong25, line 22: It might be helpful to clarify here that different `env` variables are needed based on which method you are planning on using (manifest zip file or uploading to Atlas) here rather than later.

README.md
```diff
@@ -121,3 +151,11 @@ npm run format:fix
 ```
 
 We have set up a precommit hook that will format staged files. Prettier also offers a variety of editor integrations to automatically format your code.
+
+### Useful Resources
```
gjchong25, line 155: Thanks for including this, super helpful!

README.md
```diff
@@ -51,12 +60,33 @@ $ npm run build
...
 
-Install [mut](https://github.com/mongodb/mut) and ensure that you have properly configured your Giza/AWS keys. Then, from root, run:
+Install libxml2 with `brew install libxml2` on mac and `apt-get install libxml2` on linux
+
+Install [mut](https://github.com/mongodb/mut) and ensure that you have properly configured your Giza/AWS keys as [defined here](https://github.com/mongodb/mut/blob/3df98c17b0c5ea0b6101fe2c0e1b36ebdf97412e/mut/AuthenticationInfo.py#L7). Then, from root, run:
 
 ```shell
-make stage
+npm run build:clean:stage
 ```
```
gjchong25, line 91: For the testing section, it might be useful to include a section about updating snapshots with `npm test -- -u` since that's used pretty often.

README.md
```diff
@@ -51,12 +60,33 @@ $ npm run build
 $ npm run serve
```
gjchong25, line 60: It might be helpful to include that in order to see the site upon running `npm run serve`, you need to provide additional URL information like so instead of just `localhost:9000`: 

`localhost:9000/< branch >/< docs-property >/< user >/DOP-XXXX/`

README.md
```diff
@@ -51,12 +60,33 @@ $ npm run build
...
+node --nolazy node_modules/.bin/gatsby develop --inspect-brk
+```
+
+and connect a node debugger (ie. chrome developer tools)
+
+### Testing with local manifest path (local AST zip files)
+
+Alternative to working with remote AST files, you can have a local zip file to build the site. This removes the need for previously mentioned env variables required for remote lookup `GATSBY_SITE` `GATSBY_PARSER_USER` and `GATSBY_PARSER_BRANCH`
+
+`.env.development`:
```
gjchong25, line 75: These env vars should also be specified for `.env.production` when using the manifest zip file option

gjchong25 (commented):
Thanks for doing this, I'm sure future new teammates will be very thankful for this helpful information :) Just left a few comments and suggestions!

README.md
seungpark: do you know exactly where this logic takes place? would be helpful to link

README.md
seungpark: done. linked it as an alternative within the same description

README.md
seungpark: tested and added :)

README.md
seungpark: found and linked :) thanks for the suggestion!

README.md
```diff
@@ -23,9 +32,13 @@ GATSBY_PARSER_BRANCH=<BRANCH>
 GATSBY_SNOOTY_DEV=true
 ```
 
-#### `.env.development`
+The `GATSBY_SNOOTY_DEV` variable is what allows Gatsby to know that when the application is built it should use the snooty branch name as part of the file paths. When not set, the file paths will use the value of `GATSBY_PARSER_BRANCH`.
```
mmeigs, line 35: This first sentence feels a teeny bit muddled. Maybe we can cut out "that when the application is built"? That would make it easier on first read, and that clause is probably easily inferred. Below is a possible version with some extraneous words cut out! 
"The `GATSBY_SNOOTY_DEV` variable allows Gatsby to know it should use the snooty branch name as part of the file paths."
or...
"Gatsby uses the `GATSBY_SNOOTY_DEV` variable to know if it should use the snooty branch name as part of the file paths."
Take it or leave it :)

README.md
```diff
@@ -34,29 +47,47 @@ GATSBY_PARSER_BRANCH=<BRANCH>
...
 $ npm run serve
 ```
 
+To debug server side:
+
+```shell
+node --nolazy node_modules/.bin/gatsby develop --inspect-brk
+```
+
+and connect a node debugger (ie. chrome developer tools)
```
mmeigs, line 70: Is there one you'd recommend? Could maybe link it. Even could expound on this if you think it's helpful. Just thinking as someone truly walking in on their first day. (not necessary though!)

README.md
```diff
@@ -34,29 +47,47 @@ GATSBY_PARSER_BRANCH=<BRANCH>
...
+
+### Running with local manifest path
+
+Alternative to working with remote AST files, you can have a local zip file to build the site. This removes the need for previously mentioned env variables required for remote lookup `GATSBY_SITE` `GATSBY_PARSER_USER` and `GATSBY_PARSER_BRANCH`
```
mmeigs, line 74: Should we mention how to build the local zip file?

README.md
```diff
@@ -34,29 +47,47 @@ GATSBY_PARSER_BRANCH=<BRANCH>
...
+
+```
+GATSBY_MANIFEST_PATH=/path/to/zipped/ast/file.zip
+GATSBY_SNOOTY_DEV=true
+```
+
 ## Staging
 
-Install [mut](https://github.com/mongodb/mut) and ensure that you have properly configured your Giza/AWS keys. Then, from root, run:
+Install libxml2 with `brew install libxml2` on mac and `apt-get install libxml2` on linux
```
mmeigs, line 85: Wow, cannot wait to do this! This is just an appreciation comment :)

mmeigs (commented):
Thanks for doing this. This is already so much clearer.

README.md
seungpark: i was thinking this should be omitted, since it touches on the parser repository as well.

`Local zip file is an output of the (parser)[link here]` ? how is that

README.md
seungpark: thanks! :D

README.md
seungpark: done! linked a guide to debugging node

README.md
seungpark: took it and left it. right, i omitted that, and linked to the logic of generating path prefixes

README.md
mmeigs: Ah! Makes sense. Yeah, sounds good

gjchong25 (approved):
LGTM, not sure if you want to wait for others to take a look but this looks fine to me!

PR #672, opened 2022-08-12
DOP-2779: fix scrolling issue for footnotes

seungpark:
### Stories/Links:

[DOP-2779](https://jira.mongodb.org/browse/DOP-2779)
Visiting docs URL's with a fragment (#) to a footnote scrolls past the footnote

### Current Behavior:
https://www.mongodb.com/docs/drivers/pymongo/#footnote-ssl-4.0-issue

### Staging Links:
https://docs-mongodbcom-integration.corp.mongodb.com/master/drivers/seung.park/DOP-2779/pymongo/#footnote-ssl-4.0-issue

### Notes:
footnote references at the top of a table are not able to be scrolled into view due to [block formatting context ](https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Block_formatting_context) formed by [table container](https://github.com/mongodb/leafygreen-ui/blob/%40leafygreen-ui/table%402.0.2/packages/table/src/Table.tsx#L191). 
[example](https://docs-mongodbcom-integration.corp.mongodb.com/master/drivers/seung.park/DOP-2779/pymongo/#ref-ssl-4.0-issue-id1)

think the solution here is to create hidden elements on top of table to bring table into view (if the Target is within first two rows of table). WDYT?

### DIFF
--- a/src/components/Footnote/FootnoteReference.js
+++ b/src/components/Footnote/FootnoteReference.js
@@ -1,13 +1,12 @@
-/**
- * Component used to show a clickable reference to footnote on page
- * scrolls to referenced element by id property
- */
-
 import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
 import FootnoteContext from './footnote-context';
 import { getNestedValue } from '../../utils/get-nested-value';
 
+/**
+ * Component used to show a clickable reference to footnote on page
+ * scrolls to referenced element by id property
+ */
 const FootnoteReference = ({ nodeData: { id, refname } }) => {
   const { footnotes } = useContext(FootnoteContext);
 
--- a/src/theme/docsTheme.js
+++ b/src/theme/docsTheme.js
@@ -74,7 +74,7 @@ const header = {
   navbarMobileHeight: '56px',
   docsMobileMenuHeight: '52px',
   // used for scrolling elements into place, considering sticky header
-  scrollBuffer: '175px',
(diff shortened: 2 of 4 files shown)


src/theme/docsTheme.js
```diff
@@ -73,6 +73,8 @@ const header = {
   navbarHeight: '88px',
   navbarMobileHeight: '56px',
   docsMobileMenuHeight: '52px',
+  // used for scrolling elements into place, considering sticky header
+  scrollBuffer: '175px',
```
casthewiz, line 77: minor nitpick, nonblocking: it might make sense to call this `navbarScrollOffset` given the usage (and `Buffer` is a bit of a loaded term)

casthewiz (approved):
Just some non-blocking nitpicks, it broadly LGTM! 👍

src/components/Footnote/FootnoteReference.js
```diff
@@ -1,3 +1,8 @@
+/**
+ * Component used to show a clickable reference to footnote on page
+ * scrolls to referenced element by id property
+ */
```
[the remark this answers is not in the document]
seungpark: good call. easy to see in a small component module but can easily miss with nested component definitions

src/theme/docsTheme.js
seungpark: good call. Buffer types!

PR #676, opened 2022-08-24
DOP-3167: fixed terrifying multiplication of punctuation

mmeigs:
### Stories/Links:

[DOP-3167](https://jira.mongodb.org/browse/DOP-3167)

### Current Behavior:

Reverted in production at the moment, but dangling punctuation multiplies on rerenders.
- when the user navigates back and forth between pages the punctuation is added on again and again. 
Attaching picture.

### Staging Links:

[Realm page with a Link and dangling period in first admonition on the page](https://docs-mongodbcom-integration.corp.mongodb.com/master/realm/matt.meigs/DOP-3167/sdk/kotlin/install/android/)

### Notes:

First fix of DOP-3167 - (attaching dangling punctuation to preceding Links) - was plagued by unintended consequence of never-ending additions of the same punctuation. References to the same node causes this. 
Current fix creates a deep copy of the change so as not to influence further renders from the same data.

### DIFF
--- a/src/utils/append-trailing-punctuation.js
+++ b/src/utils/append-trailing-punctuation.js
@@ -20,20 +20,20 @@ export const appendTrailingPunctuation = (nodes) => {
 };
 
 // Make a copy of node with a deep copy of the new child node with added punctuation
-function deepCopyNodesChildren(node, additionalValue) {
-  if (!node.children || !node.children.length) return node;
-  const deepCopyChildren = [...node.children];
-  const lastChild = deepCopyChildren.pop();
+function deepCopyNodesChildren(refNode, additionalValue) {
+  if (!refNode.children || !refNode.children.length) return refNode;
+  const copyOfChildren = [...refNode.children];
+  const lastChild = copyOfChildren.pop();
 
   const deepCopyLastChild = {
     ...lastChild,
     value: lastChild.value + additionalValue,
   };
 
-  deepCopyChildren.push(deepCopyLastChild);
+  copyOfChildren.push(deepCopyLastChild);
   const copyOfNode = {
-    ...node,
-    children: deepCopyChildren,
+    ...refNode,
+    children: copyOfChildren,
   };
   return copyOfNode;
 }
--- a/src/utils/append-trailing-punctuation.js
+++ b/src/utils/append-trailing-punctuation.js
@@ -21,7 +21,7 @@ export const appendTrailingPunctuation = (nodes) => {
 
 // Make a copy of node with a deep copy of the new child node with added punctuation
 function deepCopyNodesChildren(refNode, additionalValue) {
-  if (!refNode.children || !refNode.children.length) return refNode;
+  if (!refNode.children || !refNode.children.length) return { ...refNode };
   const copyOfChildren = [...refNode.children];
   const lastChild = copyOfChildren.pop();
 
--- a/src/utils/append-trailing-punctuation.js
+++ b/src/utils/append-trailing-punctuation.js
@@ -1,0 +1,39 @@
+// Append dangling punctuation to preceding Link's text
+export const appendTrailingPunctuation = (nodes) => {
+  const truncatedNodes = [];
+
+  for (let i = 0; i < nodes.length; i++) {
+    const currNode = nodes[i];
+    const nextNode = nodes[i + 1];
+    const isReference = currNode.type === 'reference';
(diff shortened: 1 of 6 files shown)


seungpark:
@mmeigs there is a merge conflict here from branding branch merge

seungpark:
@mmeigs the merge seems to have removed the call for appendTrailingPunctuation.

@schmalliso @mmeigs and i wondering if this is the right approach.. we are fixing this only for paragraphs that contain links, i see that we do have some other places ie. https://www.mongodb.com/docs/atlas/import/ where this is not in paragraph format. ideally, we would target references specifically, but they are already separated by parser into sibling nodes [{type: ref}, {type: text} ]. my 2 cents would be to update the parser to include a singular trailing '.' text in the reference node's children

src/utils/append-trailing-punctuation.js
```diff
@@ -0,0 +1,39 @@
...
+      continue;
+    }
+    truncatedNodes.push(currNode);
+  }
+  return truncatedNodes;
+};
+
+// Make a copy of node with a deep copy of the new child node with added punctuation
+function deepCopyNodesChildren(node, additionalValue) {
+  if (!node.children || !node.children.length) return node;
```
seungpark, line 24: minor. we should return {...node} here for copy of node

src/utils/append-trailing-punctuation.js
```diff
@@ -0,0 +1,39 @@
...
+      i++;
+      continue;
+    }
+    truncatedNodes.push(currNode);
+  }
+  return truncatedNodes;
+};
+
+// Make a copy of node with a deep copy of the new child node with added punctuation
+function deepCopyNodesChildren(node, additionalValue) {
```
seungpark, line 23: minor, rename param to refNode

src/utils/append-trailing-punctuation.js
```diff
@@ -0,0 +1,39 @@
+// Append dangling punctuation to preceding Link's text
+export const appendTrailingPunctuation = (nodes) => {
+  const truncatedNodes = [];
+
+  for (let i = 0; i < nodes.length; i++) {
+    const currNode = nodes[i];
+    const nextNode = nodes[i + 1];
+    const isReference = currNode.type === 'reference' || currNode.type === 'ref_role';
+    const hasDanglingSibling = nextNode?.type === 'text' && nextNode.value?.length === 1 && nextNode.value !== '\n';
```
seungpark, line 9: redundant check for newline since we are checking for length 1

src/utils/append-trailing-punctuation.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+    truncatedNodes.push(currNode);
+  }
+  return truncatedNodes;
+};
+
+// Make a copy of node with a deep copy of the new child node with added punctuation
+function deepCopyNodesChildren(node, additionalValue) {
+  if (!node.children || !node.children.length) return node;
+  const deepCopyChildren = [...node.children];
```
seungpark, line 25: the usage of 'deep' in this function name/var names are misleading since spread operators are not deep copies. https://medium.com/@kevinlai76/the-spread-operator-deep-and-shallow-copies-d193ac9b58bf

we can use JSON.parse(JSON.stringify(node)) if we really want deep copy, but this will be expensive in memory. we aren't mutating any nested attributes so the logic is good, but the 'deep' naming should be removed from vars

seungpark (approved):
minor comments below, but overall LGTM

src/utils/append-trailing-punctuation.js
mmeigs: Actually, newline wildly slips through! One of our tests caught this which I was very happy to see

src/utils/append-trailing-punctuation.js
mmeigs: I can do that, but I think it might be unnecessary. This is an escape hatch for edge cases where we don't change anything about the node, so no copy is needed just the original?

src/utils/append-trailing-punctuation.js
mmeigs: Newline is randomly stored as one character. Learned that today!

src/utils/append-trailing-punctuation.js
seungpark: thats fine, i just wanted to make sure naming is correct (ie. returning a deep copy)

PR #677, opened 2022-08-24
Update master from branding release

casthewiz:
Updates `master` with the latest contents of v0.13.24-branding.


.github/workflows/test.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     - uses: actions/checkout@v1
     - uses: actions/setup-node@v1
       with:
-        node-version: ${{ matrix.node-version }}
+        node-version: v14.16.0
```
seungpark, line 24: do we need this change? not sure what the intention here was

seungpark (approved):
went through commit history and LGTM. minor comment on the hardcoded node version but updated tests and mocked fn looks good 👍

.github/workflows/test.yml
casthewiz: there was a brief period of time yesterday where despite having the mock in place, the dynamic classnames were still present in `ci` - this was one of the changes I made to troubleshoot, since it was setting us up with Node 16.* instead of 14.16, which is what we run with locally and in other infrastructure

I think I'm going to keep it in this changeset, there's value in having the CI's node version pinned to the same version we use locally and in the infra

seungpark: oh it seems like node-version was not defined in the matrix and therefore default to latest LTS version (?) but LGTM 👍

PR #679, opened 2022-08-24
DOP-3193: add pseudo elements at top of table for scroll view

seungpark:
### Stories/Links:

[DOP-3193](https://jira.mongodb.org/browse/DOP-3193)
**issue:** footnote references within tables are not scrolled into view due [to overflow being hidden on LG/tables](https://github.com/mongodb/leafygreen-ui/blob/main/packages/table/src/Table.tsx#L44)
**solution:** add pseudo elements with footnote reference id's above table element

### Current Behavior:

[footnote reference in table (drivers)](https://www.mongodb.com/docs/drivers/pymongo/#ref-ssl-4.0-issue-id1)

### Staging Links:

[footnote reference in table (drivers)](https://docs-mongodbcom-integration.corp.mongodb.com/master/drivers/seung.park/DOP-3193/pymongo/#ref-ssl-4.0-issue-id1)

### Notes:
- caveat: this creates two elements with same id (one in table, one above table). browsers will look for the first match when scrolling via URL fragment identifier (ie. example.com/#fragment-identifer)


mmeigs (approved):
LGTM :)

PR #686, opened 2022-08-30
DOP-2592: update language switcher

seungpark:
### Stories/Links:

DOP-2592

### Current Behavior:

[server docs language switcher](https://www.mongodb.com/docs/manual/tutorial/insert-documents/)
[cloud docs](https://www.mongodb.com/docs/atlas/atlas-search/tutorial/autocomplete-tutorial/)
[tutorial docs](https://www.mongodb.com/docs/guides/crud/install/)

### Staging Links:

[server docs language switcher](https://docs-mongodbcom-integration.corp.mongodb.com/342468643cc6acc179effaebf2570854edd20aa1/master/docs/seung.park/DOP-2592/tutorial/insert-documents/) (includes version dropdown)
[cloud docs](https://docs-mongodbcom-integration.corp.mongodb.com/master/cloud-docs/seung.park/DOP-2592/atlas-search/tutorial/autocomplete-tutorial/)
[tutorial docs](https://docs-mongodbcom-integration.corp.mongodb.com/3bcc1cba5dbf186c0f4ae8d97970dd3e73d13e6e/master/guides/seung.park/DOP-2592/crud/install/)

[deprecated version selector](https://docs-mongodbcom-integration.corp.mongodb.com/3bcc1cba5dbf186c0f4ae8d97970dd3e73d13e6e/master/cloud-docs/seung.park/DOP-2592/sandbox/)
[search filters on the right](https://docs-mongodbcom-integration.corp.mongodb.com/e5a265c188cad189bb9c438c4a22e98fb4bf302e/master/landing/seung.park/DOP-2592-search-sample/search/)
[minor contents list on the right (only uses labels from select) ](https://docs-mongodbcom-integration.corp.mongodb.com/master/guides/seung.park/DOP-2592/)

### Notes:
@crisatmongo i updated the select component in this PR and seems to be working well for the language switcher, should verify it isn't breaking any other dependencies

- UPDATE: DeprecatedVersionDropdown, TabSelectors, SearchFilters, ContentsList, VersionDropdown have been verified and included in staging links. cc @allisonmui9 
- TODO: update failing unit tests


casthewiz (commented):
If possible, let's change the `select.css` file to be a `module` with the name `select.module.css` - that way the styles will be strictly scoped to wherever they're imported 👍 

I have a historical example of how it can be imported [here](https://github.com/mongodb/snooty/blob/4645757e36674ff63cff24d403b7f4b1106998d4/src/components/Sidebar.js#L7)

seungpark:
> If possible, let's change the `select.css` file to be a `module` with the name `select.module.css` - that way the styles will be strictly scoped to wherever they're imported 👍
> 
> I have a historical example of how it can be imported [here](https://github.com/mongodb/snooty/blob/4645757e36674ff63cff24d403b7f4b1106998d4/src/components/Sidebar.js#L7)

one difference is that we are pulling classnames and applying directly on components, whereas this stylesheet is targeting certain elements within LG components, not accessible in snooty components. not sure if this applies here, vs pulling in style rules per custom HTMLelement

casthewiz:
> > If possible, let's change the `select.css` file to be a `module` with the name `select.module.css` - that way the styles will be strictly scoped to wherever they're imported 👍
> > I have a historical example of how it can be imported [here](https://github.com/mongodb/snooty/blob/4645757e36674ff63cff24d403b7f4b1106998d4/src/components/Sidebar.js#L7)
> 
> one difference is that we are pulling classnames and applying directly on components, whereas this stylesheet is targeting certain elements within LG components, not accessible in snooty components. not sure if this applies here, vs pulling in style rules per custom HTMLelement

Sure, so we have 
` <LGSelect className="lg-select" (etc...)>`
and
`<Option className="option-item" key={choice.value} value={choice.value} role="option">`

I think we whould be able to use ```${`style['lg-select']`}``` and ```${`style['option-item']`}``` instead of having the className specified after making the `styles.css` into a module. Functionally, it would be equivalent - webpack should take care of everything.

But looking at this again, I'm not actually sure we need to keep the styles in a separate sheet. I think using the `cx` function out of emotion should allow us to define these styles inside of a ``` css`etc....` ``` tag, and then use `cx` in lieu of `option-item` and `lg-select`. There's an example in the [Admonition](https://github.com/mongodb/snooty/blob/master/src/components/Admonition.js#L48) component of a similar pattern.

seungpark:
> 

ah gotcha. i was resorting to classNames props here because `css`prop was not being passed down, but i will use the mix of cx to generate classnames and keep it scoped to the component without a stylesheet. on that note, i am seeing a mix use of emotion imports, ie. leafygreen/emotion, emotion/react. is there a best practice here?

rayangler:
> i am seeing a mix use of emotion imports, ie. leafygreen/emotion, emotion/react. is there a best practice here?

I don't think we really have a best practice here since `@leafygreen/emotion` is just a fork of `@emotion/react` and they pretty much do the same thing. Based on my experience though, I tend to use `@leafygreen/emotion` when styling LG components because using base/styled `emotion` on LG components could sometimes lead to a delay in the custom styles applying when the components are statically built.

src/components/Select.js
```diff
@@ -34,202 +14,87 @@ const Label = styled('p')`
...
-  const toggleSelectExpand = useCallback(() => {
-    if (enabled) setShowOptions(!showOptions);
-  }, [enabled, showOptions]);
-  /**
-   * This useEffect should only be called once the component first renders with choices,
-   * this should populate the select item with the default choice if there is one.
-   * If no choice is specified, "reset" the form.
-   */
+const PortalContainer = forwardRef(({ ...props }, ref) => (
+  <div className={props.className + ' portal-container'} ref={ref} css={portalStyle}>
```
rayangler, line 44: Is `portal-container` being used anywhere?

rayangler, line 44: Did you mean to include `portalStyle` in the className?

```
className={cx(portalStyle, props.className)}
```

I don't think LeafyGreen's `css` is directly applicable as a `css` prop

seungpark: removed. also removed css prop and moved to className and that should take care of the dropdown on upper left corner on search filters (from above comment)

> Also seeing the dropdown open from the upper left corner on the [search filters](https://docs-mongodbcom-integration.corp.mongodb.com/master/landing/seung.park/DOP-2592-search-sample/search/index.html?q=test).

seungpark: ha, yes removed this. combining the two styles

rayangler (approved):
LGTM!

PR #692, opened 2022-09-06
DOP-3227 - Add anchor tag support to tabs

carooobregon:
### Stories/Links:

DOP-3227

### Staging Links:

[Staging link opening to 'setup'](https://docs-mongodbcom-integration.corp.mongodb.com/master/landing/caro.obregon/tablink/#setup)

[Staging link opening to 'install'](https://docs-mongodbcom-integration.corp.mongodb.com/master/landing/caro.obregon/tablink/#install)

### Notes:
When opening a link with an anchor tag that points to a tab, it should open the page with the tab directly open.


src/components/Tabs/index.js
```diff
@@ -96,7 +96,12 @@ const Tabs = ({ nodeData: { children, options = {} }, page, ...rest }) => {
   const isProductLanding = page?.options?.template === 'product-landing';
 
   useEffect(() => {
-    if (!previousTabsetChoice || !tabIds.includes(previousTabsetChoice)) {
+    let anchorLinkTab = window.location.href.split('#')[1];
+
+    if (tabIds.includes(anchorLinkTab)) {
```
rayangler, line 101: Just wanted to flag a potential edge case where if there's a heading on a page with the same tab id as a heading, there might be some confusion as to which id the url is actually trying to reference. For example: A heading "Install" may have the same id "install" as a tab.

For linking purposes, I think we should have some sort of prefix to denote that a url wants to link to a tab and not a heading. Something like `#tab-install`?

src/components/Tabs/index.js
```diff
@@ -153,7 +158,7 @@ const Tabs = ({ nodeData: { children, options = {} }, page, ...rest }) => {
                 : tabId;
 
             return (
-              <LeafyTab className={cx(getTabStyling({ isProductLanding }))} key={tabId} name={tabTitle}>
+              <LeafyTab className={cx(getTabStyling({ isProductLanding }))} key={tabId} name={tabTitle} id={tabId}>
```
rayangler, line 161: Is there an example page/staging link where there are multiple of the same tabset on the same page? For example, a page that has tabs for different drivers in one section, and then the same set of tabs for the same drivers in another section. 

I'm curious what the interaction would be for something like this. I don't think it's proper to have 2 elements share the same `id` attribute on a page.

rayangler (commented):
Being able to denote which tab should be open when linking to a page is very cool! I do have a few additional concerns:

1) The href is useful for activating a tab, but it doesn't actually jump to where that tab is on the page. This might be a limitation of how tabs are rendering on the client side, but curious if there's a way to accomplish this.
2) Clicking on a tab doesn't automatically append the href to the url. If switching between two tabs and you want to share the link to a particular one, it's difficult to actually do so.

src/components/Tabs/index.js
```diff
@@ -96,7 +96,12 @@ const Tabs = ({ nodeData: { children, options = {} }, page, ...rest }) => {
   const isProductLanding = page?.options?.template === 'product-landing';
 
   useEffect(() => {
-    if (!previousTabsetChoice || !tabIds.includes(previousTabsetChoice)) {
+    let anchorLinkTab = window.location.href.split('#')[1];
```
mmeigs, line 99: nit-pick: I'm a fan of using const whenever possible!

src/components/Tabs/index.js
seungpark: https://www.mongodb.com/docs/atlas/configure-api-access/ here is an example

i believe the page scrolls to the first found id element. id sharing on the same page should be handled by the parser, as this can happen with any node of any type, WDYT? 

but for repeated tabsets, maybe we can generate a random id prefix from component and persist this for different id attribute prefixes. we can extend this logic to any components using id's

PR #693, opened 2022-09-12
DOP-3210: add version context 

seungpark:
### Stories/Links:

[DOP-3210](https://jira.mongodb.org/browse/DOP-3210)
This PR adds a version context that should be consumable by points of navigation (ie. sidenav, right column) for a consolidated ToC tree and content with embedded documentations

### Current Behavior:

[Atlas Docs](https://www.mongodb.com/docs/atlas/)
[landing](https://www.mongodb.com/docs/)

### Staging Links:

[staged Atlas Docs](https://docs-mongodbcom-integration.corp.mongodb.com/master/cloud-docs/seung.park/DOP-3210/)
[staged landing](https://docs-mongodbcom-integration.corp.mongodb.com/master/landing/seung.park/DOP-3210/)

### Notes:
There are no visible behavior changes in front end.
However, you should be able to observe network requests when first loading the page, and [debugging server side code](https://www.gatsbyjs.com/docs/debugging-the-build-process/#vs-code-debugger-auto-config) should also allow you to observe server side requests for realm app services.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -134,6 +134,10 @@ exports.createPages = async ({ actions }) => {
     await Promise.all(
       siteMetadata.associatedProducts.map(async (product) => {
         associatedReposInfo[product.name] = await db.stitchInterface.fetchRepoBranches(product.name);
+        // filter all branches of associated repo by associated versions only
+        associatedReposInfo[product.name].branches = associatedReposInfo[product.name].branches.filter((branch) => {
+          return product.versions.includes(branch.gitBranchName);
+        });
       })
     );
     let errMsg;
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -131,15 +131,17 @@ exports.createPages = async ({ actions }) => {
   try {
     const repoInfo = await db.stitchInterface.fetchRepoBranches();
 
-    await Promise.all(
-      siteMetadata.associatedProducts.map(async (product) => {
-        associatedReposInfo[product.name] = await db.stitchInterface.fetchRepoBranches(product.name);
-        // filter all branches of associated repo by associated versions only
-        associatedReposInfo[product.name].branches = associatedReposInfo[product.name].branches.filter((branch) => {
-          return product.versions.includes(branch.gitBranchName);
-        });
-      })
-    );
+    if (process.env.GATSBY_TEST_EMBED_VERSIONS) {
+      await Promise.all(
+        siteMetadata.associatedProducts.map(async (product) => {
+          associatedReposInfo[product.name] = await db.stitchInterface.fetchRepoBranches(product.name);
+          // filter all branches of associated repo by associated versions only
+          associatedReposInfo[product.name].branches = associatedReposInfo[product.name].branches.filter((branch) => {
+            return product.versions?.includes(branch.gitBranchName);
+          });
+        })
+      );
+    }
     let errMsg;
 
     if (!repoInfo) {
@@ -231,5 +233,15 @@ exports.createSchemaCustomization = ({ actions }) => {
     type SnootyMetadata implements Node @dontInfer {
         metadata: JSON!
     }
+
+    type AssociatedProduct {
+      name: String,
+      versions: [String]
+    }
+
+    type SiteSiteMetadata implements Node {
+      associatedProducts: [AssociatedProduct],
+    }
+
   `);
 };
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -127,8 +127,15 @@ exports.createPages = async ({ actions }) => {
   const { createPage } = actions;
 
   let repoBranches = null;
+  const associatedReposInfo = {};
   try {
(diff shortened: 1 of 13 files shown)


gatsby-node.js
```diff
@@ -127,8 +127,19 @@ exports.createPages = async ({ actions }) => {
...
+
+    await Promise.all(
+      siteMetadata.associatedProducts.map(async (product) => {
+        associatedReposInfo[product.name] = await db.stitchInterface.fetchRepoBranches(product.name);
+        // filter all branches of associated repo by associated versions only
+        associatedReposInfo[product.name].branches = associatedReposInfo[product.name].branches.filter((branch) => {
+          return product.versions.includes(branch.gitBranchName);
+        });
+      })
+    );
```
casthewiz, line 142: I think we should use a feature flag here, and consolidate `const associatedReposInfo = {}` to be under that flag. something like ` if (process.env.GATSBY_FEATURE_FLAG_EMBED_VERSIONS)` as a wrapper to control whether or not it's executed.

src/utils/realm.js
```diff
@@ -38,3 +38,11 @@ export const fetchSearchPropertyMapping = async (snootyEnv) => {
...
   return await fetchData('fetchOASFile', apiName, database);
 };
+
+export const fetchDocument = async (database, collectionName, query) => {
+  return await fetchData('fetchDocument', database, collectionName, query);
+};
+
+export const fetchDocuments = async (database, collectionName, query) => {
+  return await fetchData('fetchDocuments', database, collectionName, query);
+};
```
casthewiz, line 48: nit: we don't need to [return await](https://eslint.org/docs/latest/rules/no-return-await) in this circumstance, but the rest of the file has the same antipattern. I've dropped https://jira.mongodb.org/browse/DOP-3250 in to accommodate.

src/utils/setup/fetch-manifest-metadata.js
```diff
@@ -10,6 +10,14 @@ const fetchManifestMetadata = () => {
...
       if (entry.entryName === 'site.bson') {
         metadata = BSON.deserialize(entry.getData());
+
+        // TODO DOP-3210: remove test code. will be provided in metadata
+        metadata['associated_products'] = [
+          {
+            name: 'atlas-cli',
+            versions: ['v1.1', 'v1.0'],
+          },
+        ];
```
casthewiz, line 20: I think this another spot where we can add a feature flag, so that we don't pollute production builds with test data (minimal risk in this case, mostly the principle of the thing) 😄

casthewiz (commented):
This looks good on my first pass through, but I think we should use a feature flag to gate turning the mock data on/off 😄

src/utils/realm.js
seungpark: assigned issue to myself at top of backlog

gatsby-node.js
seungpark: good point on env var for feature flags. i set siteMetdata.associatedProducts [default value ](https://github.com/mongodb/snooty/pull/693/files#diff-bd8a11c708463818cc23b0e4f88ad099a9cee699d3d33e5a285a7c7b849c4418)to empty list so this should be good to merge

src/utils/setup/fetch-manifest-metadata.js
seungpark: done!

gatsby-node.js
casthewiz, line 142: I think this still needs to be covered by the feature flag.

casthewiz (commented):
We'll need to change the feature flag to have the prefix of `GATSBY_` for it to be present client-side at time of render.

There's a document [here](https://wiki.corp.mongodb.com/display/DE/Setting+Environment+Variables) with some info + context in general about environment variables throughout the whole system, if you're curious 😄

gatsby-node.js
seungpark: getting an error for graphql `Cannot query field "associatedProducts" on type "SiteSiteMetadata".` when setting it as an empty list (for default value when not testing with feature flag = false)

looking into defining optional schema for nodes exported via gatsby config...

seungpark: wrapped this with the feature flag. and addressed above comment by defining optional field for associatedProducts for SiteSiteMetadata schema

src/context/version-context.js
```diff
@@ -0,0 +1,110 @@
...
+  return branches[0]?.name || null;
+}
+
+function getInitVersions(branchListByProduct) {
+  const initState = {};
+  for (const productName in branchListByProduct) {
+    initState[productName] = getInitBranchName(branchListByProduct[productName]);
+  }
+  return initState;
+}
```
casthewiz, line 24: nitpick, nonblocking: Do we want these to be `const` declarations instead of `function`? We use arrow functions in the rest of the file - I would be pro-standardizing in this file to arrow functions.

casthewiz (approved):
This looks great, and I love the test coverage! 🙏

PR #702, opened 2022-09-27
DOP-3225: add version context

seungpark:
### Stories/Links:

[DOP-3225](https://jira.mongodb.org/browse/DOP-3225)

### Current Behavior:

[atlas docs](https://www.mongodb.com/docs/atlas/)

### Staging Links:

[atlas docs staged](https://docs-mongodbcom-integration.corp.mongodb.com/master/cloud-docs/seung.park/DOP-3225/getting-started/)

### Notes:
- using [mocked data](https://github.com/mongodb/snooty/pull/702/files#diff-1594a8cd87a233c31de3b28bc4b7639b66cd33e42ea4b28ae15a3800c8f129cc) for manifest metadata, as well as document uploaded to snooty_dev.metadata. filter by {page_id: /3225} to check out the document
- staged link has GATSBY_TEST_EMBED_VERSIONS environment variable set to true, and is reading from snooty_dev (see [changes](https://github.com/mongodb/snooty/pull/702/files#diff-67fd2be78a4e3484aedf9be9ed69bf5d5a356fa2a503d987deac14a4d2becfbeR64) )
- can optimize useEffect calls to use useMemo instead, to return a memoized version of state

### DIFF
--- a/src/context/toc-context.js
+++ b/src/context/toc-context.js
@@ -64,6 +64,8 @@ const getTocMetadata = async (db, project, parserUser, parserBranch, manifestTre
       page_id: `${project}/${parserUser}/${parserBranch}`,
     };
     if (process.env.GATSBY_TEST_EMBED_VERSIONS && project === 'cloud-docs') {
+      // TODO: remove testing code
+      db = 'snooty_dev';
       filter = {
         page_id: 'cloud-docs/spark/DOP-3225',
       };
--- a/src/context/toc-context.js
+++ b/src/context/toc-context.js
@@ -4,55 +4,54 @@ import { VersionContext } from './version-context';
 
 const TocContext = createContext({
   activeToc: {}, // table of contents
+  setActiveToc: () => {},
 });
 
+/**
+ *
+ * @param {object} tocTree toc tree head node
+ * @param {object} currentVersion see version-context currentVersion{}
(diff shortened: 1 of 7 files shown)


src/context/toc-context.js
```diff
@@ -0,0 +1,101 @@
...
+const TocContext = createContext({
+  activeToc: {}, // table of contents, represented by head node
+});
+
+/**
+ *
+ * @param {object} tocTree toc tree head node
+ * @param {object} currentVersion see version-context currentVersion{}
+ * @param {object} availableVersions see version-context availableVersions{}
+ * @modifies [tocTree.children[], tocTree.children[].options.versions[], tocTree.children[].options.versions]
```
seungpark, line 17: @casthewiz left a note for mutations here. this has possible implications of editing the graphql metadata resultant from manifest load.  would be glad to do a deep copy if this is preferred

casthewiz (approved):
This is fantastic 👀 🥳, the toc-test-data.json file should make things easier moving forwards
I really like the test suite coverage here also!

src/context/toc-context.js
casthewiz: I think shallow copy is going to be sufficient for now, but we can revisit with the UI level tickets if necessary 🤔

seungpark: sounds good!

PR #710, opened 2022-10-11
DOP-3266

seungpark:
### Stories/Links:

[DOP-3266](https://jira.mongodb.org/browse/DOP-3266)

### Current Behavior:

[atlas cli](https://www.mongodb.com/docs/atlas/cli/upcoming/install-atlas-cli/)

### Staging Links:

[staged atlas cli](https://docs-mongodbcom-integration.corp.mongodb.com/master/atlas-cli/seung.park/DOP-3266/install-atlas-cli/)

### Notes:
- `isAssociatedProduct` page context is initialized at build time, and failed calls will be caught and still go through build, but will not be considered an associated product - **should this fail the build instead?**
- [conversion to slug from git branch](https://github.com/mongodb/snooty/pull/710/files#diff-545362051c3eed6fa8804af3c9ca76102e0ccbafc366579d47c2142e90c956e6R36) comes largely from version dropdown
- [factored out memoized value for current url](https://github.com/mongodb/snooty/pull/710/files#diff-35671def5b99406749c4e33fdac790ecc876f38f151d9d3ac3839e8547ddd213) from version dropdown to share with associated version dropdown

this PR is dependent on two mocked documents in `snooty_dev.metadata`:
image

### TODO:
- unit tests for net new component


src/components/AssociatedVersionSelector/index.js
```diff
@@ -0,0 +1,149 @@
...
+      if (branch.gitBranchName === parserBranch) {
+        return setOptionSlug(branch);
+      }
+    }
+    return parserBranch;
+  }, [parserBranch, availableVersions, project]);
+
+  // refer to src/components/VersionDropdown/index.js
+  const getUrl = useCallback(
+    (versionTarget) => {
```
rayangler, line 44: Just curious if it would be worth making this into a reusable function that both the regular version selector and the associated version selector can use. I wasn't sure if there was a reason why it was separated

src/components/AssociatedVersionSelector/index.js
```diff
@@ -0,0 +1,149 @@
...
+      updatedVersion[project] = value;
+      setActiveVersions(updatedVersion);
+
+      const targetBranch = findBranchByGit(value);
+      if (!targetBranch) {
+        console.error(`target branch not found for git branch <${value}>`);
+        return;
+      }
+
+      const target = targetBranch.alias || targetBranch.urlAliases[0] || targetBranch.gitBranchName;
```
rayangler, line 91: Should we prioritize a target branch's `urlSlug`? This is what the version selector currently does.

Also, branches in our data don't have an `alias` field. Or is that coming from somewhere else?

src/components/AssociatedVersionSelector/index.js
```diff
@@ -0,0 +1,149 @@
...
+import { normalizePath } from '../../utils/normalize-path';
+import { assertTrailingSlash } from '../../utils/assert-trailing-slash';
+import { baseUrl } from '../../utils/base-url';
+import { generatePrefix } from '../VersionDropdown/utils';
+
+const buildChoices = (branches) => {
+  return branches.map((branch) => ({
+    value: branch['gitBranchName'],
+    text: getUILabel(branch),
+    slug: branch['urlSlug'] || branch['gitBranchName'],
```
rayangler, line 17: Is this `slug` field being used anywhere?

rayangler (commented):
Looks very cool at the moment 😎

gatsby-node.js
```diff
@@ -141,6 +143,17 @@ exports.createPages = async ({ actions }) => {
...
+      // check if product is associated child product
+      try {
+        const umbrellaProduct = await db.stitchInterface.getMetadata({
+          'associated_products.name': siteMetadata.project,
+        });
+        isAssociatedProduct = !!umbrellaProduct;
+      } catch (e) {
+        console.log('No umbrella product found. Not an associated product.');
+        isAssociatedProduct = false;
+      }
```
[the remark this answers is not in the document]
seungpark: this particular scenario was due to fact that fetchDocument returns an error if the document is not found whereas fetchDocuments returns an empty list. i agree that this should fail on error, but the error'ing part is not clean. any objection to return undefined/null from the realm app service if no document in a fetchDocument function? @casthewiz

src/components/AssociatedVersionSelector/index.js
seungpark: could make a separate useCallback hook here. will separate out and re-use in both 👍

src/components/AssociatedVersionSelector/index.js
seungpark: good catch. this should be urlSlug

src/components/AssociatedVersionSelector/index.js
seungpark: good catch. removing. the intention was to not have conversion and just pull the slug from the option, but seems like callback for onSelect only gives you back 'value' field

gatsby-node.js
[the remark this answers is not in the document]
casthewiz: I think that this usage is actually fine, just a case of novel usage for some of these functions. 

Imo we want the build to succeed if there's intermittency, an App Services outage, or if there's a missing entry. In the hierarchy of system failure behavior we would generally prefer to build the content even if the IA used is the original entry provided at build time, and not the shared umbrella & associated entry. 

At the same time, we don't strictly want to change the error behavior of `fetchDocument`, since it's ostensibly used for other cases, and I think the error behavior should be preserved (we just want to handle this specific case, not all cases, in theory)

So I think any error within `fetchDocument` should be logged, and then behavior should be to fall back to the original, non-associated product. Logging the error will give us context, and falling back to originally provided build artifacts preserves writer staging + production deployment behavior, even if the IA will then be one isolated to only the product + version being deployed.

seungpark:
failing tests should be addressed by https://github.com/mongodb/snooty/pull/716

toc-context is throwing errors and it has been refactored in above PR

seungpark:
> failing tests should be addressed by #716
> 
> toc-context is throwing errors and it has been refactored in above PR

above PR is merged to https://github.com/mongodb/snooty/pull/714
DOP-3267 can come before this PR

rayangler (approved):
LGTM. Very cool

PR #719, opened 2022-11-02
DOP-3350: Create mapping of OpenAPI spec stores

rayangler:
### Stories/Links:

DOP-3350

### Current Behavior:

[Atlas App Services API](https://www.mongodb.com/docs/atlas/app-services/admin/api/v3/)
[Atlas Admin API](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/)

### Staging Links:

(Since this is on the `master` branch, the old version and branding of Redoc will be used.)

[Atlas App Services API via local file](https://docs-mongodbcom-integration.corp.mongodb.com/master/atlas-app-services/raymundrodriguez/DOP-3350/admin/api/v3/index.html) (might include multiple specs on the same page due to how the rST was written, but the first one is the one to focus on)
[Atlas Admin API via URL](https://docs-mongodbcom-integration.corp.mongodb.com/master/atlas-app-services/raymundrodriguez/DOP-3350/admin/api/v4/index.html) (the site and URL intentionally do not match what's in production for the purposes of local testing)

### Notes:
* This PR is the first step to generating OpenAPI content pages through the frontend (rather than through the autobuilder, as originally planned). DOP-3352 will handle integrating a different part of the Redoc library to consume the context data introduced in this PR.
* Uses the metadata for OpenAPI content pages to create a mapping of Redoc spec stores for the corresponding pages.
* For convenience, the staging links include a console log of the `openapiSpecStore` context object passed to the OpenAPI content pages. If you navigate to other pages, the console logs should not be there.


gatsby-node.js
```diff
@@ -195,23 +196,40 @@ exports.createPages = async ({ actions }) => {
...
+
+        if (openapiPageMetadata[slug]) {
+          const openapiSpecStore = openapiPageMapping[slug];
+          if (!!openapiSpecStore) {
+            context['openapiSpecStore'] = openapiSpecStore;
+          } else {
+            // Skip content page and log error if an OpenAPI content page was
+            // unsuccessful sourcing its spec content.
+            console.error(`OpenAPI content page "${slug}" detected, but no spec store found.`);
+            return;
```
rayangler, line 225: Most likely to be unsuccessful for `cloud-docs`/Atlas specs that require pulling the latest version via git hash. Sometimes, the git hash returned doesn't always have an existing OpenAPI spec, leading to a fetch error.

For now, we just skip building the page if there's a mismatch between an expected OpenAPI content page and what spec stores were created. Happy to change this if there's a stronger preference for the entire build failing. (I don't remember `mut`'s logic as of writing this, but I'm assuming that the stale/outdated page will still exist, even if skipped in the current build)

src/utils/setup/construct-openapi-page-mapping.js
```diff
@@ -0,0 +1,67 @@
...
+  // Currently, the only expected API fetched programmatically is the Cloud Admin API,
+  // but it's possible to have more in the future with varying processes.
+  if (apiKeyword !== 'cloud') {
+    return null;
+  }
+
+  const versionURL = 'https://cloud.mongodb.com/version';
+  const gitHash = await fetchData(versionURL, 'Could not find current version or git hash.');
+  const oasFileURL = `https://mongodb-mms-prod-build-server.s3.amazonaws.com/openapi/${gitHash}.json`;
+  const oasFileContent = await fetchData(oasFileURL, `Unable to find spec from ${oasFileURL}.`);
```
rayangler, line 37: If the spec file doesn't exist in Atlas's S3 bucket for some reason, we could potentially query the last known successful git hash version stored in our db instead of throwing an error. (This is saved by our scheduled function on Atlas App Services.)

No decision needed now, but figured it was a thought worth sharing as an option to implement at a later date, if needed.

rayangler (commented):
Annotations

src/utils/setup/construct-openapi-page-mapping.js
```diff
@@ -0,0 +1,67 @@
...
+ * @returns Serialized Redoc spec store specific to Atlas APIs
+ */
+const getAtlasSourceStore = async (apiKeyword) => {
+  // Currently, the only expected API fetched programmatically is the Cloud Admin API,
+  // but it's possible to have more in the future with varying processes.
+  if (apiKeyword !== 'cloud') {
+    return null;
+  }
+
+  const versionURL = 'https://cloud.mongodb.com/version';
```
seungpark, line 34: where is this url connecting to? out of curiosity

gatsby-node.js
seungpark: if this log will be captured, can we fall back to loading the page without the specs? otherwise might be safer to kill than a silent failure? my 2 cents

src/utils/setup/construct-openapi-page-mapping.js
```diff
@@ -0,0 +1,67 @@
...
+  const mapping = {};
+  const entries = Object.entries(openapiMetadata);
+
+  for (const [slug, data] of entries) {
+    const { source_type: sourceType, source } = data;
+
+    try {
+      if (sourceType === 'url' || sourceType === 'local') {
+        mapping[slug] = await createSerializedStore(source);
+      } else if (sourceType === 'atlas') {
```
seungpark, line 56: is there specific env vars to get it to read from atlas. or url? thank you!

seungpark (changes requested):
the new util works great and the unit test is 👍  but a bit fuzzy on the remote openapi spec (the source and the validity). also not sure about sending a 404 page on fail. left comments on these places. also, is the consumption of `createStore.toJS()` TBD ?

gatsby-node.js
rayangler: Rearranged [try-catch blocks](https://github.com/mongodb/snooty/pull/719/commits/40fc57d53631975ca2d4b3bfac5ce1d8c9a33453#diff-a6f3904c41211be170d9b0eb87e49ea8d02d0fd5a7224d6772174b425dacf90fR202-R209) to avoid silent failure for OpenAPI content pages

src/utils/setup/construct-openapi-page-mapping.js
rayangler: Discussed offline, but for posterity:

`https://cloud.mongodb.com/version` is an API endpoint maintained and provided by the Cloud team. It returns the latest production git hash available for Atlas.

src/utils/setup/construct-openapi-page-mapping.js
rayangler: Discussed offline, but for posterity:

These 3 source types are assigned by the parser based on the argument of the `openapi` directive in rST. See below for examples of these 3 source types we currently support

```
.. openapi:: /openapi-admin-v3.yaml

.. openapi:: https://mongodb-mms-prod-build-server.s3.amazonaws.com/openapi/d973547b203fc7ac605685953b40141dfbc17eea.json

.. openapi:: cloud
   :uses-realm:
```

rayangler:
> is the consumption of createStore.toJS() TBD ?

@seungpark yup, the serialized store will be consumed as part of DOP-3352.

src/utils/setup/construct-openapi-page-mapping.js
```diff
@@ -0,0 +1,63 @@
...
+};
+
+/**
+ * @param {string} apiKeyword
+ * @returns Serialized Redoc spec store specific to Atlas APIs
+ */
+const getAtlasSourceStore = async (apiKeyword) => {
+  // Currently, the only expected API fetched programmatically is the Cloud Admin API,
+  // but it's possible to have more in the future with varying processes.
+  if (apiKeyword !== 'cloud') {
```
seungpark, line 30: if this is the default spec, then we should either 1) throw here for `createPages` to fail or 2) log error and build with api openapi specs. here, we are just returning no spec

seungpark (approved):
this LGTM. tested with all three sourceTypes parsed locally and seeing the resulting specs in context when on right page 👍

src/utils/setup/construct-openapi-page-mapping.js
rayangler: Thanks for catching! I intended to throw an error there. Added the throw and added a small test for it, just in case.

PR #725, opened 2022-11-09
DOP-3362: include slug in url when changing versions

seungpark:
### Stories/Links:

DOP-3362

### Current Behavior:

[prod behavior](https://www.mongodb.com/docs/manual/core/data-modeling-introduction/)
flipping the version on the left side does not automatically take you to the **same page** on a different version of the doc

### Staging Links:

[staged changes](https://docs-mongodbcom-integration.corp.mongodb.com/master/docs/seung.park/DOP-3362/core/data-modeling-introduction/)
changing the version on the left side leads to a 404 (but this is desired) since url parsing is not working well between environments, but correct slug is being appended to the destination URL

### Notes:
- would be neat to clean up url parse to work interchangeably with environments, as a follow up to env cleanups

### DIFF
--- a/src/components/RootProvider.js
+++ b/src/components/RootProvider.js
@@ -12,6 +12,7 @@ const RootProvider = ({
   children,
   headingNodes,
   selectors,
+  slug,
   repoBranches,
   associatedReposInfo,
   isAssociatedProduct,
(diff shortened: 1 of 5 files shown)


casthewiz (approved):

rayangler (approved):
Makes sense. Thanks for the quick fix

PR #736, opened 2022-11-16
DOP-3330: Check for all whitespace characters when appending dangling character

mmeigs:
### Stories/Links:

[DOP-3330](https://jira.mongodb.org/browse/DOP-3330)

### Current Behavior:

No good link for now, but will attach screenshot of the behavior that was seen.
bold_no_space

### Staging Links:

[Staging](https://docs-mongodbcom-integration.corp.mongodb.com/master/atlas-app-services/matt.meigs/DOP-3330-space/billing/)

### Notes:

Content writers saw that spaces were being removed before `<b>` bold tags within a paragraph.

This happens to be a larger issue with whitespace being taken into account by the util append-trailing-punctuation

Now we account for all whitespace with regex. 

P.S. if we'd rather account FOR certain punctuation rather than AGAINST certain characters, we could do that by matching regex with characters that could potentially be trailing such as `.?!,:;)*"']}` and backticks
imo, just making sure we don't delete/attach whitespace seemed the most succinct way to handle any situation for now.


casthewiz (approved):

casthewiz:
> P.S. if we'd rather account FOR certain punctuation rather than AGAINST certain characters, we could do that by matching regex with characters that could potentially be trailing such as .?!,:;)*"']} and backticks

This would make a good `TODO:` comment 😉

PR #740, opened 2022-11-29
DOP-3327: Move entire dangling node in appendTrailingPunctuation() rather than just value

i80and:
This fixes an issue where a trailing period would be appended in to the value of a node type with no value field, effectively causing the period to be truncated.

### Stories/Links:

DOP-3327

### Current Behavior:

* https://www.mongodb.com/docs/atlas/cli/stable/

### Staging Links:

* https://docs-mongodbcom-integration.corp.mongodb.com/master/atlas-cli/heli/DOP-3327/

### DIFF
--- a/src/utils/append-trailing-punctuation.js
+++ b/src/utils/append-trailing-punctuation.js
@@ -26,13 +26,7 @@ export const appendTrailingPunctuation = (nodes) => {
 function deepCopyNodesChildren(refNode) {
   if (!refNode.children || !refNode.children.length) return { ...refNode };
   const copyOfChildren = [...refNode.children];
-  const lastChild = copyOfChildren.pop();
 
-  const deepCopyLastChild = {
-    ...lastChild,
-  };
-
-  copyOfChildren.push(deepCopyLastChild);
   const copyOfNode = {
     ...refNode,
     children: copyOfChildren,
--- a/src/utils/append-trailing-punctuation.js
+++ b/src/utils/append-trailing-punctuation.js
@@ -11,7 +11,8 @@ export const appendTrailingPunctuation = (nodes) => {
     // ie: regex match for  .?!,:;)*"']}`
 
     if (isReference && hasDanglingSibling) {
-      const copyOfNodeWithDeepCopiedChildren = deepCopyNodesChildren(currNode, nextNode.value);
+      const copyOfNodeWithDeepCopiedChildren = deepCopyNodesChildren(currNode);
+      copyOfNodeWithDeepCopiedChildren.children.push(nextNode);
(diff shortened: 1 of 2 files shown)


src/utils/append-trailing-punctuation.js
```diff
@@ -21,15 +22,14 @@ export const appendTrailingPunctuation = (nodes) => {
   return truncatedNodes;
 };
 
-// Make a copy of node with a deep copy of the new child node with added punctuation
-function deepCopyNodesChildren(refNode, additionalValue) {
+// Make a copy of node with a deep copy of the new child node
+function deepCopyNodesChildren(refNode) {
   if (!refNode.children || !refNode.children.length) return { ...refNode };
   const copyOfChildren = [...refNode.children];
   const lastChild = copyOfChildren.pop();
```
mmeigs, line 29: Correct me if I'm wrong, but I think now we can delete anything to do with the `lastChild`. This function can now just copy the children and put that inside a copy of the node, so delete lines 29-35? Thanks for cleaning this up!

mmeigs (approved):
LGTM! Thanks for finding this better way. Only nitpick is deleting one section of old code that I think is now unnecessary.

src/utils/append-trailing-punctuation.js
i80and: I mostly left this method alone because I wasn't 100% sure on the semantics, but I can go ahead and do that!

src/utils/append-trailing-punctuation.js
```diff
@@ -11,7 +11,8 @@ export const appendTrailingPunctuation = (nodes) => {
     // ie: regex match for  .?!,:;)*"']}`
 
     if (isReference && hasDanglingSibling) {
-      const copyOfNodeWithDeepCopiedChildren = deepCopyNodesChildren(currNode, nextNode.value);
+      const copyOfNodeWithDeepCopiedChildren = deepCopyNodesChildren(currNode);
+      copyOfNodeWithDeepCopiedChildren.children.push(nextNode);
```
mmeigs, line 15: Oh! Glad you re-requested! Just realized there's an edge case here that hopefully will never happen, but who knows. In deepCopyNodeChildren we "could" return a node without a children element. So, however you want to handle that on line 15! Could use optional chaining or a conditional.

i80and: `deepCopyNodesChildren()` should only be passed a `reference` or `ref_role` node, both of which subclass from Parent, and should thus have `children`. If that contract is violated, I would rather we error immediately rather than create an illegal AST.

I could add an assertion if you'd like.

PR #741, opened 2022-11-29
DOP-3321: Fix broken 'Whats Next' Button

casthewiz:
### Stories/Links:

DOP-3321

### Current Behavior:

https://www.mongodb.com/docs/guides/atlas/account/

### Staging Links:

https://docs-mongodbcom-integration.corp.mongodb.com/master/guides/cassidy.schaufele/DOP-3321/atlas/account/

### Notes:

Effectively copies the Link fix that we performed for PLP's.
Making this fix more generic should be preferred, but in the interest of making this fixed in production, should be addressed separately.

### DIFF
--- a/src/components/GuideNext/Content.js
+++ b/src/components/GuideNext/Content.js
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
 import Button from '@leafygreen-ui/button';
-import { css, cx } from '@leafygreen-ui/emotion';
+import { css } from '@leafygreen-ui/emotion';
 import Link from '../Link';
 import { palette } from '@leafygreen-ui/palette';
 import ConditionalWrapper from '../ConditionalWrapper';
@@ -74,12 +74,10 @@ const Content = ({ argument, children, guideData }) => {
           as={Link}
           to={buttonUrl}
           variant="primary"
-          className={cx(
-            css`
-              color: #ffffff !important;
-            `,
-            'button'
-          )}
+          hideExternalIcon={true}
+          className={css`
+            color: #ffffff !important;
+          `}
         >
           {buttonText}
         </Button>
--- a/src/components/GuideNext/Content.js
+++ b/src/components/GuideNext/Content.js
@@ -2,6 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
 import Button from '@leafygreen-ui/button';
+import { css, cx } from '@leafygreen-ui/emotion';
+import Link from '../Link';
 import { palette } from '@leafygreen-ui/palette';
 import ConditionalWrapper from '../ConditionalWrapper';
 import { theme } from '../../theme/docsTheme';
@@ -68,7 +70,17 @@ const Content = ({ argument, children, guideData }) => {
       </ConditionalWrapper>
       {/* We only want to show the button if argument/children are empty */}
       {!hasCustomContent && buttonUrl && (
-        <Button href={buttonUrl} variant="primary">
+        <Button
+          as={Link}
+          to={buttonUrl}
+          variant="primary"
+          className={cx(
+            css`
+              color: #ffffff !important;
+            `,
+            'button'
+          )}
+        >
           {buttonText}
         </Button>
       )}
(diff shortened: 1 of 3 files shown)


tests/unit/GuideNext.test.js
```diff
@@ -60,4 +60,11 @@ describe('GuideNext', () => {
     // Default copy title should not be present
     expect(wrapper.queryAllByText('Become a MongoDB Professional')).toHaveLength(0);
   });
+
+  it('renders whats next with proper url"', () => {
+    const currentSlug = 'server/read';
+    const wrapper = renderGuideNext(currentSlug, mockNodeData.customContent);
+    expect(wrapper.getByText('Click This')).toBeTruthy();
+    expect(wrapper.queryAllByRole('link')[0]).toHaveAttribute('href', 'https://mongodb.com/');
```
casthewiz, line 68: I'm having difficulty isolating where we should be testing the 'What's Next' functionality and mocking the GuideData needed for the button to render as it does in production.

Adding a very barebones test here instead, though its utility/functionality is certainly suspect.

rayangler: If the intention is to render the next followup guide, using `mockNodeData.noContent` like in tests [1](https://github.com/mongodb/snooty/blob/7ff13ba3f651fad3dd6d52c19f8c65345ca83e1b/tests/unit/GuideNext.test.js#L34) and [2](https://github.com/mongodb/snooty/blob/7ff13ba3f651fad3dd6d52c19f8c65345ca83e1b/tests/unit/GuideNext.test.js#L41) should be closer to what you're looking for. If the directive has no content, then "What's Next" will be whatever the default copy is based on the current guide and chapter.

Based on the updated CSS for the snapshots of those tests, the `Link` component is successfully being used, but the resulting href/url will remain the same since we mock Gatsby's `Link` component along with its underlying implementation of `withPrefix`. For what it's worth, I think the CSS changes and staging links help verify that the intended functionality is being applied for those links

src/components/GuideNext/Content.js
```diff
@@ -68,7 +70,17 @@ const Content = ({ argument, children, guideData }) => {
       </ConditionalWrapper>
       {/* We only want to show the button if argument/children are empty */}
       {!hasCustomContent && buttonUrl && (
-        <Button href={buttonUrl} variant="primary">
+        <Button
+          as={Link}
```
rayangler, line 74: Using `Link` here causes the button to show the external icon on the final guide's [button](https://docs-mongodbcom-integration.corp.mongodb.com/master/guides/cassidy.schaufele/DOP-3321/crud/delete/index.html#see-also). Not sure what the intended styling here is, but if we want to maintain what's currently in production, we can pass in `hideExternalIcon={true}` to this `Button` component

rayangler (commented):
Things seem to work well! Just a few comments for now

src/components/GuideNext/Content.js
```diff
@@ -68,7 +70,17 @@ const Content = ({ argument, children, guideData }) => {
...
+          to={buttonUrl}
+          variant="primary"
+          className={cx(
+            css`
+              color: #ffffff !important;
+            `,
+            'button'
```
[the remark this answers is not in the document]
casthewiz: It does indeed look vestigial - going to see about removing it.

tests/unit/GuideNext.test.js
casthewiz: Thanks for giving me a little more context and direction here 😅 

I was hoping to have a test that would break if the url behavior changes here, but since we mock `Link` and `withPrefix`, I'm going to prefer shipping the fix over trying to perfect the test coverage.

src/components/GuideNext/Content.js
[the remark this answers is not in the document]
casthewiz: Safely removed!

src/components/GuideNext/Content.js
casthewiz: Safely added the prop!

casthewiz:
I have since re-staged with suggested changes 😄

tests/unit/GuideNext.test.js
rayangler: Makes sense. In the future, we could probably change our mock implementations of Gatsby components (either globally, or within specific files like this) to help flag these changes better. It is a bit difficult to differentiate between regular `a` tags and our mocked Gatsby Links

rayangler (approved):

PR #750, opened 2023-01-12
DOP-2872: imports order rule implemented in eslint

mmeigs:
### Stories/Links:

DOP-2872

### Current Behavior & Staging Links:

N/A

### Notes:

Enforces a convention in the order of import statements. 

eslint-plugin-import is being used to apply rules in eslintrc.js to specify import orderings. First run was successfully completed. This is currently set as errors but can easily be set to simple warnings. Orderings of groups can also be changed in the future if desired.


seungpark (approved):
looks clean! works with eslint plugin on vscode too 👍

rayangler (approved):
Setting it to error should be okay. We consider lint warnings as a failure when checking it anyways

PR #756, opened 2023-01-26
DOP-3367: Override LG Select mobile font-size styles

mmeigs:
### Stories/Links:

[DOP-3367](https://jira.mongodb.org/browse/DOP-3367)

### Current Behavior:

[Current](https://www.mongodb.com/docs/manual/tutorial/query-documents/)

### Staging Links:

[Staging Link with error message and description added](https://docs-mongodbcom-integration.corp.mongodb.com/master/docs/matt.meigs/DOP-3367-select-mobile/tutorial/query-documents/)

### Notes:

LG Select (along with all children) have mobile font-sizes of 16px. Our site's baseFontSize is 13px. This leads to all fonts in our Select to actually get larger as screensize transitions below 1024px.

This PR overrides the font-size for Select's label, description, error message, and options. 

This is necessary in two files. Snooty's Select and also VersionDropdown which pulls directly from LG's Select as well.

Changed Snooty's large breakpoint to 1024px from 1023px. This was leading to a one pixel gap wherein styles were incorrect. [Link to LG breakpoints](https://github.com/mongodb/leafygreen-ui/blob/main/packages/tokens/src/breakpoints.ts).


src/components/Tabs/TabSelectors.js
```diff
@@ -44,6 +45,16 @@ const TabSelector = ({ activeTab, handleClick, iconMapping, name, options }) =>
           /* Min width of right panel */
           max-width: 180px;
         }
+
+        /* Override LG mobile style of enlarged mobile font */
+        @media ${screenSize.upToLarge} {
```
seungpark, line 50: not sure if we want to only target TabSelectors.

seems like this also affects version selector in docs menu in mobile view.

would suggest using the `size` prop from [LG Select ](https://www.mongodb.design/component/select/documentation/)

tested this locally but seeing weird behavior -> `small` is good on mobile but large on desktop? `default` looks good on desktop but large on mobile? weird... we can use media query to determine this value though

seungpark (changes requested):
think we need to scope this broader than Tab Selectors here ie. our Select component

src/theme/docsTheme.js
```diff
@@ -40,7 +40,7 @@ const breakpoints = {
   xSmall: 320,
   small: 480,
   medium: 767,
-  large: 1023,
+  large: 1024,
```
mmeigs, line 43: I added a note for this in the PR description, but I realized LG breakpoints are one pixel off from ours. This closes that gap so we don't have awkward stylings at 1024px width screensize.

seungpark: good catch! tested with the breakpoint definitions below, ie. upToLarge

seungpark (approved):
minor. we can do the styling outside of the component fn and pass it to the cx fn. otherwise LGTM 👍

mmeigs:
> minor. we can do the styling outside of the component fn and pass it to the cx fn. otherwise LGTM 👍

I do think that looks cleaner and usually has a lot of benefits that way. But I don't know of a way to still utilize the `screenSize` property from `useTheme` outside the component unless:
-  I use `styled` instead of `css` but also prop-drill `screenSize` into `Select` just so that emotion can access it? 
- Or I could hard code the media query. 

But both of those seem to be less clean than the current solution.
Do you know of a different way to handle this?

seungpark:
> 

we can import theme from the module itself ([example](https://github.com/mongodb/snooty/blob/master/src/components/Card/index.js#L37-L41)) and write a media query for mobile sizes. when working with LG components (in this case LG select), we should try to follow their pattern and use their emotion package ie.

```
const styling = css`
  some styling here
`;
.
.
.
<Leafygreencomponent
     className={cx(styling)} />
```

seungpark (approved):
thanks for the update!

PR #758, opened 2023-01-27
DOP-3388-B: Writer preview build conditioned on string "production" rather than "prd"

mmeigs:
### Stories/Links:

Original ticket
[DOP-3388](https://jira.mongodb.org/browse/DOP-3388)

### Notes:

Preview build is based off of two conditions being true: the exact environment name(`siteMetadata.snootyEnv`) being equal to "prd" and the `PREVIEW_BUILD_ENABLED` flag uppercased being strictly equal to "TRUE". 

The `PREVIEW_BUILD_ENABLED` flag is correctly set in a gitPush "writer's build" job [as seen in within the ECS instance](https://us-east-2.console.aws.amazon.com/ecs/v2/task-definitions/docs-worker-pool-prd/30/json?region=us-east-2).

However, the environment designated for preview build (prd) sets the environment string passed to `siteMetadata.snootyEnv` from `SNOOTY_ENV` which is not equal to "prd" but rather set to "production" [here](https://github.com/mongodb/docs-worker-pool/blob/master/infrastructure/ecs-main/serverless.yml#L145-L150).

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -4,7 +4,8 @@ const { siteMetadata } = require('./src/utils/site-metadata');
 const pathPrefix = generatePathPrefix(siteMetadata);
 
 // TODO: Gatsby v4 will enable code splitting automatically. Delete duplicate components, add conditional for consistent-nav UnifiedNav in DefaultLayout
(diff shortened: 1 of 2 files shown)


seungpark (approved):
thanks for the quick fix!

PR #763, opened 2023-02-06
[DOP-3290]: Search the current drivers docs site by default

branberry:
### Stories/Links:

[DOP-3290](https://jira.mongodb.org/browse/DOP-3290)

### Current Behavior:

https://www.mongodb.com/docs/drivers/node/current/

### Staging Links:

https://docs-mongodbcom-integration.corp.mongodb.com/v4.11/node/brandonly/DOP-3920B/

### Notes:

To test the new functionality, try searching using the current link. It should show results for all docs. With the staging link, the search should only show results for node driver docs.

This change also requires an update to an Atlas App services function. The function can be found [here, fetchSearchPropertyMapping](https://realm.mongodb.com/groups/5bad1d3d96e82129f16c5df3/apps/5d41ef5e2de84772fc7c2de2/functions/6279548c1fdab5262a3460c8)


schmalliso:
Question! Is this applying only for drivers sites, or if I searched from https://www.mongodb.com/docs/atlas/ would we also default to search Atlas?

branberry:
Hi Allison! From my understanding, we only do this for driver sites. I could be mistaken of course

src/utils/drivers-set.js
```diff
@@ -0,0 +1,5 @@
+/**
+ * This set corresponds to the repo branch projects that have an entry
+ * in the search filter under category.
+ */
+export const driversSet = new Set(['golang', 'ruby', 'php', 'csharp', 'java', 'node']);
```
seungpark, line 5: extending @schmalliso 's question, can we fetch this async via https://docs-search-transport.mongodb.com/status and only include this from the response? as for staging we can append a result to the response ie `node-v5.0` (and get branch from useSnootyMetadata() )

src/components/Header/index.js
```diff
@@ -16,20 +17,47 @@ const StyledHeaderContainer = styled.header`
...
 const Header = ({ sidenav, eol }) => {
   const { project } = useSiteMetadata();
+  const { activeVersions, availableVersions } = useContext(VersionContext);
+
+  const gitBranchName = activeVersions[project];
+  const { node } = availableVersions;
+
+  let version;
+
+  // For some branches, such as master, the gitBranchName in the MongoDB collection differs
```
seungpark, line 27: per comment below in drivers-set.js, we can call useSnootyMetadata for branch info here (since it can be not in active versions)

seungpark (changes requested):
A/C states general mechanism for all docs, but seems like this is targeted towards a hardcoded drivers set. not sure how stable that `status` page is (from the Jira issue) but it doesn't seem like a bad idea to cross check with that response and current project/branch info from snooty metadata

package.json
```diff
@@ -80,7 +80,7 @@
     "@leafygreen-ui/tooltip": "^7.1.0",
     "@leafygreen-ui/typography": "^13.0.0",
     "@loadable/component": "^5.14.1",
-    "@mdb/consistent-nav": "1.2.27",
+    "@mdb/consistent-nav": "1.2.32",
```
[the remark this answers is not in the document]
branberry: I needed to bump the version because there was a bug in the unified nav that was addressed in the [latest release](https://github.com/10gen/consistent-nav/pull/146/files)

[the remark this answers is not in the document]
seungpark: gotcha 👍 https://github.com/10gen/consistent-nav/pull/142

branberry:
Converting this back to draft as I work on addressing issues discussed in standup. Namely, the issue with mapping to manifest values from the docs-search-transport

branberry:
@seungpark @schmalliso I reached out to Rachelle and she confirmed that this change should be for all docs sites, and not just the drivers sites.

rayangler (commented):
I wanted to touch base on an edge case first to check what the ideal behavior is. Otherwise, looks good to me

src/components/Header/index.js
```diff
@@ -16,20 +18,29 @@ const StyledHeaderContainer = styled.header`
...
-    searchProperty = new URL(window.location).searchParams.get('searchProperty');
+    const { searchParams } = new URL(window.location);
+    searchProperty = searchParams.get('searchProperty');
   }
-
   const shouldSearchRealm = project === 'realm' || searchProperty === 'realm-master';
   const unifiedNavProperty = shouldSearchRealm ? 'REALM' : 'DOCS';
 
+  const searchParams = [];
+  const projectManifest = `${project}-${branch}`;
```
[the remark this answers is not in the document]
branberry: Hi Raymund! Thank you for pointing this out. I actually updated the atlas app services function that returns the search manifest mapping to account for these variations in the search manifest name. We can discuss some time tomorrow, and I'll update the PR to reflect this change.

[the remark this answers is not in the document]
rayangler: Makes sense! Taking a look at the change

src/components/Header/index.js
```diff
@@ -16,20 +18,46 @@ const StyledHeaderContainer = styled.header`
...
+  /**
+   * The searchPropertyMapping object will contain a new property
+   * called "propertyToSearchMap". This map contains a mapping from
+   * the project name to the true search manifest name in cases where the
+   * project name is NOT the search manifest name.
+   *
+   * NOTE: The change in the fetchSearchPropertyMapping that adds this property
+   * is currently in the DRAFT state, and is not included at this time.
+   */
+  if (project in searchPropertyMapping.projectToSearchMap) {
```
rayangler, line 46: It might be good to include a check if `searchPropertyMapping.projectToSearchMap` exists. Otherwise, building might result in an error when `searchPropertyMapping` is still an empty object by default

src/components/Header/index.js
```diff
@@ -16,20 +18,46 @@ const StyledHeaderContainer = styled.header`
...
   const shouldSearchRealm = project === 'realm' || searchProperty === 'realm-master';
   const unifiedNavProperty = shouldSearchRealm ? 'REALM' : 'DOCS';
 
+  const searchParams = [];
+
+  let searchManifestName = project;
+
+  /**
+   * The searchPropertyMapping object will contain a new property
+   * called "propertyToSearchMap". This map contains a mapping from
```
rayangler, line 39: (Nit) Should probably update `propertyToSearchMap` to `projectToSearchMap` in this comment

src/components/Header/index.js
```diff
@@ -16,20 +18,46 @@ const StyledHeaderContainer = styled.header`
...
+   * the project name to the true search manifest name in cases where the
+   * project name is NOT the search manifest name.
+   *
+   * NOTE: The change in the fetchSearchPropertyMapping that adds this property
+   * is currently in the DRAFT state, and is not included at this time.
```
rayangler, line 44: Small reminder that we can remove this note from the comment now!

rayangler (approved):
LGTM! I like how the custom mapping for edge cases is in the Atlas App Services function so we can easily alter if needed

PR #775, opened 2023-02-24
Bug/DOP-3521

caesarbell:
### Stories/Links:

DOP-3521

### Current Behavior:

### Staging Links:

### Notes:

This PR fixes the initial bug which is recorded in [DOP-3521](https://jira.mongodb.org/browse/DOP-3521) (i.e. side-nav overlay issue)

https://user-images.githubusercontent.com/5807473/221238498-43979123-0c70-4507-b62c-8b8a3b9e4f7c.mov

This PR also addresses some quick browser console warnings which are highlighted below; 

**Unknown Prop isActive**
Screenshot 2023-02-24 at 10 24 01 AM

**Duplicate key value being rendered**

Screenshot 2023-02-24 at 10 12 28 AM

Right Side Bar still works after code changes 

https://user-images.githubusercontent.com/5807473/221241898-ea28e706-dccf-4be4-99c7-d9b83048fb6b.mov


src/components/Breadcrumbs/BreadcrumbContainer.js
```diff
@@ -46,8 +46,9 @@ const BreadcrumbContainer = ({ homeCrumb, lastCrumb }) => {
     <>
       {breadcrumbs.map(({ title, url }, index) => {
         const isFirst = index === 0;
+        const renderKey = typeof title === 'string' ? title : title[0].value;
```
seungpark, line 49: might be safe to null guard this title[0].value read

seungpark (approved):
this LGTM. minor comment on safeguarding null values

src/components/Breadcrumbs/BreadcrumbContainer.js
caesarbell: @seungpark good point, I shouldn't make the assumption that there will always be an object in the title array. I will update this PR.

src/components/Contents/ContentsListItem.js
```diff
@@ -55,8 +55,8 @@ const linkStyling = ({ depth, isActive }) => css`
 
 const ContentsListItem = ({ children, depth = 0, id, isActive = false }) => {
   return (
-    <li className={listItemStyling({ isActive })} isActive={isActive}>
-      <Link className={linkStyling({ depth, isActive })} to={`#${id}`} depth={depth} isActive={isActive}>
+    <li className={listItemStyling({ isActive })}>
+      <Link className={linkStyling({ depth, isActive })} to={`#${id}`} depth={depth}>
```
casthewiz, line 59: I think we want to keep the `isActive` within the `Link` - it gets spread into args [here](https://github.com/mongodb/snooty/blob/4ec59afbc51abe379c3d802612a0dc01b7c8bf2e/src/components/Link.js#L68)

casthewiz (approved):
Broadly lgtm, appreciate resolving the console errors! 

Flagging one spot where we might want to keep the `isActive`, might want to double check against the LeafyGreen component. I only did a surface level trace of where that prop goes.

src/components/Contents/ContentsListItem.js
caesarbell: @casthewiz I did catch that too and traced it to the `GatsbyLink`, which led me to check the documentation on [Gatsby's Link API](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-link/#link-api-surface-area). 

I did see we currently are passing it using the spread operator, but if we look at the docs there are no `isActive` props, which leads me to feel comfortable with removing it. I don't believe that prop was actually having an effect. I can put it back if we feel like it is being used somewhere somehow. 

I will lean on your final thought.

casthewiz: Let's keep it to be safe - documentation on Gatsby V2 is scant and hard to navigate, and not sure if it's a prop whose support was removed between v2 and v5. Should be safe to remove post v5 though 👍

PR #780, opened 2023-03-10
Remove unused Searchbar subcomponents

rayangler:
### Stories/Links:

N/A

### Current Behavior:

### Staging Links:

### Notes:

For context, the MongoDB Docs site used to have its own top nav, which housed a searchbar for docs content. This would allow users to search using both the top nav and the search results page found in `docs-landing` (using the `SearchResults` component). This searchbar was ultimately made obsolete with the new UnifiedNav and its own method of searching. The existing `SearchResults` component would still share some components with the old `Searchbar` component. This PR hopes to:

1) Remove unused `Searchbar` subcomponents to keep the repo cleaner.
2) Consolidate used `Searchbar` subcomponents with the `SearchResults` component by moving file locations. This is to help avoid further confusion and keep things more organized.


src/components/SearchResults/SearchResult.js
```diff
@@ -162,10 +161,7 @@ const SearchResult = React.memo(
       const prevSibling = getNestedValue(['current', 'previousSibling'], resultLinkRef);
       if (prevSibling) {
         // If these don't match, we have gone up out of the results
-        if (prevSibling.nodeName !== resultLinkRef.current.nodeName) {
-          // This is the first result, so let's go to the search bar
-          document.querySelector(`${StyledTextInput} input`).focus();
-        } else {
+        if (prevSibling.nodeName === resultLinkRef.current.nodeName) {
```
rayangler, line 164: Unclear if this functionality ever worked within the context of the `SearchResults` component/page (I remember it working for the old searchbar's search results). 

Wondering if we should just remove the `onArrowDown` and `onArrowUp` functions altogether now. Similar keyboard accessibility can be achieved using the "tab" key. But perhaps this is something to revisit and discuss with design for UX considerations in the future

mmeigs (approved):
LGTM! I don't have a strong opinion on the arrow functionality. I think it's nice to have, if possible. Currrently that does seem to be broken. But I also don't know how to spin up a build of the search functionality to see if your change fixes it.

src/components/SearchResults/SearchResult.js
mmeigs: All the arrow functions don't seem to be working. I just realized they all depend on the searchContainerRef, which is not in use anymore. So, either we reorganize this functionality totally or yes, just let it go

rayangler: Thanks for pointing this out! Removed the unused `searchContainerRef` and the unused arrow key functions altogether.

PR #781, opened 2023-03-13
Feature/DOP-3501

caesarbell:
### Stories/Links:

DOP-3501

### Current Behavior:

https://www.mongodb.com/docs/drivers/node/current/quick-start/download-and-install/?presentation=true

You should still see the following elements

- Main Nav
- Side Nav
- Footer
- Feedback Widget

### Staging Links:

We should see the components not render (in this case we shouldn't see the layout drastically change to a mobile view)
https://docs-mongodbcom-integration.corp.mongodb.com/faa7aecf6cc84bc73a71d336d6bc413bfbd341cf/master/node/caesarbell/feature/DOP-3501/?presentation=true

We should see the components render
https://docs-mongodbcom-integration.corp.mongodb.com/faa7aecf6cc84bc73a71d336d6bc413bfbd341cf/master/node/caesarbell/feature/DOP-3501/

**side note**
There is a small flicker if we feel this is too distracting we can look at possibly dynamically changing the grid layout to support when the side nav is missing. 

### Notes:

The ask for the ticket [DOP-3501](https://jira.mongodb.org/browse/DOP-3501), was to provide the ability to handle a certain query param to be used which would force certain elements to not render. The goal is to use this for PathFactory, in hopes they can use the new URL + query param and hide the side nav and header to reduce the distraction or navigation away from the content that will be in focus. 

To test this locally, we can add the following query param to the URL `i.e. /?presentation=true`


src/layouts/index.js
```diff
@@ -75,12 +77,29 @@ const GlobalGrid = styled('div')`
...
+      return path.has(QUERY_PARAM_FOR_DISABLING_ELEMENTS);
+    };
+
+    const forPresnetationModeLocalDev = () => {
+      const url = new URL(location.href);
+      const hash = url.hash;
+      const queryParam = `?${QUERY_PARAM_FOR_DISABLING_ELEMENTS}`;
+      return hash.includes(queryParam);
+    };
+    return process.env.GATSBY_SNOOTY_DEV ? forPresnetationModeLocalDev() : forPresentationMode();
```
seungpark, line 100: do we have the differentiate for the flag?

seems like we can useLocation and use query-string as in the [searchResults](https://github.com/mongodb/snooty/blob/master/src/components/SearchResults/SearchResults.js#L276), unless i'm missing something here

seungpark (changes requested):
AC mentioned using query-string to parse the Location object properties. is this not part of the requirements?

src/layouts/index.js
caesarbell: @seungpark good call out. I was unaware of these methods that already exist in the codebase. With the search can we limit the value? For our requirement, this should be locked down to a particular query value i.e. `? presentation` or whatever value is agreed on for use of PathFactory.

caesarbell: @seungpark just out of curiosity, why are we pulling in a 3rd party library for parsing query strings when we can use native JS methods to handle the parsing?

caesarbell: I might have typed before looking at the full question so from lines 95 - 98 I have this logic in place to handle the difference between prod & stag vs dev. The URL structure on dev has the hash trailing at the end of the routes i.e. http://localhost:8000/node/caesarbell/master/quick-start/#install-node-and-npm. So to handle the hashing and not force the developer to add the query param in between the trailing `/` and `#`  I added this logic for presentation development mode.  Pls let me know if that makes sense.

seungpark: discussed offline diffs on anchors and query params. useLocation is preferred because it's [used by gatsby](https://www.gatsbyjs.com/docs/reach-router-and-gatsby/) under the hood

caesarbell: @seungpark good point. I pushed up the changes.

src/layouts/index.js
```diff
@@ -78,9 +80,15 @@ const DefaultLayout = ({
   children,
   pageContext: { page, slug, repoBranches, template, associatedReposInfo, isAssociatedProduct, remoteMetadata },
 }) => {
+  const { search } = useLocation();
```
seungpark, line 83: this works well! tested locally and able to get the preview 👍

seungpark (approved):
think we missed a few spots to sport "presentation mode" ie. feedback widget and footer. these can be found in DocumentBody component

caesarbell:
> think we missed a few spots to sport "presentation mode" ie. feedback widget and footer. these can be found in DocumentBody component

Totally missed this. Thanks for catching that.

seungpark:
as for current behavior vs staging links, we have an [npm script](https://github.com/mongodb/snooty/blob/master/package.json#L33) that builds and stages the changes to S3, and the link provided should give us a preview of the changes.

src/components/Sidenav/Sidenav.js
```diff
@@ -238,87 +249,90 @@ const Sidenav = ({ chapters, guides, page, pageTitle, repoBranches, siteTitle, s
...
-                {ia && (
-                  <IA
-                    header={<span className={cx([titleStyle])}>{formatText(pageTitle)}</span>}
-                    handleClick={() => {
-                      setBack(false);
-                      hideMobileSidenav();
-                    }}
-                    ia={ia}
-                  />
+        {!isInPresentationMode && (
```
seungpark, line 252: can we call this on the layout file? can omit <Sidenav entirely

src/components/Widgets/index.js
```diff
@@ -16,8 +16,12 @@ const Widgets = ({ children, pageOptions, pageTitle, publishedBranches, slug })
   return (
     <FeedbackProvider page={feedbackData} hideHeader={hideFeedbackHeader}>
       {children}
-      <FeedbackTab />
-      <FeedbackForm />
+      {!isInPresentationMode && (
```
seungpark, line 19: similar here, we can omit from <Widgets> from DocumentBody

src/layouts/index.js
```diff
@@ -121,6 +125,18 @@ const DefaultLayout = ({
 };
 
 DefaultLayout.propTypes = {
+  location: PropTypes.shape({
```
seungpark, line 128: can remove if using the hook for location

src/components/Sidenav/Sidenav.js
caesarbell: The grid layout focuses on two areas, side-nav, and content, if we have the logic in the layout and omit the `<Sidenav` entirely it breaks the grid layout. i.e.

```
const GlobalGrid = styled('div')`
  display: grid;
  grid-template-areas:
    'header header'
    'sidenav contents';
  grid-template-columns: auto 1fr;
  grid-template-rows: auto 1fr;
`;
```

src/components/Widgets/index.js
caesarbell: Similarly here, if we omit on the entire DocumentBody i.e. 

```
<Widgets
        location={location}
        pageOptions={page?.options}
        pageTitle={pageTitle}
        publishedBranches={getNestedValue(['publishedBranches'], metadata)}
        slug={slug}
        isInPresentationMode={isInPresentationMode}
      >
        <FootnoteContext.Provider value={{ footnotes }}>
          <Template {...props}>
            {pageNodes.map((child, index) => (
              <ComponentFactory key={index} metadata={metadata} nodeData={child} page={page} slug={slug} />
            ))}
          </Template>
        </FootnoteContext.Provider>
      </Widgets>
```

The site will not build correctly. Good suggestion, I did think about this approach at first. Let me know if you have any more questions about this.

seungpark: see responded comment above regarding grid

caesarbell: This one is a little different than the one above. This had to do with the Widget Provider not rendering crashing the site build, but if you prefer for it to be `one hook / one template` I can look more into refactoring this as well.

src/components/Sidenav/Sidenav.js
caesarbell: That is fine, I can make the changes to the grid rather than passing the boolean via props and hiding the side nav content.

src/components/Sidenav/Sidenav.js
```diff
@@ -291,8 +291,7 @@ const Sidenav = ({ chapters, guides, page, pageTitle, repoBranches, siteTitle, s
                 <SideNavItem
                   className={cx(titleStyle, sideNavItemBasePadding)}
                   as={Link}
-                  to={isGuidesTemplate ? slug : activeToc.url || activeToc.slug || '/'}
-                  hideExternalIcon={true}
+                  to={isGuidesTemplate ? slug : '/'}
```
seungpark, line 294: minor. i think this change was unintentional

seungpark (approved):
this LGTM 👍 
thanks for putting this all together !

PR #786, opened 2023-03-22
DOP-3559

crisatmongo:
Deprecated CSSClass component removed from Component Factory, Document Body, and Document Body Preview. Component directory and subdirectories also removed.

### Stories/Links:

DOP-3559


mmeigs (commented):
Looks good! Just going to defer to @rayangler on the A/C in the ticket stating: 
`Usage of cssclass directive will yield a "not yet implemented" warning`
This warning might be more useful within snooty-parser, but either way it's something to be checked before approval.

rayangler:
> This warning might be more useful within snooty-parser, but either way it's something to be checked before approval.

@mmeigs That A/C was in reference to the Component Factory's existing [warning](https://github.com/mongodb/snooty/blob/ece8c56e84242da8b65c253daa5ecc72fa43e1a7/src/components/ComponentFactory.js#L231). I believe since the directive is only marked as "deprecated" by the parser, there's a chance a node can make it into the AST. In this case, we want to warn to the user that we're intentionally not rendering anything in the frontend.

If a directive does not exist in the parser, then it will print out an appropriate error. Will double-check this behavior later when I get the chance!

mmeigs (approved):
@rayangler was completely right. The parser emits a warning, and now snooty also emits a warning. Nice!

src/components/ComponentFactory.js
```diff
@@ -131,12 +130,10 @@ const componentMap = {
   'card-group': CardGroup,
   chapter: Chapter,
   chapters: Chapters,
-  class: CSSClass,
```
rayangler, line 134: I didn't even know that the `class` directive was mapped to the `CSSClass` component.

@schmalliso should we actually keep the `CSSClass` component alive for the `class` directive until [DOP-2317](https://jira.mongodb.org/browse/DOP-2317) is resolved to avoid potential conflicts with docs that still use `.. class:: hidden`?

schmalliso: UUUGH. Let me figure out how many instances of class are actually "live" in the content and see if we can get rid of that while we're at it. The writer who was passionate about the hiding is no longer at MongoDB so we may be able to move past it. I'll try to get an answer by tomorrow.

schmalliso: So I *think* this is all of the `.. class:: hidden` instances in the main branches of the docs repos

```
.//docs-mongodb-shell/source/reference.txt:.. class:: hidden
.//docs-mongodb-shell/source/snippets.txt:.. class:: hidden
.//docs-mongodb-shell/source/crud.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs-app-services/source/functions/mongodb/api.txt:.. class:: hidden
.//docs/source/reference/read-concern-snapshot.txt:.. class:: hidden
.//docs/source/reference/read-concern-available.txt:.. class:: hidden
.//docs/source/reference/read-concern-local.txt:.. class:: hidden
.//docs/source/reference/limits.txt:.. limit:: Hidden Indexes
```

Which is really not that many - a lot are on the app services functions/mongodb/api page handling methods. It seems plausible that we can work with content to find a better solution here, but lets not merge until we reach consensus.

schmalliso (approved):

schmalliso:
@crisatmongo Can you rebase this so we can merge it in, please?

PR #792, opened 2023-04-06
DOP-3566

caesarbell:
### Stories/Links:

DOP-3566

### Current Behavior:

### Staging Links:

### Notes:

This PR focuses on the testing for https://github.com/mongodb/snooty/pull/781

Screenshot 2023-04-06 at 1 42 55 PM


tests/unit/Presentation.test.js
```diff
@@ -0,0 +1,77 @@
...
+      history = createHistory(window);
+
+      render(
+        <LocationProvider history={history}>
+          <DocumentBody location={window.location} pageContext={mockPageContext} />
+        </LocationProvider>
+      );
+    });
+
+    it('renders the footer', () => {
```
seungpark, line 27: minor nitpick. since we are using the same rendered tree for each test, we can combine these assertions into a single test.

also would add a snapshot in the two cases

seungpark (changes requested):
this looks great 👍 mocking location.search works as expected. requesting to add snapshots in the two cases

tests/unit/Presentation.test.js
```diff
@@ -0,0 +1,77 @@
...
+      Object.defineProperty(window, 'location', {
+        value: {
+          search: '?presentation=true',
+        },
+      });
+      history = createHistory(window);
+
+      render(
+        <LocationProvider history={history}>
+          <DocumentBody location={window.location} pageContext={mockPageContext} />
```
rayangler, line 54: The [SearchResults test](https://github.com/mongodb/snooty/blob/90eed3bc7600a9e0fd763109fee49bec6c6badd4/tests/unit/SearchResults.test.js#L79) has a mock of `useLocation` that we can probably turn into a generalized util function for something like this. 

Do you think that mocked hook would be useful for this case? It might help with minimizing some of the setup here + would be nice to have since we're introducing more tests with query strings

rayangler (commented):
Looks good! Just wanted to bring up one thing that may be helpful for these tests. Feel free to let me know if it's not a good idea. Otherwise LGTM

tests/unit/Presentation.test.js
caesarbell: I will condense the assertions into a single test.

tests/unit/Presentation.test.js
caesarbell: @rayangler good call. Turning the `mockLocation` into a reusable utility help reduce the steps needed for the presentation test. In another PR we can update the [SearchREsults Test](https://github.com/mongodb/snooty/blob/90eed3bc7600a9e0fd763109fee49bec6c6badd4/tests/unit/SearchResults.test.js#L79) to use the new utility function.

rayangler (approved):
LGTM!

PR #795, opened 2023-04-24
Pin node version to 18.15

rayangler:
### Stories/Links:

N/A

### Notes:
* Pins `.nvmrc` and GH Actions workflows to use Node v18.15. v18.16 results in tests failing: https://github.com/mongodb/snooty/actions/runs/4745313973/jobs/8451939301


caesarbell (approved):
LGTM

PR #798, opened 2023-04-25
DOP-3608: wait for manifest to fire search

seungpark:
### Stories/Links:

DOP-3608

### Current Behavior:

[prod link with search results](https://www.mongodb.com/docs/search/?q=test)
opening the above link may flash a `no results found` before actually showing results
**NOTE:** opening network tab and refreshing this page shows **two** requests to `https://docs-search-transport.mongodb.com/search?q=test`

### Staging Links:

[staging link for search page](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/seung.park/DOP-3608/search/index.html?q=test)
(cannot refresh above link due to S3 hosting issues). can test refresh by copying and pasting link https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/seung.park/DOP-3608/search/index.html?q=test
- does not show `no results found` when opening link, should only fire one request to search transport server for the query

### Notes:
- this was due to the dependencies passed to React.useEffect `[searchFilter, searchPropertyMapping, searchTerm]`. the `searchPropertyMapping` is required before showing search results but this effect was being fired without checking existence of it, and also again when it changed, resulting in a `no results found` briefly

### DIFF
--- a/src/components/SearchResults/SearchResults.js
+++ b/src/components/SearchResults/SearchResults.js
@@ -386,7 +386,6 @@ const SearchResults = () => {
                         margin-top: 80px;
                       `}
                     >
-                      empty res here at 388
                       <EmptyResults />
                     </EmptyResultsContainer>
                     <FiltersContainer>
--- a/src/components/SearchResults/SearchResults.js
+++ b/src/components/SearchResults/SearchResults.js
@@ -280,15 +280,16 @@ const SearchResults = () => {
   // Update results on a new search query or filters
   // When the filter is changed, find the corresponding property to display
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


rayangler (approved):
Nice find! 

I'm assuming the log "empty res here at 388" that appears in the staging link when no results are actually found is just from local staging and is not expected to appear in production?

seungpark:
> Nice find! Thank you for the note explaining the cause of the bug
> 
> I'm assuming the log "empty res here at 388" that appears in the staging link when no results are actually found is just from local staging and is not expected to appear in production?

yup! this was staged from the first commit (before removing testing code and updating tests)

PR #801, opened 2023-05-01
DOP-3599

caesarbell:
### Stories/Links:

DOP-3599

### Current Behavior:

The page nodes contain a `position` field which isn't needed anymore.

```js
page nodes {
  type: 'root',
  position: { start: { line: 0 } },
  children: [
    {
      type: 'target',
      position: [Object],
      children: [Array],
      domain: 'std',
      name: 'label',
      html_id: 'std-label-python-language-center'
    },
    { type: 'section', position: [Object], children: [Array] }
  ],
  fileid: 'python.txt',
  options: { headings: [ [Object] ] }
}
```

### Staging Links:

After adding the new utility we have the ability to remove the `position` field

```js
page nodes {
  type: 'root',
  children: [
    {
      type: 'target',
      children: [Array],
      domain: 'std',
      name: 'label',
      html_id: 'std-label-python-language-center'
    },
    { type: 'section', children: [Array] }
  ],
  fileid: 'python.txt',
  options: { headings: [ [Object] ] }
}
```


src/utils/remove-nested-value.js
```diff
@@ -0,0 +1,26 @@
...
+ * Arguments:
+ * - t: the target to have deleted
+ * - k: the traversal target (key)
+ * - arr: an array containing the objects to search for the target value
+ */
+const removeNestedValue = (t, k, arr) => {
+  if (!arr) return undefined;
+
+  arr.forEach((entry) => {
+    Object.entries(entry).forEach(([key, val]) => {
```
i80and, line 12: What do we get by iteratomg over each key/value pair rather than inspecting `entry[t]` and `entry[k]` directly?

e.g.

```
if (entry.hasOwnProperty(t)) {
    delete entry[t];
}

if (entry.hasOwnProperty(k)) {
    const children = entry[k];
    if (Array.isArray(children) && children.length > 0) {
        removeNestedValue(t, k, children);
    }
}
```

src/utils/remove-nested-value.js
```diff
@@ -0,0 +1,26 @@
+/*
+ * Safely remove  values from a deeply nested object.
+ * Arguments:
+ * - t: the target to have deleted
+ * - k: the traversal target (key)
+ * - arr: an array containing the objects to search for the target value
+ */
+const removeNestedValue = (t, k, arr) => {
```
i80and, line 8: Long descriptive variable names would preferable; e.g. `deletionKey`, `arrayChildrenKey`, and `arr`

src/utils/remove-nested-value.js
caesarbell: Good point. After reviewing, we don't get any extra benefits from iterating over each key/value. Thanks for sharing an example as well.

src/utils/remove-nested-value.js
```diff
@@ -0,0 +1,25 @@
+/*
+ * Safely remove  values from a deeply nested object.
+ * Arguments:
+ * - t: the target to have deleted
```
i80and, line 4: Comment needs to be updated! 😸

i80and (changes requested):
Just one change needed from me!

src/utils/remove-nested-value.js
caesarbell: 💯  how did I miss that? In my mind, I was cleaning that up already. 🤣

caesarbell: @i80and done. Thank you again for catching that.

i80and (approved):
:shipit:

seungpark (approved):
very clean! LGTM

schmalliso (approved):
:shipit:

src/utils/remove-nested-value.js
```diff
@@ -0,0 +1,25 @@
+/*
+ * Safely remove  values from a deeply nested object.
+ * Arguments:
+ * - target: the target key to be deleted
+ * - traversalKey: the target key to traverse
+ * - arr: an array containing the objects to search for the target value
+ */
+const removeNestedValue = (target, traversalKey, arr) => {
+  if (!arr) return undefined;
```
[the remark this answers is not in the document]
caesarbell: @i80and making this change before merging.

PR #802, opened 2023-05-02
[DOP-3680]: Create OpenAPIChangelog filter panel component

branberry:
### Stories/Links:

[DOP-3680](https://jira.mongodb.org/browse/DOP-3680)

### Screenshots

image

image

### Notes:
This PR implements the filters panel for the OpenAPI change log. It includes functionality to select diffs between versions, and ensures that you cannot compare the same resource version. There is also a combobox that allows users to view a specific resource's change log.


tests/unit/OpenAPIChangelog.test.js
```diff
@@ -0,0 +1,143 @@
...
+      const { getByTestId, getByLabelText } = render(<OpenAPIChangelog />);
+
+      const allVersionsOption = getByTestId('all-versions-option');
+
+      const isAllVersionsSelected = allVersionsOption.firstElementChild.getAttribute('aria-selected');
+
+      expect(isAllVersionsSelected).toBe('true');
+
+      expect(() => getByLabelText('Resource Version 1')).toThrowError();
+      expect(() => getByLabelText('Resource Version 2')).toThrowError();
```
mmeigs, line 45: I think there are some more semantic methods to use to show that components are not visible that would maybe make more sense here

src/components/OpenAPIChangelog/FiltersPanel/components/FiltersPanel.js
```diff
@@ -0,0 +1,41 @@
...
+  resourceVersionOne,
+  resourceVersionTwo,
+  setSelectedResource,
+  setIsVersionCompare,
+  setResourceVersionOne,
+  setResourceVersionTwo,
+}) => {
+  return (
+    <Wrapper>
+      <VersionModeSegmentedControl isVersionCompare={isVersionCompare} handleChange={setIsVersionCompare} />
```
mmeigs, line 26: My personal opinion (take it or leave it!!) is to only spin off separate components if they will be reused *or* if those components have become unwieldy or large in some way for clarity.
I think DiffSelect makes sense! I think VersionModeSegmentedControl and ResourceSelect are both very small with hardly any logic and will never be reused in their current state. But I'm all for a discussion on this!

src/components/OpenAPIChangelog/FiltersPanel/components/DiffSelect.js
```diff
@@ -0,0 +1,42 @@
...
+    <DiffSelectContainer>
+      <DiffSelectItem>
+        <Combobox value={resourceVersionOne} label="Resource Version 1" onChange={handleVersionOneChange}>
+          {versionOneOptions}
+        </Combobox>
+      </DiffSelectItem>
+      <DiffSelectItem>
+        <Combobox value={resourceVersionTwo} label="Resource Version 2" onChange={handleVersionTwoChange}>
+          {versionTwoOptions}
+        </Combobox>
```
mmeigs, line 38: I believe these are just Select components rather than Combobox. We don't need any bells and whistles for these!

src/components/OpenAPIChangelog/FiltersPanel/components/ResourceSelect.js
```diff
@@ -0,0 +1,19 @@
...
+
+export default function ResourceSelect({ resources, selectedResource, handleChange }) {
+  return (
+    <ResourceSelectContainer>
+      <Combobox label="Select Resource" value={selectedResource} onChange={handleChange}>
+        {resources.map((version) => (
+          <ComboboxOption key={version} value={version} data-testid="resource-select-option" />
+        ))}
+      </Combobox>
+    </ResourceSelectContainer>
```
mmeigs, line 17: I believe we can use defaultValue for "All" rather than passing in a value at all. That also should help with the flicker on the screen. Could be wrong there, though.
Also, we need to add multiselect to the Combobox so the user can choose multiple Resources.

mmeigs (commented):
Looks good! Thanks for all this work! 
I have put a few comments below mainly about components and props and maybe consolidating some things.
Also, a few styles are not quite in line with the design. Next week can you do another pass through the [Figma](https://www.figma.com/file/Tal3RZFxMEz6u1tnilxwRY/API-Change-Log?type=design&node-id=1-539&t=hkiLQcmRrU8QqE0j-0)?

mmeigs (approved):

PR #806, opened 2023-05-04
DOP-2317: comply with the  hidden flag for targets

schmalliso:
### Stories/Links:

DOP-2317

### Current Behavio
We supported cssclass to hide components previously; now we'll let writers use a `:hidden:` flag on their methods if they want to hide them, and respect that.

### Staging Links:

https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOP-3627/atlas-app-services/allison/master/functions/mongodb/api/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ docs-tools/
 coverage/
 .coverage
 .env.*
+.vscode/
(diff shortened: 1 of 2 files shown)


mmeigs (approved):
Sorry, that took so long - I wanted to test it out. look gorgeous :)

PR #810, opened 2023-05-15
DOP-3668

caesarbell:
### Stories/Links:

DOP-3668

### Current Behavior:

There is an error with `webpack` and process and because of this error, some elements are not rendering on this page.

https://www.mongodb.com/docs/atlas/app-services/functions/javascript-support/

### Staging Links:

monkey patching the process and installing [browser-process](https://www.npmjs.com/package/browser-process) emulates the node process in the browser.

https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/9dc8d9fbf80ebff75c1a9a50999d9e61cd4f4ef4/master/atlas-app-services/caesarbell/DOP-3668/functions/javascript-support/index.html

### Notes:

I learned about the monkey patching approach from [here](https://www.npmjs.com/package/process)


package.json
```diff
@@ -21,6 +21,7 @@
     "babel-jest": "^29.5.0",
     "babel-plugin-emotion": "^11.0.0",
     "babel-preset-gatsby": "^3.0.0",
+    "browser-process": "^0.0.1",
```
rayangler, line 24: Is this package necessary to keep? I don't see it used anywhere and my assumption is that your changes involving `process/browser` resolve the issue.  I tried uninstalling it and it seems to be working fine?

package-lock.json
```diff
@@ -37464,19 +37564,21 @@
       "dev": true
     },
     "webpack": {
-      "version": "5.75.0",
+      "version": "5.82.1",
```
rayangler, line 37567: There are a lot of dependencies bumped in the package-lock. Are these all relevant and/or necessary to keep? I initially thought they would be tied to the `browser-process` package, but it doesn't have any dependencies

rayangler (commented):
This is a nice fix! I tried building locally and it resolves the issue for that page!

Just a couple of questions about the need for certain package changes

package-lock.json
caesarbell: @rayangler good catch, this was from some older trial-and-error approaches. My intent was to clean that up before pushing it up. Thanks for catching that.

package.json
caesarbell: @rayangler good call. I was able to remove the package as well and it still works. This was also residue from older trial-and-error attempts.

rayangler (approved):
Looks great 😎

PR #817, opened 2023-05-24
DOP-3716: dynamically fetches diff for OpenAPIChangelog

mmeigs:
### Stories/Links:

DOP-3617

### Notes:

The `OpenAPIChangelog` now dynamically fetches the diff between two `Resource Versions` whenever a user chooses two `Resource Versions` to compare. 

Also:
- moves the appending of " (latest)" to a display value rather than modifying the value in state.
- deals with `SegmentedContro`l `aria-controls` warning
- removes the change badges from the diffs as referenced by the design
- hides any changes that have the `changeType` "operation-id-changed" or "operation-tag-changed" as desired by APIx
- creates an [Atlas App Services Function](https://realm.mongodb.com/groups/5bad1d3d96e82129f16c5df3/apps/5d41ef5e2de84772fc7c2de2/functions/646cc47cf571983e9ce0671e) `fetchOADiff` to enable this querying and to bypass CORS

### DIFF
--- a/src/components/OpenAPIChangelog/OpenAPIChangelog.js
+++ b/src/components/OpenAPIChangelog/OpenAPIChangelog.js
@@ -5,12 +5,11 @@ import { Body, H2 } from '@leafygreen-ui/typography';
 import Button from '@leafygreen-ui/button';
 import { theme } from '../../theme/docsTheme';
 import useChangelogData from '../../utils/use-changelog-data';
-import { fetchOADiff } from '../../utils/realm';
 import FiltersPanel from './components/FiltersPanel';
 import ChangeList from './components/ChangeList';
+import { useFetchDiff } from './utils/useFetchDiff';
 import { ALL_VERSIONS, getDownloadChangelogUrl } from './utils/constants';
 import getDiffResourcesList from './utils/getDiffResourcesList';
-import { getDiffRequestFormat } from './utils/getDiffRequestFormat';
 
 const ChangelogPage = styled.div`
   width: 100%;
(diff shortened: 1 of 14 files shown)


src/components/OpenAPIChangelog/OpenAPIChangelog.js
```diff
@@ -54,28 +55,41 @@ const DownloadButton = styled(Button)`
...
   const [filteredDiff, setFilteredDiff] = useState(diff);
   const [filteredChangelog, setFilteredChangelog] = useState(changelog);
 
-  /*  
-    Clear filters on version mode change.
-    Different Resources are available in either mode, not always comparable.
-  */
+  /* Fetch diff on selection of two Resources to compare */
+  useEffect(() => {
+    if (!resourceVersionOne || !resourceVersionTwo || !index.runId) return;
```
branberry, line 76: NIT: I think it could be helpful to make this logic a custom hook, but also not really a big deal keeping it here

src/components/OpenAPIChangelog/OpenAPIChangelog.js
```diff
@@ -54,28 +55,41 @@ const DownloadButton = styled(Button)`
...
+    const fromAndToDiffString = getDiffRequestFormat(resourceVersionOne, resourceVersionTwo);
+
+    fetchOADiff(index.runId, fromAndToDiffString)
+      .then((response) => setDiff(response))
+      .catch((err) => console.error(err));
```
branberry, line 81: Do you think we should display an error if the fetch fails? If so, we would need an error state variable.

branberry (commented):
Hi Matt! Looks really good, just two small comments.

src/components/OpenAPIChangelog/OpenAPIChangelog.js
mmeigs: Oh, that would be cute! Just thought about this for a few minutes and am a bit stumped. Do you have any examples of custom hooks that dynamically take in changing state but don't re-render on every state change? I'm hitting a wall on this

branberry: I think you could do something like this:

```jsx

// useFetchDiff.js

function useFetchDiff({ resourceVersionOne, resourceVersionTwo, index }) {
  const [diff, setDiff] = useState([]);

  useEffect(() => {
    if (!resourceVersionOne || !resourceVersionTwo || !index.runId) return;
    const fromAndToDiffString = getDiffRequestFormat(resourceVersionOne, resourceVersionTwo);

    fetchOADiff(index.runId, fromAndToDiffString)
      .then((response) => setDiff(response))
      .catch((err) => console.error(err));
  }, [resourceVersionOne, resourceVersionTwo, index.runId]);

  return [diff, setDiff];
}

...

// OpenAPIChangelog.js

const [diff, setDiff] = useFetchDiff({ resourceVersionOne, resourceVersionTwo, index });
```

Do you think this would be worthwhile including?

mmeigs: Right, I guess my one hesitation is that is wondering if this re-renders on each state change that doesn't include those three pieces of state. I'll experiment!

branberry: Oh, gotcha! I see what you mean. If you were to change the state of some other variable that re-renders the OpenAPIChangelog component, it would call `useFetchDiff` with new values, is that correct? I guess you would have to wrap the call in a `useCallback` maybe?

mmeigs: I don't think you can put hooks inside hooks! One of React's little laws. But I think this still all works great!

branberry (approved):
LGTM!

PR #826, opened 2023-05-31
clear selected resources on change

mmeigs:
### Stories/Links:

DOP-3737

### Notes:

Tiny PR to ensure that selected resources for the resources list gets cleared when the resources list changes. Edge case for when there is more than one diff.


branberry (approved):
LGTM!

PR #827, opened 2023-06-02
DOP-3720: OpenAPI Changelog fallback to RunId

mmeigs:
### Stories/Links:

DOP-3720

### Notes:

On the event of any error in fetching OpenAPI Changelog data, Snooty should read the last successful runId from Atlas to gain valid data for the build.

The `openapi_changelog.atlas_admin_metadata` collection in Atlas will hold the last successful run's metadata.

On a successful build of changelog data, Snooty will upsert the metadata object directly into this collection. 
On errored builds of changelog data, Snooty will fallback to read this older metadata from Atlas and use the requisite data to successfully build the Atlas Admin Changelog.

Note: Also two bugs are fixed in this PR:
- on diff loading state, ensure that the previous diff is not shown along with the loading skeleton
- when sorting Resource Versions by date, slice Resource Version string down to a simple date (there were edge cases on naming conventions such as '2023-08-11~preview')

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -132,12 +132,17 @@ const createOpenAPIChangelogNode = async ({ createNode, createNodeId, createCont
     try {
       const { changelog, changelogResourcesList, mostRecentDiff } = await fetchChangelogData(runId, versions);
       changelogData = { ...changelogData, changelog, changelogResourcesList, mostRecentDiff };
+      await db.stitchInterface.updateOAChangelogMetadata(index);
     } catch (error) {
-      // TODO: Fetch runId & index from Atlas
-      const oldIndex = { runId: 'lalala', versions: [] };
-      const { runId: oldRunId, versions: oldVersions } = oldIndex;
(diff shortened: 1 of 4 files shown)


gatsby-node.js
```diff
@@ -110,15 +99,51 @@ const createOpenAPIChangelogNode = async ({ createNode, createNodeId, createCont
...
       changelog,
       changelogResourcesList,
       mostRecentDiff: {
         mostRecentDiffLabel,
         mostRecentDiffData,
       },
     };
+  } catch (error) {
+    console.warn('Changelog error: Most recent runId not successful. Using last successful runId to build Changelog.');
+    throw new Error(error);
```
rayangler, line 112: (Super nit) Would we be able to just throw the existing error instead of making a new one?

```
throw error;
```

rayangler (approved):
LGTM! Just one super nit thing

gatsby-node.js
mmeigs: Great point. I've never realized I did that, thank you!

PR #830, opened 2023-06-07
DOP-3698

caesarbell:
### Stories/Links:

DOP-3698

### Current Behavior:

[Prod URL](https://www.mongodb.com/docs/atlas/atlas-search/atlas-search-overview/)

### Staging Links:

[Staged URL](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/06f8a43664cc83d8e1cef0cd439253e2d3af5d8e/master/cloud-docs/caesarbell/DOP-3698/atlas-search/atlas-search-overview/index.html)

### Notes:
The issue was the meta options weren't showing up in the HTML. Below you can see in the screenshot after the fix the meta tags are appearing in the DOM.

Screenshot 2023-06-07 at 2 59 49 PM


src/utils/get-meta-from-directive.js
```diff
@@ -0,0 +1,13 @@
+// Grabs the metadata values in question and returns them as an array
+// for the Meta & TwitterMeta tags
+const grabMetaFromDirective = (nodes, target) => {
```
mmeigs, line 3: nit: I'm guessing you changed halfway through, but we should choose either "get" or "grab" and make sure the file and the function name match just for clarity :)

tests/unit/Meta.test.js
```diff
@@ -0,0 +1,31 @@
...
+import { render, screen } from '@testing-library/react';
+import { getNestedValue } from '../../src/utils/get-nested-value';
+import { grabMetaFromDirective } from '../../src/utils/get-meta-from-directive';
+import Meta from '../../src/components/Meta';
+
+// data for this component
+import { metaTestData } from './data/MetaData';
+
+describe('Meta Tag', () => {
+  console.log('this is a test suit');
```
mmeigs, line 11: Just wondering what the reason is for this!

src/components/DocumentBody.js
```diff
@@ -130,15 +131,18 @@ export default DocumentBody;
...
+  // getting the meta from pageNodes
+  // extract the section
+  const section = pageNodes.find((node) => node.type === 'section');
+
+  // get the nested values from section that will be used for the lookup
+  const sectionNodes = getNestedValue(['children'], section);
+
+  const meta = grabMetaFromDirective(sectionNodes, 'meta');
+
+  const twitter = grabMetaFromDirective(sectionNodes, 'twitter');
```
mmeigs, line 144: Could this all be wrapped up in one or two helper functions that live in grab/getMetaFromDirective? If not, that's fine, but it seems like it could be helpfully abstracted away

mmeigs (commented):
Just a few clarifications! :)

src/utils/get-meta-from-directive.js
caesarbell: @mmeigs good callout, yeah I think I changed halfway and didn't go back to align my initial pivot.

tests/unit/Meta.test.js
caesarbell: Oh, I thought I removed that. That little sneaky `console.log` slipped through.

src/components/DocumentBody.js
mmeigs: Cute!

src/components/DocumentBody.js
```diff
@@ -130,15 +131,15 @@ export default DocumentBody;
...
-    const lookup = c.type === 'directive' ? c.name : c.type;
-    return lookup === 'twitter';
-  });
+
+  /**
+   * @param root sets the key to look for in the top-level of pageNodes
+   * @param nodes the children of the type found (i.e. section)
+   * @param target the lookup value that we are looking for
+   */
+  const meta = getMetaFromDirective('section', pageNodes, 'meta');
```
seungpark, line 140: CMIIW, but i think some pages have it in a section and some pages do not

ie. [this page](https://github.com/10gen/cloud-docs/blob/master/source/atlas-search/atlas-search-overview.txt) has it under the heading whereas [this page has it above the section](https://github.com/10gen/cloud-docs/blob/master/source/tutorial/access-encrypted-snapshot.txt) (CMIIW, my rst knowledge is limited)


for the second page example, the [prod page](https://www.mongodb.com/docs/atlas/tutorial/access-encrypted-snapshot/) is showing the metadata whilst the [staged page is not](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/982add268c83b1d3a1e724c176ed109e8dc7e02c/master/cloud-docs/caesarbell/DOP-3698/tutorial/access-encrypted-snapshot/index.html)

think we might want to account for any level metadata, or give an warning/error on the parser side that `.. meta` directives should be used at the top level of each page.

src/utils/get-meta-from-directive.js
```diff
@@ -0,0 +1,25 @@
...
+const grabMetadata = (nodes, target) => {
+  if (!nodes) {
+    return [];
+  }
+  return nodes.filter((c) => {
+    const lookup = c.type === 'directive' ? c.name : c.type;
+    return lookup === target;
+  });
+};
+const getMetaFromDirective = (root, nodes, target) => {
```
mmeigs, line 14: Total nit: I was confused by "root" because I thought it might be a node. Maybe we could name it "type" or something along those lines?

mmeigs (approved):
LGTM, just a naming possibility - take it or leave it :)

seungpark (changes requested):
do we have any (unenforced) rules of `..meta` directives in terms of placement? would be weird to let writers write (as an example) `..meta` directive within `..table` or so. that said, see below comment about support via recursive(?) search through pageNodes

src/utils/get-meta-from-directive.js
caesarbell: @mmeigs no worries, this is a good callout. I was evening trying to remove confusing with the comment above, but that was a fail. I will update the param name to be something clearer.

src/components/DocumentBody.js
rayangler: Good catch!

It seems like, pre-Gatsby v5, the `meta` directive was mapped via the `ComponentFactory` ([old commit](https://github.com/mongodb/snooty/blob/90eed3bc7600a9e0fd763109fee49bec6c6badd4/src/components/ComponentFactory.js#L173)), so it was allowed to be at any level without worrying too much about it. We've had to avoid doing this since Gatsby v5 does not use React Helmet anymore, but I would agree that accounting for `meta` at any level (or maybe the topmost levels?) might be okay for now if the plan is to get this fix out ASAP without needing to change a bunch of content.

I think it would be ideal to enforce the `meta` directive to only be used at a certain level (probably above the heading) in the future to allow us to limit the scope of searching for the directive here. (There are also other solutions, but I don't think we need to get too deep rn)

caesarbell: @rayangler @seungpark I agree ideally it would probably be better if we enforced the `meta` directive to only be used at a certain level. Unless there is some benefit or purpose for being flexible where the `meta` directive can be placed? The end results if I understand correctly would be those `meta` tags would be rendered in the head of the HTML so if the results are the same why not enforce the directives to be above the header in the content?

seungpark: > get this fix out ASAP without needing to change a bunch of content

i think that means we would need to search for `meta` directives anywhere? as per the `ComponentFactory` comment. 

**or, as an immediate fix,** it seems like they are in the head of the rst (which it currently works for in prod), and `within the first section`, as [in the same example as above](https://github.com/10gen/cloud-docs/blob/master/source/atlas-search/atlas-search-overview.txt). we can look for these two levels, but ideally we would account for all levels as we previously did with React Helmet.

[gatsby head](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-head/) seems like it needs each Page to define its Head components, but with our structure of using createPage at once, this doesn't really work well. **ideally** SEO/meta should be at page level in rst so that it translates to page level in snooty

rayangler: I would be in favor of limiting the search for the 2 topmost levels for now to get the fix out. I definitely see a lot more instances of the `meta` directive being used below the title heading rather than above it

src/utils/get-meta-from-directive.js
```diff
@@ -9,5 +11,21 @@ const grabMetaFromDirective = (nodes, target) => {
     return lookup === target;
   });
 };
+const getMetaFromDirective = (type, nodes, target) => {
+  let collectionOfMetadata = [];
```
seungpark, line 15: minor. can init `let collectionOfMetadata = grabMetadata(nodes, target)` and add from sectionNodes after

seungpark (approved):
minor comment below but LGTM 👍

src/components/DocumentBody.js
```diff
@@ -130,15 +131,15 @@ export default DocumentBody;
...
-
-  const twitter = pageNodes.filter((c) => {
-    const lookup = c.type === 'directive' ? c.name : c.type;
-    return lookup === 'twitter';
-  });
+
+  /**
+   * @param root sets the key to look for in the top-level of pageNodes
+   * @param nodes the children of the type found (i.e. section)
+   * @param target the lookup value that we are looking for
```
rayangler, line 138: Wondering if this comment would be better fit above the `getMetaFromDirective` function declaration

src/utils/get-meta-from-directive.js
```diff
@@ -0,0 +1,31 @@
...
+  return nodes.filter((c) => {
+    const lookup = c.type === 'directive' ? c.name : c.type;
+    return lookup === target;
+  });
+};
+const getMetaFromDirective = (type, nodes, target) => {
+  let collectionOfMetadata = [];
+  // check the root level first
+  // and collect the metadata
+  collectionOfMetadata = [...collectionOfMetadata, ...grabMetadata(nodes, target)];
```
rayangler, line 18: Is spreading `collectionOfMetadata` here necessary? Aren't we spreading an empty array?

src/utils/get-meta-from-directive.js
```diff
@@ -0,0 +1,31 @@
...
+  // getting the meta from pageNodes
+  // extract the section to look for the metadata in the second layer
+  const section = nodes.find((node) => node.type === type);
+
+  // get the nested values from section
+  // and look for the metadata to collect
+  const sectionNodes = getNestedValue(['children'], section);
+  collectionOfMetadata = [...collectionOfMetadata, ...grabMetadata(sectionNodes, target)];
```
rayangler, line 26: (Nit) We could always do:

```
collectionOfMetadata.push(...grabMetadata(sectionNodes, target));
```

to avoid needing to spread `collectionOfMetadata` again. (This doesn't matter in the grand scheme of things since there should only really ever be one place where the directive is used per page, but figured it would be neater)

src/components/DocumentBody.js
caesarbell: So there seem to be a few cases where the `meta` directive is placed above the header, and most as you mentioned are right below the header, but I haven't seen any deeper than the second level of the nested data. I wouldn't understand the purpose of going any deeper either, but I could be wrong.

rayangler (approved):

PR #831, opened 2023-06-08
DOP-3651: version context cleanup, use local storage only on front end

seungpark:
### Stories/Links:

DOP-3651

console logs in atlas docs show react hydration errors

### Current Behavior:

[atlas docs](https://www.mongodb.com/docs/atlas/)
[atlas cli docs](https://www.mongodb.com/docs/atlas/cli/upcoming/)

### Staging Links:

[atlas docs](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/seung.park/DOP-3651/)
[atlas cli docs](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/atlas-cli/seung.park/DOP-3651/index.html)

### Notes:
two fixes to address these issues:
- [react error](https://legacy.reactjs.org/docs/error-decoder.html/?invariant=418) -> ie. `build time component does not match client side initial render` this was due to attempting to read local storage for initial values, which the server side does not. removed any local storage assumptions for init values. **note** this results in associated ToC being rendered on client side only (this is fail safe-d with regards to server side fallback values)
- also cleaned up SideNav component, removing unnecessary div wrapper around `SideNavItem<li>`, converting wrong `<button>` wrapper to an `<a>` wrapper:

previously shows nested `<divs>` inside `ul`:
image

on branch (cleaned up with `<li>` elements with no `<button>` wrapper:
image

### DIFF
--- a/src/components/Sidenav/VersionSelector.js
+++ b/src/components/Sidenav/VersionSelector.js
@@ -52,7 +52,7 @@ const selectStyle = css`
   }
 `;
 
-const WrapperStyle = css`
+const wrapperStyle = css`
   margin-left: auto;
 `;
 
@@ -76,7 +76,7 @@ const VersionSelector = ({ versionedProject = '', tocVersionNames = [] }) => {
   }, []);
 
   return (
-    <div onClick={onClick} className={cx(WrapperStyle)}>
+    <div onClick={onClick} className={cx(wrapperStyle)}>
       <Select
         value={activeVersions[versionedProject]}
         className={cx(selectStyle)}
--- a/src/components/Sidenav/VersionSelector.js
+++ b/src/components/Sidenav/VersionSelector.js
@@ -1,5 +1,5 @@
 import React, { useCallback, useContext, useEffect, useState } from 'react';
-import styled from '@emotion/styled';
+import { cx, css } from '@leafygreen-ui/emotion';
 import Select from '../Select';
 import { VersionContext } from '../../context/version-context';
 import { theme } from '../../theme/docsTheme';
@@ -15,7 +15,7 @@ const buildChoices = (branches, tocVersionNames) => {
   return !branches ? [] : branches.filter((branch) => tocVersionNames.includes(branch.gitBranchName)).map(buildChoice);
 };
 
-const StyledSelect = styled(Select)`
+const SelectStyle = css`
   flex: 1 0 auto;
 
   > div:first-of-type {
@@ -42,7 +42,6 @@ const StyledSelect = styled(Select)`
         transform: rotate(180deg);
       }
     }
-    z-index: 3;
   }
 
   span {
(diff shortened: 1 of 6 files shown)


src/context/version-context.js
```diff
@@ -160,11 +155,15 @@ const VersionContextProvider = ({ repoBranches, associatedReposInfo, isAssociate
   );
   // update local storage when active versions change
   useEffect(() => {
+    if (firstLoad) {
+      firstLoad.current = false;
+      return;
+    }
```
mmeigs, line 161: As much as I love this, I'm confused why the boolean `isBrowser` in setLocalValue didn't deal with this issue?

mmeigs (commented):
Just questions for my sake, and one note. Love the refactoring.

src/context/version-context.js
seungpark: this isn't necessary where it was caused

it was mainly caused by [here](https://github.com/mongodb/snooty/pull/831/files#diff-01a0244c29be535118fb0f73acb9a634f297343ee3327f2e3e8d6004b4e5f7feL105-L113) where we initializing state values by reading from local storage (different in build time vs client side render time)

but by making initial state vars not read from local storage, had to make sure it wasn't getting overwritten by server defaults. let me split this up and try to make it a little more intuitive...

seungpark: good catch guys! this was actually done to patch a bad behavior in [toc context](https://github.com/mongodb/snooty/pull/831/files#diff-67fd2be78a4e3484aedf9be9ed69bf5d5a356fa2a503d987deac14a4d2becfbeR53)

removed this and should be updating local storage any time activeVersions change 👍

src/components/Sidenav/VersionSelector.js
```diff
@@ -15,7 +15,7 @@ const buildChoices = (branches, tocVersionNames) => {
   return !branches ? [] : branches.filter((branch) => tocVersionNames.includes(branch.gitBranchName)).map(buildChoice);
 };
 
-const StyledSelect = styled(Select)`
+const SelectStyle = css`
```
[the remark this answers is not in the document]
seungpark: thanks @mmeigs ! updated

src/components/Sidenav/VersionSelector.js
```diff
@@ -53,6 +52,10 @@ const StyledSelect = styled(Select)`
   }
 `;
 
+const WrapperStyle = css`
```
rayangler, line 55: (Nit) Same styling convention that Matt pointed out. Might want to lowercase to `wrapperStyle` since this isn't a component

mmeigs (approved):
LGTM!

rayangler (approved):
LGTM! Thanks for figuring out the issue

src/components/Sidenav/VersionSelector.js
seungpark: thanks for the catch !

PR #838, opened 2023-06-14
changed the default from choosing index 0 to choosing nodejs

bianca-laube:
### Stories/Links:

DOP-2542
Change the default language from C to Node.js 

### Current Behavior:
https://www.mongodb.com/docs/manual/changeStreams/
Screenshot 2023-06-14 at 2 07 47 PM

### Staging Links:
https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/bianca.laube/DOP-2542/changeStreams/
Screenshot 2023-06-14 at 2 08 11 PM

### Notes:
- What I found in the code was that if there was no history of a previous active tab then by default we set it to the first object in the array which was C in this case. 
- So now with the change I made, instead of setting it to index 0, I search for the index of the nodejs. And if the nodejs language is found I set that to the default instead of the C language. But if it is not found then I just set it to the first object in the list. I did this incase we ever use the tab selector element for a list that doesn't contain node.js
- With this change it will also keep the same functionality that is the user has previously selected another language from the drop down the default would be the language the user has previously selected. 
        * the reason I mention this is that you might have to open the links via an incognito tab since that doesn't have the history of what you have previously selected if you want to test it correctly
        * if you want to see another page that uses the same tab selector to make sure the active tab works for using the default as the selected language from before you can also open the transactions tab instead of the change streams tab that is on the left hand side


src/components/Tabs/index.js
```diff
@@ -99,8 +99,14 @@ const Tabs = ({ nodeData: { children, options = {} }, page, ...rest }) => {
 
   useEffect(() => {
     if (!previousTabsetChoice || !tabIds.includes(previousTabsetChoice)) {
-      // Set first tab as active if no tab was previously selected
-      setActiveTab({ name: tabsetName, value: getTabId(children[0]) });
+      const index = children.findIndex((item) => item.options.tabid === 'nodejs');
+      if (index > 0) {
```
seungpark, line 103: minor note! this can be 0 index as well so would be `index > -1`

seungpark (changes requested):
minor index error but LGTM otherwise

great work debugging on your first PR! see you were able to stage some changes as well :) on a minor note, current staging URL's are not working so will have to add `/index.html` to the end of the link

src/components/Tabs/index.js
bianca-laube: perfect! Thanks for the notes! Just made the change :)

seungpark (approved):

mmeigs (approved):

PR #840, opened 2023-06-14
[DOP-3657] moving the fetch to a try catch

bianca-laube:
### Stories/Links:

DOP-3657

### Current Behavior:
https://www.mongodb.com/docs/manual/

### Staging Links:
https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/bianca.laube/DOP-3657/index.html

### Notes:
- There should be no changes in anything you can see, they only change would be to the console for when it catches an error on the fetch call

### DIFF
--- a/src/hooks/use-marian-manifests.js
+++ b/src/hooks/use-marian-manifests.js
@@ -17,7 +17,7 @@ export const useMarianManifests = () => {
         const jsonResult = await result.json();
         setFilters(parseMarianManifests(jsonResult.manifests, propertyMapping));
       } catch (err) {
-        console.log('Failed to fetch the MARIAN_URL.', err);
+        console.error('Failed to fetch the MARIAN_URL.', err);
       }
     }
     const fetchMapping = async () => {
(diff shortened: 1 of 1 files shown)


src/hooks/use-marian-manifests.js
```diff
@@ -12,9 +12,13 @@ export const useMarianManifests = () => {
...
     async function fetchManifests(propertyMapping) {
-      const result = await fetch(MARIAN_URL + `status`);
-      const jsonResult = await result.json();
-      setFilters(parseMarianManifests(jsonResult.manifests, propertyMapping));
+      try {
+        const result = await fetch(MARIAN_URL + `status`);
+        const jsonResult = await result.json();
+        setFilters(parseMarianManifests(jsonResult.manifests, propertyMapping));
+      } catch (err) {
+        console.log('Failed to fetch the MARIAN_URL.', err);
```
rayangler, line 20: (Nit) We might want to use `console.error` here to denote that this is still definitely an error

rayangler (commented):
Nice, can confirm that I no longer see complaints about this error being unhandled locally! Just one nitpick to output the message in stderr, but should be good otherwise

src/hooks/use-marian-manifests.js
bianca-laube: Fixed! Thanks for catching that

rayangler (approved):

PR #843, opened 2023-06-21
DOP-3785: adding a persistent search bar

bianca-laube:
### Stories/Links:

DOP-3785

### Current Behavior:

https://www.mongodb.com/docs/search/?q=test
Screenshot 2023-06-21 at 4 42 22 PM

### Staging Links:

https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/bianca.laube/DOP-3785/search/index.html?q=test
Screenshot 2023-06-21 at 4 49 14 PM

### Notes:

Things Changed:
- Updated header
- The staging link goes automatically to the search page since if you try to go the search page through the home page it leaves the staging link. However, the first word given is taken from that search bar, from the header and automatically loaded to the new persistent search bar. 
- was able to add the new leafygreen-ui search-input for the search bar


src/components/SearchResults/SearchResults.js
```diff
@@ -320,7 +316,17 @@ const SearchResults = () => {
         {!!searchTerm ? (
           <SearchResultsContainer>
             <HeaderContainer>
-              <HeaderText>Search results for "{searchTerm}"</HeaderText>
+              <H1 style={{ color: '#00684A', paddingBottom: '40px' }}> Search Results</H1>
```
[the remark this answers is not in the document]
seungpark: and a final note, we should put this behind the feature flag from @mmeigs 's [PR](https://github.com/mongodb/docs-worker-pool/pull/845) for feature flags :)

seungpark (changes requested):
looks great!! thanks for putting this together @bianca-laube

a couple comments on persisting the search bar on empty results. will need to consult with Product team to verify base case scenarios

src/components/SearchResults/SearchResults.js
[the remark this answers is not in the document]
bianca-laube: - updated it so on empty search the header and search bar will stay
- placeholder of the search input changed to "Search"
- Using H1 from leafy green
new staging link to view changes : 
https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/bianca.laube/DOP-3785/search/index.html?q=$test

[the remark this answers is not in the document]
bianca-laube: new features are pushed behind a feature flag now and old features will show if feature flag is set to false

src/components/SearchResults/SearchResults.js
```diff
@@ -317,96 +322,118 @@ const SearchResults = () => {
...
+                    </>
+                  )}
+                </>
+              )}
+              {showMobileFilters && isTabletOrMobile && <MobileFilters />}
+            </>
+          ) : (
+            <>
+              {!searchResults?.length && (
+                <EmptyResultsContainer>
```
seungpark, line 430: this should be separate of SearchResultsContainer to prevent the grid layout taking over. as an alternative, we can style this to be `grid-area: results;` as well

src/components/SearchResults/SearchResults.js
```diff
@@ -317,96 +322,118 @@ const SearchResults = () => {
...
-                <FiltersContainer>
-                  <FilterHeader>{specifySearchText}</FilterHeader>
-                  <StyledSearchFilters />
-                </FiltersContainer>
+                <H1 style={{ color: '#00684A', paddingBottom: '40px' }}> Search Results</H1>
+                <SearchInput
+                  value={searchField}
+                  placeholder="Search"
+                  onSubmit={(event) => {
+                    setSearchTerm(event.target[0].value);
```
seungpark, line 334: minor nitpick. we should prevent users from entering cleared searches, or if we do allow, show the empty placeholder:

(ss below shows entering cleared input after initial load)
image

seungpark (changes requested):
some minor differences in behavior from whats on prod vs this branch (with flag set to "false") PTAL !

src/components/SearchResults/SearchResults.js
```diff
@@ -317,96 +322,118 @@ const SearchResults = () => {
...
-                      <EmptyResults />
-                    </EmptyResultsContainer>
-                    <FiltersContainer>
-                      <FilterHeader>{specifySearchText}</FilterHeader>
-                      <StyledSearchFilters />
-                    </FiltersContainer>
-                  </>
-                )}
-              </>
+              <HeaderText>Search results for "{searchTerm}"</HeaderText>
```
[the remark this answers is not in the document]
bianca-laube: Just pushed a commit that when flag is false, gets rid of the header if search is empty and added proper grid layout

src/components/SearchResults/SearchResults.js
bianca-laube: Added css with correct grid styling, had to add the css to this instance of EmptyResultsContainer so it wouldn't mess up the css for the emptyResultsContainer on line 409

src/components/SearchResults/SearchResults.js
bianca-laube: now on empty search result for the new persistent search bar is will show the empty result container instead of nothing

src/components/SearchResults/SearchResults.js
```diff
@@ -317,96 +322,123 @@ const SearchResults = () => {
...
-                      searchProperty={searchProperty?.[0]}
-                    />
-                  ))}
-                </StyledSearchResults>
-                <FiltersContainer>
-                  <FilterHeader>{specifySearchText}</FilterHeader>
-                  <StyledSearchFilters />
-                </FiltersContainer>
+                <H1 style={{ color: '#00684A', paddingBottom: '40px' }}> Search Results</H1>
+                <SearchInput
```
seungpark, line 330: minor nitpick on styling.
mockup shows search bar horizontally extends over the search results, out into the search filters on desktop:

image

but on branch, ends with the width of the search results:

image

i actually prefer the second but would verify this is passable or not with design

src/components/SearchResults/SearchResults.js
```diff
@@ -317,96 +322,123 @@ const SearchResults = () => {
...
+                  <FiltersContainer>
+                    <FilterHeader>{specifySearchText}</FilterHeader>
+                    <StyledSearchFilters />
+                  </FiltersContainer>
+                </>
+              ) : (
+                <>
+                  {!searchFinished ? (
+                    <>
+                      <StyledSearchResults>
```
seungpark, line 393: these loading skeletons are not being rendered since `!!searchTerm` always validates to true with the updated logic of searchTerm and searchField

src/components/SearchResults/SearchResults.js
```diff
@@ -317,96 +322,123 @@ const SearchResults = () => {
...
+                </>
+              )}
+              {showMobileFilters && isTabletOrMobile && <MobileFilters />}
+            </>
+          ) : (
+            <>
+              {!searchResults?.length && (
+                <EmptyResultsContainer
+                  css={css`
+                    grid-area: results;
```
seungpark, line 432: now this is taking up a different grid space vs prod (with `GATSBY_TEST_SEARCH_UI` flag set to false)

image

vs prod

image

seungpark (changes requested):
some minor styling nitpicks to keep prod untouched. also not seeing loading states with the new component state variables. PTAL! 🙏

src/components/SearchResults/SearchResults.js
bianca-laube: Confirmed with Iz that the search-bar ending in line with the search results is okay, so no changes needed :)

src/components/SearchResults/SearchResults.js
bianca-laube: updated so now when feature flag is true loading skeleton will show up.  (lines 454-455)

src/components/SearchResults/SearchResults.js
bianca-laube: fixed this issue! took the empty results container out of the search Results container. Then for when the feature flag is TRUE re-put the headercontainer in that section. Not the perfect solution but couldn't find another way for it to work correctly for the grid and the feature flag simultaneously.

src/components/SearchResults/SearchResults.js
```diff
@@ -338,68 +361,107 @@ const SearchResults = () => {
...
           <>
+            {newSearchInput && (
+              <SearchResultsContainer>
+                <HeaderContainer>
+                  <H1 style={{ color: '#00684A', paddingBottom: '40px' }}> Search Results</H1>
+                  <SearchInput
+                    value={searchField}
+                    placeholder="Search"
+                    onSubmit={(event) => {
+                      setSearchResults([]);
```
seungpark, line 454: this is causing a infinite load when searching for same term, due to useEffect not firing

```suggestion
                      const newValue = event.target[0]?.value;
                      if (newValue === searchTerm) return;
                      setSearchResults([]);
```

can also setSearchTerm(newValue) here

seungpark (commented):
this overall LGTM. one minor functional catch below (onSubmit function for SearchInput)

thanks again for putting this together @bianca-laube !

mmeigs (commented):
Looks great! I believe there are a couple unused conditionals! Check me on it though, because there are so many layers to unpack.

Also, I feel like I may have found a bug with our search, but it's not a part of this ticket. Any one letter search seems to hang in loading state oddly. Something to keep an eye on. Bonus points!

src/components/SearchResults/SearchResults.js
```diff
@@ -320,7 +325,25 @@ const SearchResults = () => {
         {!!searchTerm ? (
```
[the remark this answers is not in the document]
bianca-laube: You're right! I was able to get rid of the double redundancy checks. Forgot to get rid of them after I switched up the format for the empty container. Thanks for checking that !

seungpark (approved):
this looks great! minor styling issue below and a possible refactor (not blocking). besides that LGTM !

src/components/SearchResults/SearchResults.js
```diff
@@ -400,6 +429,27 @@ const SearchResults = () => {
...
+                    }}
+                    onChange={(e) => {
+                      setSearchField(e.target.value);
+                    }}
+                  />
+                </HeaderContainer>
+              </SearchResultsContainer>
+            )}
             {!searchResults?.length && (
               <EmptyResultsContainer>
```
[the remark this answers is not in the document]
bianca-laube: Took notes of this comment! Will be implementing in DOP-3786

PR #845, opened 2023-06-27
DOP-3783: Akzidenz Happen

mmeigs:
### Stories/Links:

DOP-3783

### Notes:

Removes all mentions of `Akzidenz`.
Some snapshot testing still mentions it as a fallback, but these are from `leafygreen` components that will not have access to the font and therefore will not occur.
Replaced most mentions of `Akzidenz` with `Euclid Circular A`. Please advise if this is expected behavior.

Furthermore: @asurman is an engineer now and helped with this ticket


schmalliso:
My goodness Matt I should approve this merely on the strength of its PR title

schmalliso (approved):
Maaaaahvelous.

seungpark (approved):
fixer of Akizidenz! 🙌 

verified some of our LG components are still a little behind and referring to [old lg/tokens package](https://github.com/mongodb/leafygreen-ui/blob/%40leafygreen-ui/tokens%401.3.0/packages/tokens/src/fontFamilies.ts) which results in needing insurance 🥁

PR #852, opened 2023-07-11
DOP-3079: de-hardcode DeprecatedVersionSelector

mmeigs:
### Stories/Links:

DOP-3079

### Current Behavior:

[Current Prod](https://www.mongodb.com/docs/legacy/)

### Staging Links:

[Staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/matt.meigs/DOP-3079-name-url/legacy/index.html)

### Notes:

This approach to de-hardcode the `DeprecatedVersionSelector`'s naming and url creation now relies mostly on the `repos_branches` collection in Atlas.

- The `url` and `prefix` properties are used to create each product's url.
- The `displayName` is a newly added property to get rid of the frontend incomplete mapping. This will consolidate into our one source of truth (other than mms, grrr)
-- These display names have been sourced from the `title` in each of the product's `snooty.toml` files.
-- There are three repos in which there are no apparent names. I am checking on these, but I don't think they will be needed. They are `docs-404`, `docs-csfle-merge`, and `devhub-content`

**Before merging**: I have added the property `displayName` to the entries in the `pool_test` database. Before merging, I will add these to the `pool` database.


src/components/DeprecatedVersionSelector.js
```diff
@@ -64,16 +43,49 @@ const isVersioned = (versionOptions) => {
...
+  return true;
+};
+
+// Add mms-docs to reposMap. It does not have a document in repos_branches collection.
+// TODO: Remove when mms-docs is added to repos_branches
+const addOldGenToReposMap = (reposMap) => {
+  return {
+    ...reposMap,
+    mms: {
+      displayName: 'MongoDB Ops Manager',
```
seungpark, line 58: leaving this comment for this branch on `pool_test.repos_branches` where `project = 'docs-mongodb-internal`. this has displayName `MongoDB Manual` vs prod's `MongoDB Server`

src/components/DeprecatedVersionSelector.js
```diff
@@ -87,23 +99,24 @@ const DeprecatedVersionSelector = ({ metadata: { deprecated_versions: deprecated
   const generateUrl = () => {
     // Our current LG button version has a bug where a disabled button with an href allows the disabled
     // button to be clickable. This logic can be removed when LG button is version >= 12.0.4.
-    if (buttonDisabled) {
+    if (buttonDisabled || isEmpty(reposMap) || !hasValidHostName(reposMap[product])) {
       return null;
     }
 
+    const hostName = reposMap[product].url.dotcomprd + reposMap[product].prefix.dotcomprd;
```
seungpark, line 106: minor. can we add a comment for why we're using dotcomprd here instead of env specific variables. ie `legacy sites are not available on dev/stage`. we can even log this message and print the URL it should have been to verify (only in non prod)

src/components/DeprecatedVersionSelector.js
seungpark: similar for `MongoDB CLI` on prod vs `MongoDB Command Line Interface` in pool_test.

likewise for `Ruby MongoDB Driver`

seungpark (commented):
overall LGTM! 👍👍 for doing away with hardcoded values. @mmeigs can you enlighten me on the display name changes? was that from IA or ToC name or...? not sure where the name changes i commented below are coming from

src/components/DeprecatedVersionSelector.js
mmeigs: Gotcha, I get that. yeah, i was hoping to use each environment. :( I'll make that comment!

src/components/DeprecatedVersionSelector.js
mmeigs: Right. These names were originally hard-coded in three years ago. I took these newest names from each product's `snooty.toml` `title` which is where they pull for their title pages. Seems like three or four have changed. Could be worth surfacing with the group if we want to ensure these are the correct values?

seungpark: ah that makes sense! would verify with team for the names. otherwise, largely LGTM besides previous comments 👍

package.json
```diff
@@ -110,6 +110,8 @@
     "https-browserify": "^1.0.0",
     "immer": "^9.0.6",
     "json-schema": "^0.4.0",
+    "lodash.isempty": "^4.4.0",
```
seungpark, line 113: this should be "lodash", not the individual functions

PR #855, opened 2023-07-14
DOP-3079: use lodash package

seungpark:
follow up PR to use lodash package


package.json
```diff
@@ -110,8 +110,7 @@
     "https-browserify": "^1.0.0",
     "immer": "^9.0.6",
     "json-schema": "^0.4.0",
-    "lodash.isempty": "^4.4.0",
-    "lodash.keyby": "^4.6.0",
+    "lodash": "^4.6.0",
     "minimist": "^1.2.6",
```
mmeigs, line 114: I understand that our upgraded Gatsby will code split for us for client-side. But I'm wondering why we should use the entire package when these are the only two functions in use in Snooty atm

PR #860, opened 2023-07-21
DOP-3865

caesarbell:
### Stories/Links:

DOP-3865

### Current Behavior:

[Prod](https://www.mongodb.com/docs/atlas-open-service-broker/)

### Staging Links:

[Staging Content](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/0a60c744fc1433853fb8332575d66dfa4f939b91/master/atlas-open-service-broker/caesarbell/DOP-3865/index.html)

### Notes:

This builds off of `DOP-3865` and handles the edge case where a page is completely EoL'd. This PR is all dependent on the `snooty-parser` change found [here](https://github.com/mongodb/snooty-parser/pull/492).

### DIFF
--- a/src/components/DocumentBody.js
+++ b/src/components/DocumentBody.js
@@ -157,8 +157,21 @@ export const Head = ({ pageContext }) => {
     const stableBranch = repoBranches.branches.find((branch) => {
       return branch.active && branch.isStableBranch;
     });
-    // use the final path prefix
-    canonical = `${siteUrl}/${repoBranches.siteBasePrefix}/${stableBranch.urlSlug}`;
+
+    if (stableBranch) {
+      // if a stable branch is found, use the following canonical tag
+      // which points to the most current version
+      canonical = `${siteUrl}/${repoBranches.siteBasePrefix}/${stableBranch.urlSlug}`;
+    } else {
+      // this means the entire page is EoL'd and a writer should provide the canonical tag
+      let _canonical = `${siteUrl}`;
+      if (metadata.canonical) {
+        _canonical = metadata.canonical;
+      } else {
+        console.warn('The needed canonical is missing from the .toml');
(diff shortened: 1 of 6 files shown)


src/components/DocumentBody.js
```diff
@@ -157,8 +157,21 @@ export const Head = ({ pageContext }) => {
...
+      // if a stable branch is found, use the following canonical tag
+      // which points to the most current version
+      canonical = `${siteUrl}/${repoBranches.siteBasePrefix}/${stableBranch.urlSlug}`;
+    } else {
+      // this means the entire page is EoL'd and a writer should provide the canonical tag
+      let _canonical = `${siteUrl}`;
+      if (metadata.canonical) {
+        _canonical = metadata.canonical;
+      } else {
+        console.warn('The needed canonical is missing from the .toml');
```
rayangler, line 171: (Nit) We might have support for more than just `snooty.toml` in the future. Can we explicitly state that this is the `snooty.toml` file so that people who see this warning are aware of the specific file?

mmeigs, line 171: nit: I'm wondering if there's a clearer way to explain this for the writers. Maybe mentioning that this seems like an EoL'd page and in that case a canonical is needed in the `snooty.toml`?

mmeigs (approved):
LGTM! Just left one nit on possible phrasing!

src/components/DocumentBody.js
caesarbell: @rayangler good call. I wasn't thinking in terms of supporting other files outside of `snooty.toml`. I will make this change.

caesarbell: @mmeigs , will do. I will update this warning to be a bit clearer for the writer.

PR #862, opened 2023-07-31
DOP-3844: Create OAuth flow for API staging instance

rayangler:
### Stories/Links:

DOP-3844

### Staging Links:

[Landing](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/raymundrodriguez/DOP-3844/BRANCH--rc-20230727/index/index.html) - The paths for this staging link are a little atypical, but the data used here was fetched from the source plugin using the staging instance of the Snooty Data API.

### Notes:

For testing purposes, the Autobuilder's client credentials were used. Seems like the client credentials for Snooty are a little wonky right now, so we'll have to make do. The core logic for OAuth was adapted from the [Autobuilder's](https://github.com/mongodb/docs-worker-pool/blob/a43fe88f1b58d70d51e83f408c3366f34546a25d/src/services/cdn.ts). The OAuth flow should be more or less the same. If anything changes with Snooty's (working) client credentials, then we can change whatever's necessary.

I tried using Gatsby Cloud to share a preview link, but since the client credentials for the Autobuilder are tied to IP addresses, there are errors when used on GC machines (but work fine locally if connected to the office VPN, as seen by the staging link above). TBD if Snooty's client credentials will be affected by this.


plugins/gatsby-source-snooty-preview/utils/kanopy-auth.js
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Returns a valid client access token that can be used for machine-machine communication
+ * between a client and a service hosted on Kanopy.
+ * See: https://kanopy.corp.mongodb.com/docs/development/authentication_and_authorization/
+ * @param prevToken
+ */
+const fetchClientAccessToken = async (prevToken) => {
+  let token = prevToken;
+  if (!token) {
+    token = await generateNewAccessToken();
```
seungpark, line 65:
```suggestion
    return generateNewAccessToken();
```

plugins/gatsby-source-snooty-preview/utils/kanopy-auth.js
```diff
@@ -0,0 +1,76 @@
...
+  } else {
+    const decodedValue = JSON.parse(Buffer.from(token.split('.')[1], 'base64').toString('ascii'));
+    // Token is expired, or near expiration
```
seungpark, line 68: out of curiosity, where do we config the expiry of the token? wondering the validity of this assumption / what will happen if we don't refresh. WDYT about testing a route with the token itself?

plugins/gatsby-source-snooty-preview/utils/kanopy-auth.js
```diff
@@ -0,0 +1,76 @@
...
+
+  // Request a new access token from Kanopy's token authentication endpoint
+  const authRequestUrl = `${authUrl}?grant_type=${grantType}&scope=${scope}`;
+  const headers = {
+    authorization: `Basic ${getClientCredentialsHeader()}`,
+    accept: 'application/json',
+    'Content-Type': 'application/x-www-form-urlencoded',
+    'cache-control': 'no-cache',
+  };
+  const res = await fetch(authRequestUrl, { method: 'POST', headers });
```
seungpark, line 42: i see we caught `!res.ok` but would suggest a catch block here to throw as well

plugins/gatsby-source-snooty-preview/utils/kanopy-auth.js
```diff
@@ -0,0 +1,76 @@
+/**
+ * Checks if the configuration property exists, and throws an error otherwise.
+ * @param configProp
+ * @param configType
+ */
+const validateConfigType = (configProp, configType) => {
```
seungpark, line 6: simple but awesome validation 👍

plugins/gatsby-source-snooty-preview/utils/kanopy-auth.js
```diff
@@ -0,0 +1,76 @@
...
+  if (!configProp) {
+    throw new Error(`Missing ${configType} for Snooty Data API access`);
+  }
+};
+
+/**
+ * Generates the authZ token needed for requesting an access token to Kanopy.
+ */
+const getClientCredentialsHeader = () => {
+  const clientId = process.env.OAUTH_CLIENT_ID;
```
seungpark, line 16: generally wondering where these env variables are coming from

plugins/gatsby-source-snooty-preview/utils/kanopy-auth.js
seungpark, line 65:
```suggestion
    return generateNewAccessToken();
```
the same for line 70

seungpark (commented):
thanks for the explaining the new gatsby [node api helpers](https://www.gatsbyjs.com/docs/reference/config-files/node-api-helpers/#cache) 😄 few questions below

plugins/gatsby-source-snooty-preview/gatsby-node.js
```diff
@@ -63,21 +64,32 @@ exports.sourceNodes = async ({ actions, createNodeId, getNode, createContentDige
   let pageCount = 0;
   const fileWritePromises = [];
   const lastFetched = (await cache.get(`lastFetched`)) || 0;
+  const lastClientAccessToken = await cache.get('lastClientAccessToken');
   console.log({ lastFetched });
+
+  // Generate client access token only if trying to access Snooty Data API's staging instance
+  const clientAccessToken = APIBase.includes('.staging') ? await fetchClientAccessToken(lastClientAccessToken) : '';
   let url;
```
mmeigs, line 72: I don't know Gatsby's underpinnings well enough, but within the new utils there are many Errors thrown on undesired outcomes. But I cannot find what catch would be triggered by any of the bubbling out effects of those thrown errors. Should we move this invocation into a try catch? Or does Gatsby handle an overarching catch for us?

plugins/gatsby-source-snooty-preview/utils/kanopy-auth.js
rayangler: The token and its expiration are returned by the auth server, so we don't control it. As with standard OAuth flows, if an expired token is used, Kanopy's CorpSecure proxy will most likely reject access to the staging instance of Snooty Data API.

> WDYT about testing a route with the token itself?

Are you thinking of a unit test with a mock call to a fake auth server?

plugins/gatsby-source-snooty-preview/utils/kanopy-auth.js
rayangler: These were provided by IT. I followed the steps outlined in this [docs section](https://kanopy.corp.mongodb.com/docs/development/authentication_and_authorization/#requesting-client-credentials) to ask for ones particularly for the Snooty frontend.

plugins/gatsby-source-snooty-preview/gatsby-node.js
rayangler: Thanks for flagging! I was working under the assumption that the topmost function (`sourceNodes` in this case) would propagate the error and the process would stop, but it seems like the process does continue. 

I've put this under the `sourceNodes` function's existing try/catch block, and we use Gatsby's [Reporter object](https://www.gatsbyjs.com/docs/reference/config-files/node-api-helpers/#GatsbyReporter) to panic. Any error during the OAuth flow should cause the build to terminate.

This should be the same as:
```js
console.error(error);
process.exit(1);
```

plugins/gatsby-source-snooty-preview/utils/kanopy-auth.js
rayangler: Any error caught here will propagate up to the `sourceNodes` function and result in being printed out and exiting the process (as of the latest commit). I don't think catching the error here would be beneficial since we'll probably just print it and re-throw anyways.

See [this comment](https://github.com/mongodb/snooty/pull/862#discussion_r1280997364) for more details

plugins/gatsby-source-snooty-preview/utils/kanopy-auth.js
rayangler: Done. Modified the function to use `prevToken` directly and removed the unnecessary `else` clause

mmeigs (approved):
LGTM! Also, thanks for deleting the `getNode` parameter!

plugins/gatsby-source-snooty-preview/utils/kanopy-auth.js
seungpark: resolved over the catch function in sourceNodes 👍

plugins/gatsby-source-snooty-preview/utils/kanopy-auth.js
seungpark: discussed offline. these are set per gatsby site, each has to be updated accordingly. thanks @rayangler !

following up on that, will have to copy env credentials across sites

seungpark (approved):
LGTM . nice find on `reporter` helper. pls input comment on `decodedValue` from token before merge :)

plugins/gatsby-source-snooty-preview/utils/kanopy-auth.js
seungpark: discussed offline. was mentioning a dummy request to test the token validity vs guessing expiration, but decoded token shows auth object with exp attribute

PR #867, opened 2023-08-16
DOP-3540 🚚  moves from stitch to realm

caesarbell:
### Stories/Links:

DOP-3540

### Current Behavior:

N/A

### Staging Links:

[Staging Comment](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/a76f3509d80187ab29be4c2e4999920afe92abf8/master/landing/caesarbell/DOP-3540/index.html)

### Notes:

This completes the attempt to move from stitch to realm. Because of the codebase's successful structure, this feat was possible with minimal code change.


src/components/Banner/SiteBanner.js
```diff
@@ -10,7 +10,7 @@ import { fetchBanner } from '../../utils/realm';
 
 const getBannerSource = (src) => {
   if (src == null || src === '') return null;
-  const srcUrl = `${SNOOTY_STITCH_ID}.mongodbstitch.com/${src}`;
+  const srcUrl = `${SNOOTY_REALM_APP_ID}.mongodbstitch.com/${src}`;
   return `https://${normalizePath(srcUrl)}`;
 };
```
mmeigs, line 15: I have to say I'm completely ignorant of what this url is, but does the domain need to be changed too? from mongodbstitch.com to something else? Or is this still the dependency

src/utils/realm.js
```diff
@@ -1,60 +1,52 @@
...
-      console.error('Failed to log in', err);
-      rej(err);
+const loginAnonymous = async () => {
+  try {
+    if (!app.currentUser) {
+      await app.logIn(Realm.Credentials.anonymous());
     }
-  });
-  return loginDefer;
+    return app.currentUser;
```
mmeigs, line 11: Doesn't seem like we use this below, but I'm also fine with keeping it in. Feels right?

package.json
```diff
@@ -115,6 +116,7 @@
     "mongodb-stitch-browser-sdk": "^4.8.0",
```
seungpark, line 116: are we not removing these two deprecated packages? seems like we are still using both within feedbackWidget items?

src/utils/realm.js
```diff
@@ -1,60 +1,52 @@
...
+import { SNOOTY_REALM_APP_ID } from '../build-constants';
 
-const config = {
-  id: SNOOTY_STITCH_ID,
-};
-const app = new Realm.App(config);
-let loginDefer;
+const app = new Realm.App({ id: SNOOTY_REALM_APP_ID });
 
-const loginAnonymous = () => {
```
seungpark, line 10: this was to prevent multiple login attempts. can we include this as well? otherwise, this will introduce another slew of login attempts for the same realm client. https://jira.mongodb.org/browse/DOP-3473

looking back, we can also add a `finally` block to clear the `loginDefer` var

seungpark (commented):
CMIIW but i think we are supposed to remove both deprecated stitch packages from snooty. seems to be some remnants, specifically for feedback widget

src/components/Banner/SiteBanner.js
rayangler: It should be good to keep the url as-is. The hostname can be visible in our Atlas App Service's [hosting page](https://realm.mongodb.com/groups/5bad1d3d96e82129f16c5df3/apps/5d41ef5e2de84772fc7c2de2/hosting/assets?prefix=/)

caesarbell: I figure we would not want to update or change the hostname in our Atlas Apps Services. If I made an incorrect assumption I apologies. The value is still `snooty-koueq`. To keep things more descriptive around the change I only updated the name of the constant from build-constants.

package.json
caesarbell: @seungpark good catch. I didn't even catch that the feedback widget were using stitch as well.

src/utils/realm.js
caesarbell: @mmeigs sorry if I misunderstanding, are you referring to the catch logic?

src/utils/realm.js
caesarbell: @seungpark gotcha, okay I will add this back in and provide a comment for future referencing.

src/utils/setup/init-realm.js
```diff
@@ -0,0 +1,13 @@
+const Realm = require('realm-web');
+const { SNOOTY_REALM_APP_ID } = require('../../build-constants');
+
+const initRealm = async () => {
```
rayangler, line 4: I know that our Node files on Gatsby currently require `require` for modules instead of ES6 imports, but is there a reason to separate this file from `src/utils/realm.js`? Since we're sticking with `realm-web` and aren't using two separate packages anymore, are we able to consolidate the two files now?

src/components/Widgets/FeedbackWidget/realm.js
```diff
@@ -1,29 +1,28 @@
...
 // User Authentication & Management
 export async function loginAnonymous() {
-  if (!app.auth.user) {
-    const user = await app.auth.loginWithCredential(new AnonymousCredential());
+  if (!app.currentUser) {
+    const user = await app.logIn(Realm.Credentials.anonymous());
     return user;
   } else {
-    return app.auth.user;
+    return app;
```
rayangler, line 14: Since `loginAnonymous` is called with the expectation of a user returned, should we be returning `app.currentUser` instead of just `app`?

gatsby-node.js
```diff
@@ -202,7 +202,7 @@ exports.sourceNodes = async ({ actions, createContentDigest, createNodeId }) =>
     db = manifestDocumentDatabase;
   } else {
     console.log('Loading documents from stitch');
```
rayangler, line 204: Might be good to update the comments/logs in this file to say "Realm" rather than "Stitch"

rayangler (commented):
Thanks for working on this! Just a few comments from me rn

src/utils/setup/init-realm.js
caesarbell: @rayangler  good question, yes we are actually able to consolidate the two files, or should be at least because we are using `realm-web`. I was just being mindful of how much refactoring, but it makes sense to consolidate when we can. I just wasn't prepared to make any assumptions.

src/components/Widgets/FeedbackWidget/realm.js
caesarbell: @rayangler  good catch. You are correct, we should be returning in the `loginAnonymous`  the `app.currentUser` rather than `app`.

caesarbell: @rayangler good catch, you are correct about that. We should be returning the `app.currentUser` rather than `app`. Thanks for catching that.

src/utils/setup/init-realm.js
caesarbell: @rayangler when you say to separate the two files, are you referring to realm creating the anonymous use in the `src/utils/realm.js` and in the `src/components/Widgets/FeedbackWidget/realm.js`? Would you like to see if we can consolidate these and just use the export function from the latter?

seungpark: @rayangler do you mean consolidating `src/utils/realm.js` and `src/init/DocumentDatabase` and this file `src/utils/setup/init-realm.js`?

like you mentioned, one the first is using ES6 exports and imports (for front end), and the latter two are used (for server load).

i think this makes it scope creep a little bit and perhaps we can make it another task to convert to es6? WDYT

seungpark (approved):
largely LGTM! lets either keep the discussion of consolidation open, or drop in another ticket to move to ES6 on server side of snooty

src/utils/setup/init-realm.js
rayangler: > do you mean consolidating src/utils/realm.js and src/init/DocumentDatabase and this file src/utils/setup/init-realm.js?

Just `src/utils/realm.js` and `src/utils/setup/init-realm.js`. I think it would be good to ultimately have a unified way to at least connect to the Snooty app service, instead of having the same/similar logic be in two different places. We have several functions that use CommonJS exports that are used both server side and within the frontend.

I'm okay with considering this as scope creep and dealing with this another time! For what it's worth, there's a [ticket](https://jira.mongodb.org/browse/DOP-3009) that handles conversion to ES6 imports in a more general scale.

rayangler (approved):

src/utils/setup/init-realm.js
caesarbell: @rayangler gotcha. Okay, this makes sense. I think this can fall into the ticket you shared, but I do agree that it is a better approach to unify the way we connect to the Snooty app service.

src/utils/realm.js
mmeigs: Ignore this! Lots of things have changed, and I might have missed where a return value was used. Now it definitely is!

PR #870, opened 2023-08-22
DOP-3787 b

seungpark:
### Stories/Links:

DOP-3787

### Current Behavior:

### Staging Links:


src/components/SearchResults/SearchResults.js
```diff
@@ -298,6 +299,16 @@ const SearchResults = () => {
     if (q === '' || q === undefined) {
       if (!firstRenderComplete) setFirstLoadEmpty(true);
       setSearchFinished(true);
+      return;
```
mmeigs, line 302: I'm wondering if we still need to do any of the setting of the state later on in this useEffect? ie setSearchTerm, setSearchField, setSearchFilter? This might be why I'm not seeing filters? (total guess)

src/components/SearchResults/SearchResults.js
```diff
@@ -314,29 +325,68 @@ const SearchResults = () => {
     }
     if (newSearchInput) setSearchFinished(false);
     const fetchNewSearchResults = async () => {
+      let searchParams = new URLSearchParams(window.location.search);
```
mmeigs, line 328: nit: this can be a constant!

src/components/SearchResults/SearchResults.js
```diff
@@ -314,29 +325,68 @@ const SearchResults = () => {
...
+      const searchParams = new URLSearchParams(window.location.search);
+      const currentPage = parseInt(searchParams.get('page'));
+      const newPage = isForward ? currentPage + 1 : currentPage - 1;
+      if (newPage < 1) {
+        return;
+      }
+      searchParams.set('page', newPage);
+      const newRelativePathQuery = window.location.pathname + '?' + searchParams.toString();
+      window.history.replaceState(null, '', newRelativePathQuery);
+      // setSearchResults([]);
```
mmeigs, line 378: Do we want this or can we delete?

src/components/SearchResults/SearchResults.js
```diff
@@ -314,29 +325,68 @@ const SearchResults = () => {
...
+      setSearchFinished(false);
+      const result = await fetch(searchParamsToURL(searchTerm, searchFilter, newPage));
+      const resJson = await result.json();
+      if (!!resJson?.results) {
+        setSearchResults(resJson.results);
+      }
+      setSearchFinished(true);
```
mmeigs, line 385: If we get rid of the commented our setSearchResults above, we probably need an else here to set it to an empty array? Correct me if I'm wrong!

src/components/SearchResults/SearchResults.js
```diff
@@ -452,6 +503,19 @@ const SearchResults = () => {
...
+                  {
+                    <>
+                      <Pagination
+                        currentPage={parseInt(new URLSearchParams(window.location.search).get('page'))}
+                        // TODO: add count after facet meta query
+                        onForwardArrowClick={onPageClick.bind(null, true)}
+                        onBackArrowClick={onPageClick.bind(null, false)}
+                        shouldDisableBackArrow={parseInt(new URLSearchParams(window.location.search).get('page')) === 1}
+                        // TODO: should disable if at max count from meta query
+                        shouldDisableForwardArrow={searchResults?.length && searchResults?.length < 10}
```
mmeigs, line 515: total nit: Since you already checked it's truthiness, you can remove the second optional chaining `?` because it's definitely present.

src/components/SearchResults/SearchResults.js
```diff
@@ -452,6 +503,19 @@ const SearchResults = () => {
                       searchProperty={searchProperty?.[0]}
                     />
                   ))}
+                  {
+                    <>
+                      <Pagination
+                        currentPage={parseInt(new URLSearchParams(window.location.search).get('page'))}
+                        // TODO: add count after facet meta query
+                        onForwardArrowClick={onPageClick.bind(null, true)}
+                        onBackArrowClick={onPageClick.bind(null, false)}
```
mmeigs, line 512: Wait, fun!! I never use bind anymore and just go with arrow functions, but I looked it up and bind is more performant! til

src/components/SearchResults/SearchResults.js
```diff
@@ -374,9 +424,10 @@ const SearchResults = () => {
                 }}
               />
               <ResultTag style={{ paddingTop: '10px' }}>
-                <Overline style={{ paddingTop: '11px', paddingRight: '8px' }}>
+                {/* TODO: add number of results from metadata */}
+                {/* <Overline style={{ paddingTop: '11px', paddingRight: '8px' }}>
                   {!firstLoadEmpty && <>{searchResults?.length ? searchResults.length : '0'} RESULTS</>}
-                </Overline>
+                </Overline> */}
```
mmeigs, line 430: Do we need?

mmeigs (commented):
Looks great! I have two things:
- This is probably a configuration thing on my end, but I'm not seeing any "Category" options in the filters doing this locally on any of my searches.
- Total nit: Is there a way on "page change" to still have the pagination component stay where it is during the loading? This would help with the flash of the loading screen between pages.

src/components/SearchResults/SearchResults.js
seungpark: the filters are missing due to /status route not being returned from local search transport! sorry, should've specified that local search server should be started with `--load-manifests` option for the /status route to be functional.

ie. `npm run search-transport -- --load-manifests`

seungpark: but related, this PR is mixing usage of `window.location` and [useLocation](https://reach.tech/router/api/useLocation). going to try to consolidate these

src/components/SearchResults/SearchResults.js
```diff
@@ -314,29 +325,68 @@ const SearchResults = () => {
...
+      const newPage = isForward ? currentPage + 1 : currentPage - 1;
+      if (newPage < 1) {
+        return;
+      }
+      searchParams.set('page', newPage);
+      const newRelativePathQuery = window.location.pathname + '?' + searchParams.toString();
+      window.history.replaceState(null, '', newRelativePathQuery);
+      // setSearchResults([]);
+      setSearchFinished(false);
+      const result = await fetch(searchParamsToURL(searchTerm, searchFilter, newPage));
```
rayangler, line 381: (Non-blocking; optional) It might be good to reuse the new `fetchNewSearchResults` function and just pass the page as a variable

src/components/SearchResults/SearchResults.js
seungpark: good catch! removed

src/components/SearchResults/SearchResults.js
```diff
@@ -314,29 +325,68 @@ const SearchResults = () => {
...
+      const currentPage = parseInt(searchParams.get('page'));
+      const newPage = isForward ? currentPage + 1 : currentPage - 1;
+      if (newPage < 1) {
+        return;
+      }
+      searchParams.set('page', newPage);
+      const newRelativePathQuery = window.location.pathname + '?' + searchParams.toString();
+      window.history.replaceState(null, '', newRelativePathQuery);
+      // setSearchResults([]);
+      setSearchFinished(false);
```
rayangler, line 380: Do we need this state change? Looks like we're setting to false right before setting to true

src/components/SearchResults/SearchResults.js
seungpark: converted to a default fallback if no res 👍

src/components/SearchResults/SearchResults.js
seungpark: left it as a TODO as part of facet metadata search!

src/components/SearchResults/SearchResults.js
seungpark: CMIIW but i didn't see [any param passed to these events](https://github.com/mongodb/leafygreen-ui/blob/main/packages/pagination/src/Pagination/Pagination.tsx#L144C22-L157), and bind to make it reusable! :D

src/components/SearchResults/SearchResults.js
```diff
@@ -452,6 +502,19 @@ const SearchResults = () => {
...
+                    <>
+                      <Pagination
+                        currentPage={parseInt(new URLSearchParams(search).get('page'))}
+                        // TODO: add count after facet meta query
+                        onForwardArrowClick={onPageClick.bind(null, true)}
+                        onBackArrowClick={onPageClick.bind(null, false)}
+                        shouldDisableBackArrow={parseInt(new URLSearchParams(search).get('page')) === 1}
+                        // TODO: should disable if at max count from meta query
+                        shouldDisableForwardArrow={searchResults?.length && searchResults.length < 10}
+                      ></Pagination>{' '}
```
rayangler, line 515: Is this whitespace needed?

src/components/SearchResults/SearchResults.js
```diff
@@ -298,11 +300,21 @@ const SearchResults = () => {
...
+      return;
+    }
+    if (newSearchInput) {
+      const searchParams = new URLSearchParams(search);
+      const page = parseInt(searchParams.get('page'));
+      if (!page && !!search) {
+        searchParams.set('page', '1');
+      }
+      const newRelativePathQuery = pathname + '?' + searchParams.toString();
+      navigate(newRelativePathQuery);
```
rayangler, line 312: `navigate` is relative to the path prefix of the site. This can cause `pathname` (which includes said path prefix) to be repeatedly added if not ran locally (try with `npm run build && npm run serve`).

I'd recommend excluding the `pathname` altogether as `navigate` seems to understand what to do with query params (based on the example on [this page](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-link/#recommendations-for-programmatic-in-app-navigation))

rayangler (commented):
There's a weird behavior where if you go on page 2 of the search results with a search filter and you refresh or use the link with the page query param directly (such as `localhost:8000/search?q=test&page=2&searchProperty=atlas-master&searchVersion=Latest`), it automatically goes back to page 1. Not a big deal, but figured it might be worth pointing out in case this is definitely undesired

EDIT: Updated the URL to include search filter. This doesn't seem to happen without?

src/components/SearchResults/SearchResults.js
mmeigs: Oh, this is great!!

src/components/SearchResults/SearchResults.js
seungpark: removed the commented code. we set it to false, then await for results, then set to true after now

src/components/SearchResults/SearchResults.js
seungpark: thanks for the tip!

seungpark:
@mmeigs @rayangler one thing i noticed while going through the comments, was that we are doubling down on state + URL; we are storing and updating state of searchTerm, searchProperty within the component when these are persisted into the URL and should be referenced from there. integrating this change so that all search fields are read from, and updated to the URL

src/components/SearchResults/SearchResults.js
seungpark: removed!

seungpark:
> There's a weird behavior where if you go on page 2 of the search results with a search filter and you refresh or use the link with the page query param directly (such as `localhost:8000/search?q=test&page=2&searchProperty=atlas-master&searchVersion=Latest`), it automatically goes back to page 1. Not a big deal, but figured it might be worth pointing out in case this is definitely undesired
> 
> EDIT: Updated the URL to include search filter. This doesn't seem to happen without?

So this was happening due to the [comment i mentioned earlier](https://github.com/mongodb/snooty/pull/870#issuecomment-1690449938) about double scoping. I couldn't point it down to a specific cause, but I did a sweep of all 
1) mixed usage of `useLocation` and references to `window.location.search`  -> consolidated to useLocation
2) mixed usage of `useState` and references to URL -> removed useState in favor of reference to URL

finally, I removed the Context definition within SearchResults, and moved this all to the context, and this context will be available to all components of the Search UI.

> note: followup improvement to scope the context to only the search results page.

src/components/SearchResults/SearchResults.js
rayangler: Oh, duh, forgot the intention of state change here. Thanks!

src/components/SearchResults/SearchFilters.js
```diff
@@ -26,75 +26,71 @@ const SearchFilters = ({ manuallyApplyFilters = false, onApplyFilters, ...props
...
+    selectedVersion,
     setSelectedVersion,
     setSelectedCategory,
   } = useContext(SearchContext);
 
-  // Current category and version for dropdown. If manuallyApplyFilter === true, selectedCategory + selectedVersion
+  // Current selectedCategory and selectedVersion for dropdown. If manuallyApplyFilter === true, selectedCategory + selectedVersion
   // will not be set automatically.
   const [categoryChoices, setCategoryChoices] = useState([]);
-  const [category, setCategory] = useState(null);
```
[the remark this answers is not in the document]
rayangler: For what it's worth, `selectedCategory` differed from `category` to accommodate the behavior of the mobile search filters, where the filters set in the dropdowns wouldn't actually be applied until the "apply filters" button was hit. With the latest changes, seems like this is no longer true.

No big deal if we want to change this behavior as part of this PR, but I just wanted to make sure that this change in behavior is intended and/or desired.

src/components/SearchResults/SearchContext.js
```diff
@@ -15,4 +19,73 @@ const SearchContext = createContext({
...
+  const searchFilter = searchParams.get('searchProperty');
+
+  // state vars to derive selected category and versions in dropdown
+  // changes reflected in UI, not necessarily in URL
+  const [selectedCategory, setSelectedCategory] = useState(null);
+  const [selectedVersion, setSelectedVersion] = useState(null);
+  const [showMobileFilters, setShowMobileFilters] = useState(false);
+
+  // navigate changes and store state in URL
+  const onSearchChange = ({ searchTerm, searchFilter, page }) => {
```
rayangler, line 39: Nice. This is a good function to have

rayangler (commented):
This is great! I think the latest changes pretty much resolve the issues from before, and I think the logic between states and query params is looking good.

Just wanted to flag one non-blocking behavior for mobile search filters that is altered by the recent change.

seungpark:
> This is great! I think the latest changes pretty much resolve the issues from before, and I think the logic between states and query params is looking good.
> 
> Just wanted to flag one non-blocking behavior for mobile search filters that is altered by the recent change.

thanks for the catch and explaining the double state var! i actually removed the test that checked this, but the behavior was just added back (along with mobile state trackers on the filters), as well as the test! PTAL

rayangler (approved):

seungpark:
im putting a DNM label on this until this [PR on search server](https://github.com/mongodb/docs-search-transport/pull/63) side gets released. will be merging once a new tag (presumably v0.1.8) is released

PR #873, opened 2023-08-30
DOP-3982: Fix internal navigation in Gatsby plugin

mmeigs:
### Stories/Links:

DOP-3982

### Notes:

Currently each page's internal nav (back and forward) buttons are just a forward button to go back to "What is MongoDB Atlas?" example: https://preview-mongodbmmeigs.gatsbyjs.io/cloud-docs/matt-build/cloud-providers-regions/

This was caused by the slug passed to each page having a forward slash prepended which does not match the toctree mappings.

### DIFF
--- a/plugins/gatsby-source-snooty-preview/gatsby-node.js
+++ b/plugins/gatsby-source-snooty-preview/gatsby-node.js
@@ -307,6 +307,7 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
         pagePath = node.page_id;
       }
       let slug = node.page_id;
+      // Slices off leading slash to ensure slug matches an entry within the toctreeOrder and renders InternalPageNav components
       if (slug !== '/' && slug[0] === '/') slug = slug.slice(1);
 
       createPage({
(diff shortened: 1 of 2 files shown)


plugins/gatsby-source-snooty-preview/gatsby-node.js
```diff
@@ -306,13 +306,15 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
       } else {
         pagePath = node.page_id;
       }
+      let slug = node.page_id;
+      if (slug !== '/' && slug[0] === '/') slug = slug.slice(1);
```
rayangler, line 310: (Super duper nit) It might be good to just include a comment above explaining why we would need to perform this slice. This important operation may be easy to overlook / lose context on

rayangler (approved):
Nice fix! And thanks for looking into this

plugins/gatsby-source-snooty-preview/gatsby-node.js
mmeigs: Great point

PR #880, opened 2023-09-05
DOP-3996: Chatbot server url

mmeigs:
### Stories/Links:

DOP-3996

### Notes:

The chatbot team would like our non-prod environments to point the Chatbot towards their staging link. This uses the SNOOTY_ENV to determine this.

Would love a second pair of eyes and confirmations on the fact that SNOOTY_ENV's values could all be 'production' and 'dotcomprd'

### DIFF
--- a/src/components/ChatbotUi.js
+++ b/src/components/ChatbotUi.js
@@ -4,7 +4,11 @@ import styled from '@emotion/styled';
 import { css } from '@emotion/react';
 import { theme } from '../theme/docsTheme';
 import 'react-loading-skeleton/dist/skeleton.css';
-import { CHATBOT_SERVER_BASE_URL } from '../constants';
+
+const CHATBOT_SERVER_BASE_URL =
+  process.env.SNOOTY_ENV === 'dotcomprd'
+    ? 'https://knowledge.corp.mongodb.com/api/v1'
+    : 'https://knowledge.staging.corp.mongodb.com/api/v1';
 
 const SKELETON_BORDER_RADIUS = '12px';
 
--- a/src/components/ChatbotUi.js
+++ b/src/components/ChatbotUi.js
@@ -7,7 +7,7 @@ import 'react-loading-skeleton/dist/skeleton.css';
 
 const CHATBOT_SERVER_BASE_URL =
   process.env.SNOOTY_ENV === 'production' || process.env.SNOOTY_ENV === 'dotcomprd'
-    ? 'https://knowledge.corp.mongodb.com/api/v1'
+    ? 'https://knowledge.mongodb.com/api/v1'
     : 'https://knowledge.staging.corp.mongodb.com/api/v1';
 
 const SKELETON_BORDER_RADIUS = '12px';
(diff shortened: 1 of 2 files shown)


src/components/ChatbotUi.js
```diff
@@ -5,6 +5,11 @@ import { css } from '@emotion/react';
 import { theme } from '../theme/docsTheme';
 import 'react-loading-skeleton/dist/skeleton.css';
 
+const CHATBOT_SERVER_BASE_URL =
+  process.env.SNOOTY_ENV === 'production' || process.env.SNOOTY_ENV === 'dotcomprd'
+    ? 'https://knowledge.corp.mongodb.com/api/v1'
```
rayangler, line 10: I don't think the `corp` is desired here. I believe this URL would redirect to the main MongoDB site as-is

src/components/ChatbotUi.js
```diff
@@ -5,6 +5,11 @@ import { css } from '@emotion/react';
 import { theme } from '../theme/docsTheme';
 import 'react-loading-skeleton/dist/skeleton.css';
 
+const CHATBOT_SERVER_BASE_URL =
+  process.env.SNOOTY_ENV === 'production' || process.env.SNOOTY_ENV === 'dotcomprd'
```
rayangler, line 9: > Would love a second pair of eyes and confirmations on the fact that SNOOTY_ENV's values could all be 'production' and 'dotcomprd'

Double-checked our current task definitions and it seems like these are the only 2 prod-based values we'd expect

src/components/ChatbotUi.js
mmeigs: Right, it does seem to. This is the link referenced in the ticket. I'll check with Ben or Chris!

rayangler: The URL without the `corp` is the base URL mentioned in the [npm package](https://www.npmjs.com/package/mongodb-chatbot-ui), so I was just following that

rayangler (approved):

src/components/ChatbotUi.js
mmeigs: Thanks! Had to check with a couple people, but you were right!

src/components/ChatbotUi.js
rayangler, line 9: I know this PR is already approved and merged, but I just remembered that I should have asked this: 

Does the chatbot team want us to use the staging API for GitHub staging jobs and the production API for only production deployments? If so, we might need to update this logic to exclude `process.env.SNOOTY_ENV === 'production'`. @caesarbell might have some insight.

mmeigs: oo @rayangler just getting fully caught up on my email and saw this! I think I might be confused at what SNOOTY_ENV === 'production' would mean in this case? Is writers staging using the value "production"?

rayangler: > Is writers staging using the value "production"?

Yes, I believe so, based on its value in this [task definition](https://us-east-2.console.aws.amazon.com/ecs/v2/task-definitions/docs-worker-pool-prd/39/json?region=us-east-2). `SNOOTY_ENV` can be seen as representative of the overall environment that Snooty is being ran within the context of the Autobuilder, with "production" being "prd" and "dotcomprd" being "dotcomprd".

mmeigs: Ok, great catch. Thank you! I'll open up a PR!

PR #881, opened 2023-09-06
DOP-3855: Adding roles for LeafyGreen icons

mayaraman19:
### Stories/Links:

[DOP-3855](https://jira.mongodb.org/browse/DOP-3855)

### Current Behavior:

Added role support for LeafyGreen icons. If ``:icon-lg:`Glyph` `` is specified in the docs source material, that icon will be parsed and reflected in the frontend. 

[Link to snooty-parser PR](https://github.com/mongodb/snooty-parser/pull/512)

### Staging Links:

[Staging environment.](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/maya/DOP-3855/configure-alerts/index.html) This links to the Configure Alerts page, which I used for testing by adding some random LeafyGreen icons (Database, GlobeAmericas, ActivityFeed using the roles ``:icon-lg:`Database` ``, ``:icon-lg:`GlobeAmericas` ``, and ``:icon-lg:`ActivityFeed` `` respectively).


src/components/Roles/Icon.js
```diff
@@ -33,6 +34,8 @@ const RoleIcon = ({ nodeData: { target, name } }) => {
     return <i className={`charts-icon-${target} charts-icon`}></i>;
   } else if (name === 'icon-mms') {
     return <i className={`mms-icon-${target} mms-icon`}></i>;
+  } else if (name === 'icon-lg') {
+    return <Icon glyph={target} className="lg-icon" />;
```
branberry, line 38: It looks like the `glyph` property is being overridden by using the `className` property. I took a screenshot of the `class` property for the DB icon in the staging link you provided:
image

Based on the [documentation for LeafyGreen icons](https://github.com/mongodb/leafygreen-ui/tree/main/packages/icon), it appears the the `className` isn't a property that can be provided, but I see why you have added it so that the icon doesn't float above the text. It looks like in the first example, the css class is autogenerated, and is prefixed with `leafygreen-ui-`. This leads me to believe that the class name is based off of the `glyph` property. 

In terms of moving forward, I think we might need to remove the `className` property here, but I am also not very well versed with styling in snooty. I would defer to @rayangler @seungpark @mmeigs on this.

 Here's the example from the link:

## Example

```js

const SomeComponent = () => <Icon glyph="Plus" fill="#FF0000" />;
```

**Output HTML**

```html
<svg
  width="16"
  height="16"
  xmlns="http://www.w3.org/2000/svg"
  viewBox="0 0 16 16"
  class="leafygreen-ui-1gvnm9k"
  role="img"
  aria-label="Beaker Icon"
>
.....
</svg>
```

branberry (commented):
Hey Maya! This looks really good, I left one comment about the styling of the icon.

src/components/Roles/Icon.js
rayangler: > It looks like in the first example, the css class is autogenerated, and is prefixed with leafygreen-ui-. This leads me to believe that the class name is based off of the glyph property.

I don't believe the class name has any relation to the glyph property, especially since it's an SVG. It is correct that the `leafygreen-ui-` class names are generated from css, but I don't think adding the `lg-icon` class name directly has any negative impact in this case. (LG's [examples](https://www.mongodb.design/component/icon/example/) show that by default, icons don't even have an autogenerated class name due to the lack of css).

IMO, we can either keep things as-is for consistency with most of the other icons, or we can insert the css directly similar to the example below instead of relying on traditional class names.

```js
import { cx, css } from '@leafygreen-ui/emotion';

const lgIconCss = css`
  // foo
`;

const RoleIcon = ({ target }) => {
  return <Icon glyph={target} className={cx(lgIconCss)} />;
};
```

branberry: Ok gotcha! Thank you for pointing that out, what you're saying makes sense. This looks all good then 👍

branberry (approved):
LGTM!

PR #882, opened 2023-09-06
DOP-3961

seungpark:
### Stories/Links:

DOP-3961

### Current Behavior:

### Staging Links:


src/utils/search-params-to-url.js
```diff
@@ -8,3 +8,19 @@ export const searchParamsToURL = (searchQuery, searchFilters, pageNumber = 1, is
...
 };
+/**
+ *
+ * @param {string} searchQuery
+ * @param {string} searchFilters
+ * @param {{key}[]} facetSelections
+ */
+export const searchParamsToMetaURL = (searchQuery, searchFilters, facetSelections) => {
+  let queryParams = `?q=${searchQuery}${searchFilters ? `&searchProperty=${searchFilters}` : ''}`;
+  // for (const facetSelection of facetSelections) {
```
rayangler, line 19: (Nit) If we're planning to keep this code commented, it might be good to leave a comment so that nobody else accidentally deletes this until fully implemented

src/components/SearchResults/SearchResults.js
```diff
@@ -431,7 +450,7 @@ const SearchResults = () => {
                 <>
                   <Pagination
                     currentPage={parseInt(new URLSearchParams(search).get('page'))}
-                    // TODO: add count after facet meta query
+                    numTotalItems={searchCount}
```
rayangler, line 453: Seeing this weird behavior with the `Pagination` component. Do we know if this is an issue with the component itself?

Screen Shot 2023-09-07 at 2 48 56 PM

seungpark: good catch! fixed with last commit for defaulting to `page 1` for pagination. (previously was reading from url search params, which does require page to be included )

src/utils/search-params-to-url.js
seungpark: good catch! done

src/components/SearchResults/SearchResults.js
```diff
@@ -387,26 +414,18 @@ const SearchResults = () => {
         )}
 
         {/* empty search results */}
-        {!!searchFinished && !searchResults.length && (
+        {!isFirstLoad && !!searchFinished && !searchResults.length && (
```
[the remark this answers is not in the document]
seungpark: good catch on searchResults nullish check ! 👍

PR #883, opened 2023-09-06
DOP-4009: Avoid fetching TOC data for preview builds

rayangler:
### Stories/Links:

DOP-4009

### Current Behavior:

[prd Gatsby Cloud site](https://preview-mongodbkanchanamongodb.gatsbyjs.io/cloud-docs/DOCSP-32895/atlas-search/knn-beta/) - Notice how the side nav does not show the "knn beta" page

### Staging Links:

[stg Gatsby Cloud site](https://preview-mongodbrayanglerstg.gatsbyjs.io/cloud-docs/test-master/atlas-search/knn-beta/) - The side nav now shows the expected content and no longer shows the version selector for associated products (this behavior matches current staging builds).

### Notes:
Seems like the TOC is fetching data on initial load of the page, which can result in incorrect TOC data, especially if the current changes of the branch don't match up with the TOC metadata being fetched from the db.

The `parserBranch` variable, which is tied to the declared branch that the site is for, is currently used when fetching metadata. This makes sense for builds that use local manifests (like production deployments) since the site being built will only ever consist of content for one branch at a time. Although this PR does make the logic compatible with the ≥1 branches that a Gatsby Cloud site can support, I decided to also avoid the fetching logic altogether since (1) it acts as an explicit guard to ensure Embedded Versioning data will not be available in content staging builds, and (2) the fetched data should always have the same content as the latest metadata/changes of the build anyways.

This PR also cleans up and consolidates logic for `process.env.GATSBY_IS_PREVIEW === 'true'`. Some instances where this logic occurs are either not relevant given the current state of the Gatsby source plugin, or are redundant (i.e. no point in checking the env in the preview source plugin since the preview source plugin would only be used if the env is already true).

### DIFF
--- a/src/context/toc-context.js
+++ b/src/context/toc-context.js
@@ -15,22 +15,22 @@ const TocContext = createContext({
 // filters all available ToC by currently selected version via VersionContext
 const TocContextProvider = ({ children, remoteMetadata }) => {
   const { activeVersions, setActiveVersions, availableVersions, showVersionDropdown } = useContext(VersionContext);
-  const { project, toctree, associated_products: associatedProducts } = useSnootyMetadata();
-  const { database, parserBranch } = useSiteMetadata();
+  const { project, branch, toctree, associated_products: associatedProducts } = useSnootyMetadata();
+  const { database } = useSiteMetadata();
   const [activeToc, setActiveToc] = useState(remoteMetadata?.toctree || toctree);
   const [isLoaded, setIsLoaded] = useState(false);
 
   const getTocMetadata = useCallback(async () => {
     // Embedded versioning is not expected to work in staging builds, so we should
     // be able to safely return the default toctree.
-    if (isGatsbyPreview) {
+    if (isGatsbyPreview()) {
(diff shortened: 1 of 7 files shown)


src/context/toc-context.js
```diff
@@ -14,16 +15,22 @@ const TocContext = createContext({
...
-  const { database, parserBranch } = useSiteMetadata();
+  const { project, branch, toctree, associated_products: associatedProducts } = useSnootyMetadata();
+  const { database } = useSiteMetadata();
   const [activeToc, setActiveToc] = useState(remoteMetadata?.toctree || toctree);
   const [isLoaded, setIsLoaded] = useState(false);
 
   const getTocMetadata = useCallback(async () => {
+    // Embedded versioning is not expected to work in staging builds, so we should
+    // be able to safely return the default toctree.
+    if (isGatsbyPreview()) {
```
schmalliso, line 26: NICE

schmalliso (approved):
So clean! So logical!

PR #888, opened 2023-09-11
DOP-3988, DOP-39899: client libraries and product intros

caesarbell:
### Stories/Links:

DOP-3989

### Current Behavior:

[Prod](https://www.mongodb.com/docs/)

### Staging Links:

[Staged Commit](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/50f0f7a9bf593856c8fb68c00dfbb879902af007/DOP-3988/landing/caesarbell/DOP-3989/index.html)

### Notes:

This PR adds the new standalone component that is used with the Card Group Component. The designs are found [here](https://www.figma.com/file/SxKPGkhEJ8ixM9giIPEuJC/Docs-Homepage-Redesign?node-id=1351%3A59676&mode=dev). This new standalone component will alter the card group component's style only if it is its sibling.


src/components/Products/ProductItem.js
```diff
@@ -0,0 +1,80 @@
...
+
+  @media ${theme.screenSize.largeAndUp} {
+    width: 58vw;
+  }
+`;
+
+const StyledImage = styled.img`
+  height: 61px;
+  width: 61px;
+`;
```
rayangler, line 30: Is it possible to just assign `height` and `width` by property instead of css? I noticed the height/width readjusting after initial page load, but that shouldn't happen when using the native element properties since the browser does all of the size adjustments prior to css load.

```js
<img height={IMG_SIZE} width={IMG_SIZE} />
```

src/components/Products/ProductItem.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from '@emotion/styled';
+import { css } from '@leafygreen-ui/emotion';
+import { withPrefix } from 'gatsby';
+import { theme } from '../../theme/docsTheme';
+import ComponentFactory from '../ComponentFactory';
+import SectionHeader from '../SectionHeader';
+
+const FONT_SIZE = '16px';
```
rayangler, line 10: (Nit) We have a constant for 16px font size located in `docsTheme`: `theme.fontSize.default`, in case you want to use that

src/components/SectionHeader.js
```diff
@@ -0,0 +1,30 @@
...
+const customStyleHeader = css`
+  color: #061621;
+  font-size: 24px;
+  font-weight: 500;
+  margin: 0;
+`;
+
+const SectionHeader = ({ children, customStyles, as }) => {
+  return (
+    <Body className={cx([customStyleHeader, customStyles])} as={as}>
```
rayangler, line 15: It might be better to use LG's `H3` component instead of `Body`. `Body` is mostly representative of paragraphs rather than headings. I think it would also help keep most of the CSS we expect from Figma, while also ensuring the element is already an `h3`.

src/components/SectionHeader.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css, cx } from '@leafygreen-ui/emotion';
+import { Body } from '@leafygreen-ui/typography';
+
+const customStyleHeader = css`
+  color: #061621;
```
rayangler, line 7: Might be a nit, but this color is only expected for the standalone heading and not the product item headings

src/components/Products/ProductItem.js
caesarbell: @rayangler will make this change. Thanks for catching this.

src/components/SectionHeader.js
caesarbell: This was a good nit catch. Thanks.

src/components/SectionHeader.js
caesarbell: That is fair. My thought was to keep it flexible in case we wanted to use it in other place. But it makes sense to lock it down to just an H3 and inherit most of the styles from LG.

src/components/Products/ProductItem.js
caesarbell: @rayangler made this change as well.

src/components/StandaloneHeader.js
```diff
@@ -0,0 +1,53 @@
+import React, { useEffect, useRef } from 'react';
+import PropTypes from 'prop-types';
+import styled from '@emotion/styled';
+import { theme } from '../theme/docsTheme';
+import ComponentFactory from './ComponentFactory';
+import SectionHeader from './SectionHeader';
+import Link from './Link';
+
+const StyledStandaloneContainer = styled.div`
```
seungpark, line 9: similar here (trying to stray away from `styled`) , we can use cx(css)

src/components/StandaloneHeader.js
```diff
@@ -0,0 +1,53 @@
...
+`;
+
+const StandaloneHeader = ({ nodeData: { argument, options } }) => {
+  const ele = useRef(null);
+  useEffect(() => {
+    // When using the standalone header next to the card-group element, remove card group
+    // margin top styles.
+    const nextSib = ele.current.nextSibling;
+    if (nextSib.classList.contains('card-group')) {
+      nextSib.style.marginTop = '0px';
```
seungpark, line 27: would change this to styling within card group by passing an option. this will be hard to debug and overwrite

src/components/Card/CardGroup.js
```diff
@@ -41,8 +48,15 @@ const StyledGrid = styled('div')`
   grid-column-gap: ${theme.size.medium};
   grid-row-gap: ${theme.size.medium};
   grid-template-columns: ${(props) => `repeat(${getColumnValue(props)}, 1fr)`};
-  margin: ${({ isForDrivers }) =>
-    isForDrivers ? `${theme.size.large} 0 ${theme.size.xlarge}` : `${theme.size.large} 0`};
+  margin: ${({ isForDrivers }) => getMarginStyles(isForDrivers, 0)};
```
seungpark, line 51: might be a good place to replace styled for cx(css). seeing a hydration here (post SSR)

seungpark (commented):
is this for both 3988 and 3989? seeing changes from product list (3988)

also `client libraries` section is missing kotlin entry, and the svg for c++

caesarbell:
> 

This is for both `DOP-3988` and `DOP-3989` I closed the PR for `DOP-3988` since the work for both was also in this PR.  I must have missed Kotlin due to it not being in the mobile [designs](https://www.figma.com/file/SxKPGkhEJ8ixM9giIPEuJC/Docs-Homepage-Redesign?node-id=1351%3A59676&mode=dev).

src/components/ComponentFactory.js
```diff
@@ -213,7 +219,7 @@ const ComponentFactory = (props) => {
     }
 
     // Warn on unexpected usage of domains, but don't break
-    const validDomains = ['mongodb', 'std'];
+    const validDomains = ['mongodb', 'std', 'landing'];
```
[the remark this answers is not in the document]
caesarbell: @seungpark Yeah, we were getting warnings from this [logic](https://github.com/mongodb/snooty/blob/50f0f7a9bf593856c8fb68c00dfbb879902af007/src/components/ComponentFactory.js#L223) in the console for it not being included as a valid domain.

src/components/Products/ProductItem.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from '@emotion/styled';
```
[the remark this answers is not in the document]
caesarbell: The cards for the client-library? We would need to update the styles in the CardGroup, not the ProductItem component if I understood you correctly.

src/components/Products/ProductItem.js
```diff
@@ -0,0 +1,80 @@
...
+`;
+
+const ProductItem = ({ nodeData: { children, options, argument } }) => {
+  const icon = options.icon;
+  const iconAlt = options['icon-alt'];
+  return (
+    <StyledSection>
+      <StyledImage src={withPrefix(icon)} alt={iconAlt} />
+      <StyledArticle>
+        <SectionHeader as="h3" customStyles={customStyleHeader}>
```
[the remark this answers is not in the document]
caesarbell: @seungpark that has already been changed, and it is using SectionHeader which uses H3 from LG.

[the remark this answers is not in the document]
caesarbell:
```
<StyledSection>
      <img src={withPrefix(icon)} alt={iconAlt} height={HEIGHT} width={WIDTH} />
      <StyledArticle>
        <SectionHeader customStyles={customStyleHeader}>
          {argument.map((child, i) => (
            <ComponentFactory nodeData={child} key={i} />
          ))}
        </SectionHeader>
        {children.map((child, i) => (
          <ComponentFactory nodeData={child} key={i} showLinkArrow={true} />
        ))}
      </StyledArticle>
    </StyledSection><StyledSection>
      <img src={withPrefix(icon)} alt={iconAlt} height={HEIGHT} width={WIDTH} />
      <StyledArticle>
        <SectionHeader customStyles={customStyleHeader}>
          {argument.map((child, i) => (
            <ComponentFactory nodeData={child} key={i} />
          ))}
        </SectionHeader>
        {children.map((child, i) => (
          <ComponentFactory nodeData={child} key={i} showLinkArrow={true} />
        ))}
      </StyledArticle>
    </StyledSection>
```

src/components/StandaloneHeader.js
caesarbell: @seungpark the hydration issue happens when using styled components on LG, but if we use styled on HTML elements we don't get the hydration issue. Unless the ask is to not use HTML tags like `<article>` and find a LG version of it?

src/components/Card/CardGroup.js
caesarbell: @seungpark I could refactor this, but that logic was pre-existing. I can work on this though.

src/components/StandaloneHeader.js
caesarbell: What happens if this needs to be removed in the future, then we would need to go back to the card group and remove this prop. My thought was since it was directly associated with the card we could have the logic in one place. if we remove this standalone the card group works as normal and no additional changes are needed.

caesarbell: That's fair, and since this is the new layout we didn't need to keep the margin-top styles from the old layout. No prop and no logic are needed.

src/components/Card/CardGroup.js
caesarbell: I believe the hydration issue comes from using style components with LG, but if we use styled-components on an HTML element it works fine. No styles changed post SSR

caesarbell: Let me know if you are still seeing any hydration issues.

seungpark: good catch! coming in right away from SSR 👍

src/components/ComponentFactory.js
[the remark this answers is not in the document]
seungpark: verified the existence... question for @i80and here. what exactly is node domain and how does this get populated? (is this related to the double "landing" [domain key](https://github.com/mongodb/snooty-parser/blob/master/snooty/parser.py#L1005)?)

src/components/Products/ProductItem.js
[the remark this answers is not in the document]
seungpark: works for me! as you mentioned, non LG components here 👍

src/components/StandaloneHeader.js
seungpark: good catch! generally, i'd like to stick to one styling format but this suffices

seungpark (approved):
LGTM ! its coming alive 🧟 !

PR #891, opened 2023-09-12
DOP-3996: Allow writer's staging to use staging Chatbot server url

mmeigs:
### Stories/Links:

Clean up of 
DOP-3996

### Notes:

As Raymund pointed out [here](https://github.com/mongodb/snooty/pull/880#discussion_r1319143056), the SNOOTY_ENV value of "production" is used for writer's staging. Docs writer's staging should use the staging Chatbot server url for visibility in testing.


rayangler (approved):
> Docs writer's staging should use the staging Chatbot server url for visibility in testing.

Thanks for the confirmation here! I wasn't actually sure and it was a bit ambiguous from the ticket

PR #905, opened 2023-09-18
Redesign bugs: target only second section of card group in landing

seungpark:
Targeting only the second card group in landing page by.
1) adding new className for card-group (one has [type=drivers](https://github.com/mongodb/docs-landing/pull/132/files#diff-f9560dd277cc6f848625b9db1465ddf3779112495f2d1cc559ff3f91ea6d186cR86))
2) targeting `drivers` classname from landing page stylesheet.

[staged changes](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/newhomepagewhodis/landing/seung.park/landing-cards-margins/index.html)

[vs master branch](https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/)

### DIFF
--- a/src/components/Card/CardGroup.js
+++ b/src/components/Card/CardGroup.js
@@ -90,7 +90,14 @@ const CardGroup = ({
 
   return (
     <StyledGrid
-      className={['card-group', isCompact ? 'compact' : '', isExtraCompact ? 'extra-compact' : '', className].join(' ')}
+      className={[
+        'card-group',
+        isCompact ? 'compact' : '',
+        isExtraCompact ? 'extra-compact' : '',
+        isForDrivers ? 'drivers' : '',
+        'card-group',
+        className,
+      ].join(' ')}
       columns={columns}
       noMargin={true}
       isCarousel={isCarousel}
--- a/src/templates/landing.js
+++ b/src/templates/landing.js
@@ -44,12 +44,13 @@ const Wrapper = styled('main')`
         grid-column: 2 / -2 !important;
       }
 
-      &:not(.compact, .extra-compact) {
+      &:not(.compact, .extra-compact, .drivers) {
         p {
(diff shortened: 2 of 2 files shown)


src/components/Card/CardGroup.js
```diff
@@ -90,7 +90,14 @@ const CardGroup = ({
 
   return (
     <StyledGrid
-      className={['card-group', isCompact ? 'compact' : '', isExtraCompact ? 'extra-compact' : '', className].join(' ')}
+      className={[
+        'card-group',
+        isCompact ? 'compact' : '',
+        isExtraCompact ? 'extra-compact' : '',
+        isForDrivers ? 'drivers' : '',
+        'card-group',
```
rayangler, line 98: This `card-group` is a dupe

rayangler (commented):
Correct me if I'm wrong, but the text in the card still looks a little bit off center. I'm wondering if we need `align-items: center;` applied to the drivers cards styling

src/components/Card/CardGroup.js
seungpark: ha! caught it before my push. removed

seungpark:
> Correct me if I'm wrong, but the text in the card still looks a little bit off center. I'm wondering if we need `align-items: center;` applied to the drivers cards styling

is this referring to the driver cards? good catch. done!

src/templates/landing.js
```diff
@@ -44,12 +44,19 @@ const Wrapper = styled('main')`
         grid-column: 2 / -2 !important;
       }
 
-      &:not(.compact, .extra-compact) {
+      &.drivers {
+        > div {
+          align-items: center;
+        }
+      }
```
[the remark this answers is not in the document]
seungpark: good catch @rayangler ! updated to use this instead

PR #911, opened 2023-09-18
DOP-4028: Fix margin for links with svgs

anabellabuckvar:
### Stories/Links:

[DOP-4028](https://jira.mongodb.org/browse/DOP-4028)

### Current Behavior:
The spacing for the caret dropdown icons on the side nav is slightly out of line
[staging] (https://preview-mongodbrayanglerstg.gatsbyjs.io/cloud-docs/gc-test-img/)
image

### Staging Links:

[Docs-landing](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/newhomepagewhodis/landing/anabella.buckvar/DOP-4028/)

### Notes:

As part of [DOP-3989](https://jira.mongodb.org/browse/DOP-3989), some [css](https://github.com/mongodb/snooty/pull/888/files#diff-927f813ea3772395ca9598b2d8d6dda38e38460089aba25bffc443d4859b84cbR37-R39) was added to add a margin for svgs nested in links. It seems that it was necessary for correct spacing because the last link in the [products list](https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/) didn't space the arrow correctly due to how the link was being generated. However, the margin fix ended up caused the spacing to change slightly for the caret dropdown icons on the side nav. 

In fixing this, I noticed that the method of displaying arrows in `Links.js` was different for LGLinks, which used a LG prop, than for GatsbyLinks, which used `ArrowRightIcon`. This was rendering them with slightly different types of arrows as well as margins between link text and the arrows. For consistency, I changed both to use `ArrowRightIcon` instead


src/components/RefRole.js
```diff
@@ -42,30 +47,8 @@ const RefRole = ({ nodeData: { children, domain, fileid, name, url }, slug, card
...
-    if (filename === slug) {
-      // Internal page link
-      link = `#${html_id}`;
-    } else if (html_id === '') {
-      // :doc: link
-      link = filename;
-    } else {
-      link = `${filename}/#${html_id}`;
-    }
-  }
```
rayangler, line 60: I'm not sure if removing this logic altogether is desired. This logic is used to help determine how internal references in the docs can be translated to an actual link. Without this, the rendered link doesn't resolve anywhere (the link in the staging link renders okay with the arrow icon, but clicking the link just brings you back to the current page)

Is there another solution we can take that won't affect the way these links are resolved?

rayangler (changes requested):

src/components/RefRole.js
anabellabuckvar: I undid the changes to this logic and found the real reason that the last `ProductItem` was rendered with a different svg margin(and, a very slightly different arrow icon). Arrows should be standardized across all links now

src/components/Link.js
```diff
@@ -81,7 +80,14 @@ const Link = ({
     // Ensure trailing slash
     to = to.replace(/\/?(\?|#|$)/, '/$1');
 
-    const decoration = showLinkArrow ? <ArrowRightIcon role="presentation" size={12} /> : '';
+    const decoration = showLinkArrow ? (
+      <ArrowRightStyling>
+        {' '}
+        <ArrowRightIcon role="presentation" size={12} />{' '}
+      </ArrowRightStyling>
```
rayangler, line 87: It's not typical to use a `css` variable as an element. Did you mean to use a `styled` component?

As-is, I also don't see the left margin being applied (due to the use of the `css` variable). I'm curious if this means that we don't even need the css and some whitespace is all we need?

src/components/Link.js
```diff
@@ -101,17 +107,25 @@ const Link = ({
...
     <LGLink
       className={joinClassNames(LGlinkStyling, className)}
       href={to}
       hideExternalIcon={!showExtIcon}
-      arrowAppearance={showLinkArrow ? 'persist' : 'none'}
       target={target}
       {...other}
     >
       {children}
+      {decoration}
```
rayangler, line 128: Is it possible to leave the `arrowAppearance` prop as-is? I think it might be preferable to use LG's built-in prop for this instance

rayangler:
Can you also please update the staging link @anabellabuckvar ? I think the previous changes (incorrect link) are still present.

src/components/Link.js
caesarbell: That might be the trade-off or I wonder if we could leverage the custom `className` and apply the `className` value in the ProductItem.js file 

```
<StyledArticle>
        <SectionHeader customStyles={customStyleHeader}>
          {argument.map((child, i) => (
            <ComponentFactory nodeData={child} key={i} />
          ))}
        </SectionHeader>
        {children.map((child, i) => (
          <ComponentFactory nodeData={child} key={i} showLinkArrow={true} className='some-custom-className' />
        ))}
      </StyledArticle>
```

Then we apply the style using that custom class name. This way it doesn't have an effect on the other links.

caesarbell (commented):
I left some suggestions.

src/components/Link.js
caesarbell: For more clarity. 

**ProductItem**
```
const customStylesForArrows = css` // choose a better name
  > svg {
    margin-left: 3px;
  }
`;

// then pass it down as a prop
{children.map((child, i) => (
  <ComponentFactory nodeData={child} key={i} showLinkArrow={true} className={cx(customStylesForArrows)} />
))}
```

Then in `RefRole` we can pass that down to the Link in question

```
 const stylingClass = cardRef ? cardRefStyling : '';
  if (url) {
    return (
      <Link className={cx(stylingClass, className)} to={url} showLinkArrow={showLinkArrow}>
        {children.map((node, i) => (
          <ComponentFactory key={i} nodeData={node} />
        ))}
      </Link>
    );
  }
```
This should only allow the style to be applied from the ProductItem component as the style class is defined there.

caesarbell: @rayangler thoughts? ☝️

rayangler: I think both this and Anabella's current solution are reasonable. 

Ideally, we would just use the prop from LG's Link component, but unfortunately, since we use both LG Link and Gatsby Link, it makes things a little annoying and/or inconsistent. The benefit of the current solution is that it enforces consistency in both styling (even if the difference is subtle) and implementation while keeping the styling all within the component (big win in terms of readability and traceability).

Since the intention of this change is to prevent a cascading shift in styling for unintended elements, I think the current solution is fine for now. In the future, we should revisit if it's possible and/or desirable to render a Gatsby Link as a LG Link so that we can much more cleanly use the same shared prop.

src/components/RefRole.js
```diff
@@ -32,6 +32,7 @@ const stopPropagation = function (e) {
 const RefRole = ({ nodeData: { children, domain, fileid, name, url }, slug, cardRef, showLinkArrow }) => {
   // Render intersphinx target links
   const stylingClass = cardRef ? cardRefStyling : '';
+  console.log(url);
```
rayangler, line 35: Can we remove this?

rayangler (approved):
LGTM, as long as the `console.log` is removed

caesarbell:
@anabellabuckvar outside of our convo, if it doesn't work this solution is okay as well. I will mark approved to not block or hold you up.

caesarbell (approved):

src/components/Link.js
anabellabuckvar: implemented these changes

caesarbell:
Another option could be adding the style changes in the `RefRole.js` if this spacing needs to be more globally acceptable.

PR #912, opened 2023-09-19
DOP-4029: Remove old feedback widget from headings on mobile

anabellabuckvar:
### Stories/Links:

[DOP-4029](https://jira.mongodb.org/browse/DOP-4029)

### Current Behavior:
Currently, a non-functional version of the old feedback widget appears next to headings on mobile in addition to the new feedback widget.
[atlas getting started](https://www.mongodb.com/docs/atlas/getting-started/)

### Staging Links:

[atlas getting started
](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/anabella.buckvar/DOP-4029/getting-started/index.html)

### Notes:
There is other unused code related to the old feedback widget but in files that new feedback widget uses that can likely be removed as well.


src/components/Link.js
```diff
@@ -13,6 +13,9 @@ import { joinClassNames } from '../utils/join-class-names';
  * https://www.gatsbyjs.org/docs/gatsby-link/#recommendations-for-programmatic-in-app-navigation
  */
 
+const ArrowRightStyling = css`
```
schmalliso, line 16: These changes shouldn't be included in this PR. Can you double check that you've not got commits from your other branch in this patchset?

anabellabuckvar: Oops, sorry about that! they should be gone now

src/components/Heading.js
```diff
@@ -53,9 +50,7 @@ const Heading = ({ sectionDepth, nodeData, page, ...rest }) => {
         wrapper={(children) => (
           <HeadingContainer stackVertically={isMobile}>
             {children}
-            <ChildContainer isStacked={isMobile}>
-              {hasSelectors ? <TabSelectors /> : <FeedbackHeading isStacked={isMobile} />}
-            </ChildContainer>
+            <ChildContainer isStacked={isMobile}>{hasSelectors ? <TabSelectors /> : ''}</ChildContainer>
```
rayangler, line 53: (Nit) Since we don't need to render anything when the condition is false, we can probably simplify this to be like the following:

```js
{hasSelectors && <TabSelectors />}
```

rayangler (commented):
Looks pretty good. I'm curious if we still need [this line](https://github.com/mongodb/snooty/pull/912/files#diff-fa28c3c4f26fa56c54a4a3a38a1b775147471e1673bed13ffaeef10bd3b516f4L44) in the `Heading` component, since this seems related to the conditional rendering of the feedback header:

```js
const hidefeedbackheader = page?.options?.hidefeedback === 'header' || page?.options?.hidefeedback === 'page';
```

rayangler (approved):

PR #915, opened 2023-09-20
Implement Gatsby Cloud source plugin

rayangler:
### Stories/Links:

DOP-3444 (project epic)

### Current Behavior:

[Server docs staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/raymundrodriguez/master/index.html) - master branch
[Cloud docs staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/raymundrodriguez/master/index.html) - master branch

### Staging Links:

[Server docs staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/raymundrodriguez/gatsby-cloud-rc/index.html) - using new prod plugin
[Cloud docs staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/raymundrodriguez/gatsby-cloud-rc/index.html) - using new prod plugin
[Gatsby Cloud preview](https://preview-mongodbrayangler.gatsbyjs.io/docs/test-page-queries/) - random cloud-docs branch. This Gatsby Cloud site uses the current `gatsby-cloud-rc` branch. Uses new preview plugin

### Notes

Summary of the changes made:

* The `gatsby-node.js` file that existed in the root of the repo has been split up into two `gatsby-node` files: one for prod builds (default), and one for Gatsby Cloud preview builds (requires an environment variable to use). This is representative of the 2 plugins found in the new `plugins/` folder. Some of the shared logic has been exported into util functions, but we should be mindful of any changes needing to be shared between the two plugins to minimize duplicate logic.
* The new `gatsby-source-snooty-preview` plugin is to be used for Gatsby Cloud:
  * The name of the plugin itself is used for Autobuilder webhook calls to Gatsby Cloud to allow any data passed through `webhookBody` to be consumed. This will mostly just be the Autobuilder `jobId` to help track jobs.
  * The plugin fetches all build data for a single user from the Snooty Data API one document (metadata, page AST, etc.) at a time and creates nodes for them. Nodes made for page paths are queried against to create pages using the `DocumentBodyPreview` component. Certain layout components (like the side nav, context, etc.) needed to be rearranged to accommodate passing page data through graphql instead of through page context. (This remains unchanged for prod builds though)
  * Gatsby Cloud sites expect data from multiple projects and branches to live together. To accommodate this, pages for the preview plugin expect to have page paths that fall under the format `/<PROJECT>/<BRANCH>/<PAGE_PATH>` instead of just `<PAGE_PATH>` with a static path prefix. Components such as `Link` have been modified to be compatible with these paths, only for preview builds.
  * The `useSnootyMetadata` hook has been adjusted to become a context rather than a static query to accommodate graphql data changes. Leaving it as a static query would cause unsuccessful incremental builds on Gatsby Cloud due to the static query's data being updated whenever new metadata was being added. Functionally, this should still work the same as before, regardless of which plugin is being used.


plugins/gatsby-source-snooty-preview/gatsby-node.js
```diff
@@ -0,0 +1,319 @@
...
+        }
+      } else if (entry.type === `page`) {
+        if (entry.data?.ast?.options?.template === 'changelog') hasOpenAPIChangelog = true;
+        pageCount += 1;
+        const { source, ...page } = entry.data;
+
+        const filename = getNestedValue(['filename'], page) || '';
+        // The old gatsby sourceNodes has this code — I'm not sure it actually
+        // is a concern (I couldn't find any page documents that didn't end in .txt)
+        // but Chesterton's Fence and all.
```
mmeigs, line 170: <3

plugins/gatsby-source-snooty-preview/gatsby-node.js
```diff
@@ -0,0 +1,319 @@
...
+        if (filename.endsWith('.txt')) {
+          const branch = page.page_id.split(`/`)[2];
+          const raw_page_id = page.page_id.split(`/`).slice(3).join(`/`);
+          const page_id = raw_page_id === `index` ? `/` : `/${raw_page_id}`;
+          const project = page.page_id.split(`/`)[0];
+
+          const pageNodeId = createNodeId(page_id + project + branch);
+          const pagePathNodeId = pageNodeId + `/path`;
```
mmeigs, line 178: qq: what's the reason we're not using `createNodeId` here too?

plugins/gatsby-source-snooty-preview/gatsby-node.js
```diff
@@ -0,0 +1,319 @@
...
+            page.metadata = createSnootyMetadataId({ createNodeId, branch, project });
+            page.id = pageNodeId;
+            page.internal = {
+              type: `Page`,
+              contentDigest: createContentDigest(page),
+            };
+
+            const pagePathNode = {
+              id: pagePathNodeId,
+              page_id: page_id,
```
mmeigs, line 194: nittiest nit: Just for consistency and readability, would prefer the `page_id,` rather than `page_id: page_id,`

src/components/DocumentBodyPreview.js
```diff
@@ -62,8 +63,13 @@ const getAnonymousFootnoteReferences = (index, numAnonRefs) => {
...
   const {
     location,
-    pageContext: { page, slug, template },
+    pageContext: { slug },
+    data,
   } = props;
+  const page = data.page.ast;
+  const metadata = data.page.metadata.metadata;
+  const template = page?.options?.template;
+  props.pageContext.page = page;
```
mmeigs, line 72: Seems like we're doubling up on passing `page` through to the pageContext here and lower when we spread out props.pageContext.

mmeigs (commented):
Okay, I'm almost halfway. Taking a break to refresh 😅

plugins/gatsby-source-snooty-preview/gatsby-node.js
```diff
@@ -0,0 +1,319 @@
...
+  });
+  await fs.writeFile(path.join('public', file), data, 'binary');
+};
+
+const APIBase = process.env.API_BASE || `https://snooty-data-api.mongodb.com`;
+const GATSBY_CLOUD_SITE_USER = process.env.GATSBY_CLOUD_SITE_USER || `mmeigs`;
+
+function createSnootyMetadataId({ branch, project, createNodeId }) {
+  return createNodeId(`metadata-${branch}-${project}`);
+}
```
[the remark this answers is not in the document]
rayangler: Yup, exactly!

plugins/gatsby-source-snooty-preview/gatsby-node.js
rayangler: I think Kyle left this comment, but I'm also down to keep the check 🤠 Ideally, the Snooty Data API should continue to return only txt page data

plugins/gatsby-source-snooty-preview/gatsby-node.js
rayangler: Seems like we're already using `createNodeId` for `pageNodeId`, so uniqueness should already be guaranteed

plugins/gatsby-source-snooty-preview/gatsby-node.js
```diff
@@ -0,0 +1,319 @@
...
+
+const saveFile = async (file, data) => {
+  await fs.mkdir(path.join('public', path.dirname(file)), {
+    recursive: true,
+  });
+  await fs.writeFile(path.join('public', file), data, 'binary');
+};
+
+const APIBase = process.env.API_BASE || `https://snooty-data-api.mongodb.com`;
+const GATSBY_CLOUD_SITE_USER = process.env.GATSBY_CLOUD_SITE_USER || `mmeigs`;
```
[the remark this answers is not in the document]
rayangler: Haha, whoops! I ended up introducing a new line to check if `GATSBY_CLOUD_SITE_USER` is truthy, and throw an error otherwise. This might make it easier to spot if no site user is declared. Let me know if you disagree!

plugins/gatsby-source-snooty-preview/gatsby-node.js
rayangler: Done!

src/components/DocumentBodyPreview.js
rayangler: I spent some time digging into this and it seems like this is being done to keep `pageContext` consistent between the two high-level components that use `page` across both plugins: the `Layout` component, and the `Document` template. By omitting this line, we fail to pass this down to the `Document` template for the preview plugin, but this is automatically done by the prod plugin (since it just shoves everything down the `pageContext`).

To prevent doubling up, I've removed the line where we do:

```js
pageContext={{
  page, // this is gone now
  ...props.pageContext,
}}
```

We should probably reevaluate how `page` is being passed down the `pageContext` in the prod plugin in the future and see if it's even necessary to pass down the whole object. This should help to reduce the size of `pageContext` and to keep the structure of `pageContext` more consistent between both plugins.

mmeigs: Cool, makes sense! Removing the other line works for me!

mmeigs (approved):
Woohoo!! LGTM! Thanks for doing the most massive work ever in putting this together.

plugins/gatsby-source-snooty-preview/gatsby-node.js
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    const headers = {};
+    if (clientAccessToken) {
+      headers['Authorization'] = `Bearer ${clientAccessToken}`;
+    }
+    const httpStream = got.stream(url, { headers });
+
+    const decode = parser();
+    decode.on(`data`, async (_entry) => {
```
casthewiz, line 124: Larger than a nit but less than a blocker:
Updating this block of if statements would be tricky in the future should the interface coming from the data api change.

We might need to think a bit about how to make this less nested and more contextual (function per `entry.type`?) in the future. It's hard to tell what some of the cases handle at first glance, and adding comments to nested ternaries like this might make it worse 🤔

plugins/gatsby-source-snooty-preview/other-things-to-source.js
```diff
@@ -0,0 +1,30 @@
...
+  let hasCloudDocsProject = false;
+  for (const { metadata } of nodes) {
+    if (metadata.project === `cloud-docs`) {
+      hasCloudDocsProject = true;
+      break;
+    }
+  }
+
+  if (hasCloudDocsProject && hasOpenAPIChangelog)
+    await createOpenAPIChangelogNode({ createNode, createNodeId, createContentDigest, siteMetadata, db });
```
casthewiz, line 29: nonblocking, for my curiosity:
is there a strong reason for adding cloud-docs detection here? This feels like a landmine if we need to add a changelog to another project.

casthewiz (approved):
Just a few questions/open threads, nothing actually blocking though imo. 

This looks awesome!

plugins/gatsby-source-snooty-preview/other-things-to-source.js
mmeigs: This is from me back in the spring! Before, accessing the project was a simple value check, and so I used it as another safeguard in creation of the changelog until another team wanted to use it. But I'm fine with taking it out! The check for hasOpenAPIChangelog is plenty!

rayangler: No strong reason for the detection other than to keep it in sync with current prod implementation! I can go ahead and remove it as part of this PR and rely on `hasOpenAPIChangelog`. 

Although thinking about it now, we'll probably still need to refactor this logic to be able to support different changelogs, and differentiate them by project, and maybe branch (similar to how we had to do it for metadata for different projects + branches)

plugins/gatsby-source-snooty-preview/gatsby-node.js
rayangler: Yup, I agree with the sentiment here. I think it might make sense to make this more readable as a separate fast-follow task. This PR has already gotten pretty chunky and I'd like to make this more of an isolated change to make it easier to revert in case the light refactoring goes awry

plugins/gatsby-source-snooty-preview/other-things-to-source.js
rayangler: I went ahead and removed the check for `cloud-docs` 👍

plugins/gatsby-source-snooty-preview/gatsby-node.js
rayangler: Created DOP-4063 to handle this change

PR #919, opened 2023-09-26
DOP-3903: API Changelog Support to omit entries

anabellabuckvar:
### Stories/Links:

[DOP-3903](https://jira.mongodb.org/browse/DOP-3903)

### Current Behavior:

[Production API Changelog
](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/changelog/)

### Staging Links:

[Staging API changelog](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/anabella.buckvar/DOP-3903-b/reference/api-resources-spec/changelog/index.html)

### Notes:
Atlas App Services Function [fetchOADiff](https://realm.mongodb.com/groups/5bad1d3d96e82129f16c5df3/apps/5d41ef5e2de84772fc7c2de2/functions/646cc47cf571983e9ce0671e) was updated in the process of completing this ticket, please review it as well before giving a LGTM!

### DIFF
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -10,3 +10,10 @@ export const wrapRootElement = ({ element }) => (
     <LeafyGreenProvider baseFontSize={16}>{element}</LeafyGreenProvider>
   </ThemeProvider>
 );
+
+export const shouldUpdateScroll = ({ routerProps: { location } }) => {
+  if (location?.state?.preserveScroll) {
+    return false;
+  }
+  return true;
+};
--- a/src/components/Figure/Lightbox.js
+++ b/src/components/Figure/Lightbox.js
@@ -10,21 +10,24 @@ const CAPTION_TEXT = 'click to enlarge';
 const StyledModal = styled(Modal)`
   // Set z-index to appear above side nav and top navbar
   z-index: 10;
+  margin-top: 64px;
 
   @media ${theme.screenSize.largeAndUp} {
     div[role='dialog'] {
-      width: 80%;
+      max-width: 80%;
+    }
+    img {
+      max-height: 620px;
     }
   }
 
   @media ${theme.screenSize.upToLarge} {
     div[role='dialog'] {
-      width: 100%;
+      max-width: 100%;
     }
-  }
-
-  img {
-    width: 100%;
+    img {
+      max-height: 300px;
+    }
   }
 `;
 
@@ -43,6 +46,7 @@ const LightboxWrapper = styled('div')`
   margin-top: ${theme.size.medium};
   margin-bottom: ${theme.size.medium};
   display: block;
+  max-height: 620px;
 `;
 
 const Lightbox = ({ nodeData, ...rest }) => {
@@ -57,7 +61,7 @@ const Lightbox = ({ nodeData, ...rest }) => {
         </div>
         <CaptionLegend {...rest} nodeData={nodeData} />
       </LightboxWrapper>
-      <StyledModal size="large" open={open} setOpen={setOpen}>
+      <StyledModal size="medium" open={open} setOpen={setOpen}>
         <Image nodeData={nodeData} />
       </StyledModal>
     </React.Fragment>
--- a/src/components/SearchResults/Facets/FacetTags.js
+++ b/src/components/SearchResults/Facets/FacetTags.js
@@ -1,0 +1,170 @@
+import React, { useCallback, useContext, useEffect, useMemo, useRef, useState } from 'react';
+import styled from '@emotion/styled';
+import Icon from '@leafygreen-ui/icon';
+import { css, cx } from '@leafygreen-ui/emotion';
+import { palette } from '@leafygreen-ui/palette';
+import { Overline } from '@leafygreen-ui/typography';
+import { theme } from '../../../theme/docsTheme';
+import Tag, { searchTagStyle } from '../../Tag';
+import SearchContext from '../SearchContext';
+import useFacets from './useFacets';
+import { initChecked } from './FacetValue';
+
+// util to get all current facets, derived from search params
+const getActiveFacets = (facetOptions, searchParams) => {
+  const res = [];
+
+  function checkFacetValue(facetValues) {
+    for (const facetValue of facetValues) {
+      // if it exists in search params, include
+      if (initChecked(searchParams, facetValue.key, facetValue.id)) {
+        res.push(facetValue);
+      }
+
+      // search its descendant options
+      if (facetValue.facets) {
+        checkFacetGroup(facetValue.facets);
+      }
+    }
+  }
+
+  function checkFacetGroup(facetOptions) {
+    for (const facetOption of facetOptions) {
+      checkFacetValue(facetOption.options);
+    }
+  }
+
+  checkFacetGroup(facetOptions);
+
+  return res;
+};
+
+const StyledTag = styled(Tag)`
+  column-gap: ${theme.size.small};
+  ${searchTagStyle}
+`;
+
+const clearButtonStyling = css`
+  text-transform: uppercase;
+  font-weight: 500;
+`;
+
+const MAX_HEIGHT = 30;
+
+const TagsFlexbox = styled('div')`
+  display: flex;
+  flex: 0 1 auto;
+  transition: max-height 0.15s ease-out;
+  overflow: hidden;
+`;
+
+const SelectionsFlexbox = styled('div')`
+  display: flex;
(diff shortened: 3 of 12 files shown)


caesarbell (commented):
Is there a way you can provide a staging link to verify the success of this PR?

caesarbell:
I am sorry, I just noticed this PR is a Draff.

gatsby-node.js
```diff
@@ -82,7 +82,42 @@ const fetchChangelogData = async (runId, versions, s3Prefix) => {
...
+      const versionUpdate = (version) => {
+        if (version !== null && version.changes !== null) {
+          version.changes = version.changes.filter((change) => !change.hideFromChangelog);
+        }
+        return version;
+      };
+
+      //pathUpdate takes the array of versions from the specific path passed in and takes each version and runs versionUpdate on it
+      const pathUpdate = (path) => {
+        path.versions = path.versions.map(versionUpdate);
```
branberry, line 97: Mutating the current object that is provided in the `map` could lead to unintended side effects. Just to err on the side of caution, you can make a copy of `path` and update that. You can make a copy like so: 
```js
const updatedPath = {...path}
```

gatsby-node.js
```diff
@@ -82,7 +82,42 @@ const fetchChangelogData = async (runId, versions, s3Prefix) => {
...
+          (version) => version && version.changes !== null && version.changes.length !== 0
+        );
+        return path;
+      };
+
+      //dateUpdate takes the array of paths from the specific date section passed in and takes each path and runs pathUpdate on it
+      const dateUpdate = (dateSection) => {
+        dateSection.paths = dateSection.paths.map(pathUpdate);
+        dateSection.paths = dateSection.paths.filter(
+          (path) => path !== null && path.versions !== null && path.versions.length !== 0
```
branberry, line 108: We can omit `path !== null`, we'll want to also check if `path.versions` is also `undefined`

gatsby-node.js
```diff
@@ -82,7 +82,42 @@ const fetchChangelogData = async (runId, versions, s3Prefix) => {
...
+        );
+        return dateSection;
+      };
+
+      //changelog is the json file with everything in it. Map at this level takes each date section and runs dateUpdate on it
+      changelog = changelog.map(dateUpdate);
+      return changelog.filter(
+        (dateSection) => dateSection !== null && dateSection.paths !== null && dateSection.paths.length !== 0
```
branberry, line 116: We can omit `dateSection !== null`, we'll want to also check if `dateSection.paths` is also `undefined`

src/components/OpenAPIChangelog/OpenAPIChangelog.js
```diff
@@ -72,6 +72,15 @@ const StyledLoadingSkeleton = styled.div`
 const OpenAPIChangelog = () => {
   const { snootyEnv } = useSiteMetadata();
   const { index = {}, changelog = [], changelogResourcesList = [] } = useChangelogData();
+  //updatedChangeLog = changelog.filter()
+  //console.log(changelog);
```
branberry, line 76: This can be removed

branberry (changes requested):
Overall, this is really good! I left some comments about some of the checks, and a small concern I had about the updating of values in the map calls.

Also, I would recommend adding unit tests to ensure that it's working as expected.

mmeigs (commented):
Mainly a comment below agreeing and expanding on what Brandon mentioned!
Also, just reminding here that this merge with the latest master will be an odd one. This will mostly all go in `plugins/utils/openapi` I would guess.

branberry (approved):
LGTM!

tests/unit/__snapshots__/OpenAPIChangelog.test.js.snap
```diff
@@ -1068,6 +1068,53 @@ exports[`OpenAPIChangelog tests OpenAPIChangelog renders correctly 1`] = `
...
+            class="emotion-60 emotion-61"
+          >
+            <li
+              class="emotion-62 emotion-63"
+            >
+              Resource version 2023-01-01 was removed, latest available resource version is 2023-02-01
+            </li>
+          </ul>
+        </div>
+      </section>
```
mmeigs, line 1117: Hmm, this makes me slightly worried? I believe this is the change you added that has the hideFromChangelog flag as true? So, it shouldn't be in the snapshot

tests/unit/__snapshots__/ChangeList.test.js.snap
```diff
@@ -254,6 +254,53 @@ exports[`OpenAPIChangelog ChangeList ChangeList renders all version changelog co
...
+          class="emotion-17 emotion-18"
+        >
+          <li
+            class="emotion-19 emotion-20"
+          >
+            Resource version 2023-01-01 was removed, latest available resource version is 2023-02-01
+          </li>
+        </ul>
+      </div>
+    </section>
```
mmeigs, line 303: Same here

plugins/utils/openapi.js
```diff
@@ -18,7 +48,23 @@ const fetchChangelogData = async (runId, versions, s3Prefix) => {
...
+        const updatedPath = { ...path };
+        if (path?.changes) {
+          updatedPath.changes = path.changes.filter((change) => !change.hideFromChangelog);
+        }
+        return updatedPath;
+      };
+
+      const updatedDiffData = diffData.map(pathUpdate);
+      return updatedDiffData.filter((path) => path.changes?.length);
+    };
```
mmeigs, line 65: Oh, and reminder to have this function somewhere where it can be exported so we can reuse it rather than declaring it twice!

anabellabuckvar: Refactored the functions, let me know what you think!

tests/unit/__snapshots__/ChangeList.test.js.snap
anabellabuckvar: I redid the tests, the snapshots should be clearer now.

tests/unit/ChangeList.test.js
```diff
@@ -54,37 +54,37 @@ describe('OpenAPIChangelog ChangeList', () => {
...
+  //   }, []);
+
+  //   const { queryByText } = render(<ChangeList changes={mockDiff} versionMode={COMPARE_VERSIONS} />);
+
+  //   expect(hiddenChanges).toHaveLength(2);
+
+  //   hiddenChanges.forEach((change) => {
+  //     expect(queryByText(change.change)).toBeNull();
+  //   });
+  // });
```
mmeigs, line 89: These can be removed, I believe

mmeigs (commented):
Two main things about placement of hiding the changes!
1. You can/should use the hideChanges util function in gatsby-node! That way it only occurs once, server-side. 
2. We still need to use hideDiffChanges in gatsby-node as well. Otherwise, the most recent diff will not have the changes hidden.

src/components/OpenAPIChangelog/utils/filterHiddenChanges.js
```diff
@@ -0,0 +1,41 @@
+export const hideChanges = (changelog) => {
```
mmeigs, line 1: Would also love a comment above this explaining what this first util function is doing (ie what field it's hiding and that it is meant to be used on the changelog rather than the diff

tests/unit/filterHiddenChanges.test.js
```diff
@@ -0,0 +1,79 @@
...
+      parserBranch: '',
+      patchId: '',
+      pathPrefix: '',
+      snootyBranch: '',
+      user: '',
+    },
+  },
+}));
+
+describe(' ChangeList data', () => {
```
mmeigs, line 29: I think all of this is fine! I also think if you wanted, this could just be a utility function test, so no need to render any components. Just pass a changelog and a diff through the functions and make sure the output is correct. But I'm totally fine with leaving all this. It accomplishes the same goal!

mmeigs (commented):
It looks good overall. But I am worried about using CommonJS exports in the frontend/react. Does this actually work? Especially if you comment out lines 17-20 in useFetchDiff?

anabellabuckvar:
> It looks good overall. But I am worried about using CommonJS exports in the frontend/react. Does this actually work? Especially if you comment out lines 17-20 in useFetchDiff?

Yes, it worked when I commented them out, and I used CommonJs exports only because I was following the format I saw in some of the other files. Would you like me to change it?

mmeigs (approved):
Ok, LGTM!

PR #927, opened 2023-10-02
DOP-3986

mayaraman19:
### Stories/Links:

[DOP-3986](https://jira.mongodb.org/browse/DOP-3986)

### Current Behavior:

Currently, if local storage is full, the browser will just display a white screen. This PR catches those errors so the browser will still render as well as adds unit tests. Error messages need to be refined.

### Staging Links:


seungpark (approved):
LGTM! nice test! mocked and restored.

minor note. we don't use set/getSessionValue but would wrap these around try/catch as well

PR #928, opened 2023-10-02
DOP-4060 adds language selector to unified footer on docs' pages.

caesarbell:
### Stories/Links:

DOP-4060

### Current Behavior:

[current behavior](https://www.mongodb.com/docs/)

### Staging Links:

[staged with hiding language selector](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/882f0e864792833928caca013c5b1f20bbad1662/master/atlas-cli/caesarbell/DOP-4060/index.html)

[staged with showing language selector](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/c58679996a06128466e9e7f89a1776f36d0cf21a/master/atlas-cli/caesarbell/DOP-4060/index.html)

### Notes:

This enables the language selector for the unified footer on all docs pages and points the href to the A/C of the ticket. This feature is also hidden behind a feature flag that is enabled only for non-prod env at this time.


src/components/DocumentBody.js
```diff
@@ -90,6 +110,17 @@ const DocumentBody = (props) => {
 
   const isInPresentationMode = usePresentationMode()?.toLocaleLowerCase() === 'true';
 
+  const onSelectLocale = (locale) => {
+    const localeHrefMap = {
+      'zh-cn': 'https://mongodbcom-cdn.website.staging.corp.mongodb.com/zh-cn/docs-qa/',
```
seungpark, line 115: can we update the A/C of this ticket? i'm pretty sure we don't want users going to preprd when they switch. should they be navigated to a different page at all? couldnt find a mongodb docs page in chinese. even this page https://www.mongodb.com/zh-cn the docs link goes back to our english docs page

src/components/DocumentBody.js
```diff
@@ -109,8 +140,8 @@ const DocumentBody = (props) => {
         </FootnoteContext.Provider>
       </Widgets>
       {!isInPresentationMode && (
-        <div data-testid="consistent-footer">
-          <UnifiedFooter hideLocale={true} />
+        <div data-testid="consistent-footer" id="footer-container">
+          <UnifiedFooter hideLocale={HIDE_UNIFIED_FOOTER_LOCALE} onSelectLocale={onSelectLocale} />
```
seungpark, line 144: tested with passing in a `locale` prop here. changes the UnifiedFooter language, which i think would be desired

src/components/DocumentBody.js
```diff
@@ -68,10 +69,29 @@ const getAnonymousFootnoteReferences = (index, numAnonRefs) => {
...
   } = props;
+
+  useEffect(() => {
+    // A workaround to remove the other locale options.
+    if (!HIDE_UNIFIED_FOOTER_LOCALE) {
+      const footer = document.getElementById('footer-container');
+      const footerUlElement = footer?.querySelector('ul');
+      if (footerUlElement) {
+        // For DOP-4060 we only want to support English and Simple Chinese (for now)
+        const en = footerUlElement.firstChild;
```
seungpark, line 86: (assuming we only want eng and chinese) instead of assuming first and last, i would construct the children here (as same as in consistentnav) and pipe them in here

src/components/DocumentBody.js
```diff
@@ -68,10 +69,29 @@ const getAnonymousFootnoteReferences = (index, numAnonRefs) => {
...
   const {
     location,
     pageContext: { page, slug, template },
   } = props;
+
+  useEffect(() => {
+    // A workaround to remove the other locale options.
+    if (!HIDE_UNIFIED_FOOTER_LOCALE) {
+      const footer = document.getElementById('footer-container');
+      const footerUlElement = footer?.querySelector('ul');
```
seungpark, line 83: seems like theres multiple `ul` in the footer, but there is only one with `'ul[role=listbox]'`

seungpark (changes requested):
@caesarbell can you verify what the end result of selecting eng/chinese on the dropdown should be? i'm not sure if we should be navigating them to a certain production page, or just changing the footer language, etc.

src/components/DocumentBody.js
seungpark: ah i see this is for testing in stage only for now 👍 however, i cannot access the link above. getting hit with a 502 err

caesarbell: If we look at the A/C [ticket](https://jira.mongodb.org/browse/DOP-4060) it says says the href should point to the following URLs. They were provided in the ticket as well.

caesarbell:
> @caesarbell can you verify what the end result of selecting eng/chinese on the dropdown should be? i'm not sure if we should be navigating them to a certain production page, or just changing the footer language, etc.

This is from the ticket A/C

```
There is a way to swap between the english-language version of the site, located at https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/,  and the Chinese-language version, located at [https://mongodbcom-cdn.website.staging.corp.mongodb.com/zh-cn/docs-qa/](https://mongodbcom-cdn.website.staging.corp.mongodb.com/%7B*%7Dzh-cn%7B*%7D/docs-qa/)
```

src/components/DocumentBody.js
```diff
@@ -68,10 +69,29 @@ const getAnonymousFootnoteReferences = (index, numAnonRefs) => {
 };
 
 const DocumentBody = (props) => {
+  const HIDE_UNIFIED_FOOTER_LOCALE = process.env['GATSBY_HIDE_UNIFIED_FOOTER_LOCALE'] === 'true';
```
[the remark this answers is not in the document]
caesarbell: That is fair. I can move this outside the component scope.

src/components/DocumentBody.js
caesarbell: I might be a bit confused about the above, are you confirming it worked or saying it didn't work?

src/components/DocumentBody.js
caesarbell: @seungpark can you share an example when you say the same as in consistentnav? Yes, we would as of now only show those two options.

src/components/DocumentBody.js
caesarbell: I could make that swap, even tho this one is the first one that appears and will only change if the design of the footer changes and `querySelector` returns the first match.

src/components/DocumentBody.js
caesarbell: @seungpark idk if that is on me or the reporter of the ticket. What is our process for that?

caesarbell: As of now, I was told the 503 is okay, for the page we are sending them to. I can confirm if you prefer.

seungpark: i think this is fine for now, since it is hidden behind a feature flag? but i'm personally confused if we:
1. want to make use out of the UnifedFooter prop `locale` or
2. navigate them to a potential(?) Chinese version of docs site

src/components/DocumentBody.js
seungpark: actually, i'm not sure how these elements are bringing back the values for the `onSelectLocale` so i will leave it as this... 🙏

src/components/DocumentBody.js
caesarbell: I believe it is more #2 that they want. Something similar to the unified footer on the homepage. It allows us to select the language and navigates the user somewhere.

Correct, it is hidden behind a feature flag.

src/components/DocumentBody.js
caesarbell: No worries. I made the change anyways and it is bit more dynamic just incase they are swapped around from `localMap` or if we want to add a new one.

src/components/DocumentBody.js
```diff
@@ -67,11 +68,37 @@ const getAnonymousFootnoteReferences = (index, numAnonRefs) => {
...
     pageContext: { page, slug, template },
   } = props;
+
+  useEffect(() => {
+    // A workaround to remove the other locale options.
+    if (!HIDE_UNIFIED_FOOTER_LOCALE) {
+      const footer = document.getElementById('footer-container');
+      const footerUlElement = footer?.querySelector('ul[role=listbox]');
+      if (footerUlElement) {
+        // For DOP-4060 we only want to support English and Simple Chinese (for now)
```
seungpark, line 86: nice!

seungpark (approved):
changes LGTM ! will follow up with review for autobuilder changes

rayangler (approved):

PR #878, opened 2023-08-31
DOP-3938 Search Results Padding and font changes

anabellabuckvar:
### Stories/Links:

[DOP-3938](https://jira.mongodb.org/browse/DOP-3938)

### Current Behavior:
"Search Results" heading is currently H1, and the padding between the heading and the search results bar is 40px. Font was changed to H3 and bottom padding to 24px as requested.

https://www.mongodb.com/docs/search/?q=atlas

### Staging Links:

[Staging Link](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/newhomepagewhodis/landing/anabella.buckvar/DOP-3938/search/index.html)

### Notes:
UI element listing the number of search results is currently not included in master branch because the master branch currently has unreleased changes related to the search UI. Once the search result count is added, margins must be adjusted again accordingly

PR #930, opened 2023-10-02
DOP-3938: Update spacing on the search results page

anabellabuckvar:
### Stories/Links:

[DOP-3938](https://jira.mongodb.org/browse/DOP-3938)

### Current Behavior:

[Production search results](https://www.mongodb.com/docs/search/?q=atlas)

### Staging Links:

[Search staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/anabella.buckvar/DOP-3938-b/search/index.html)

### Notes:

A [PR](https://github.com/mongodb/snooty/pull/878) was previously opened and merged for this same ticket, but was reverted as there was weird behavior observed. On hover at this [search staging link]( https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/search/?q=mongodb&page=1), the highlighting behavior is off and the borders regressed for the search results compared to those in [production]( https://www.mongodb.com/docs/search/?q=aggregate&page=1).


src/components/SearchResults/SearchResults.js
```diff
@@ -382,7 +382,7 @@ const SearchResults = () => {
       <SearchResultsContainer>
         {/* new header for search bar */}
         <HeaderContainer>
-          <H1>Search Results</H1>
+          <H3 as="h1"> Search Results</H3>
```
seungpark, line 385:
```suggestion
          <H3 as="h1">Search Results</H3>
```

src/components/SearchResults/SearchResults.js
```diff
@@ -382,7 +382,7 @@ const SearchResults = () => {
...
           <SearchInput
             ref={searchBoxRef}
             value={searchField}
```
seungpark, line 388: can we remove those inline `style={}`  (except for the one that reads `display: none;`

for `ResultTag`, the styling is already done in line 222 (already styled div) 

and for LG components, we can use `<Overline className={ cx(css`` ) }>`  see example [here](https://github.com/mongodb/snooty/blob/master/src/components/Paragraph.js#L21)

seungpark: we can combine padding for Overline and ResultTag

seungpark (approved):
header changes LGTM.

cleanup on removing `style` attribute can be made in a follow up PR.

slightly off tangent, but seems like search results in design do not match search results on page in terms of height, font sizing. we can handle this in a different ticket

casthewiz (commented):
I think it's worth consolidating some of the inline styles in this PR as well - the ResultTag style on line 355 of SearchResults.js can be moved to line 224 and placed within the `styled` definition, and likewise for [inline styles on line 444](https://github.com/mongodb/snooty/blob/e59148143a7db4a72b5520bcbe452e93d878fcc9/src/components/SearchResults/SearchResults.js#L444-L447)

src/components/SearchResults/SearchResults.js
anabellabuckvar: what do you mean by combine padding?

src/components/SearchResults/SearchResults.js
```diff
@@ -222,6 +224,12 @@ const StyledTag = styled(Tag)`
 const ResultTag = styled('div')`
   display: flex;
   flex-direction: row;
+  padding-top: 21px;
```
seungpark, line 227:
```suggestion
  padding-top: ${theme.size.default};
  align-items: center;
```

seeing the tags a little off center:

image

seungpark: we should remove the lines 216-218 since ResultTag has the padding already

src/components/SearchResults/SearchResults.js
seungpark: you got it! but seems like it should be adjusted to 16px. left a suggestion above

seungpark (changes requested):
minor changes within the ResultTag! see screenshot below

seungpark (approved):
LGTM besides minor padding on res count

PR #935, opened 2023-10-09
DOP-4063: Reorganize data consumption for preview plugin

rayangler:
### Stories/Links:

DOP-4063

### Staging Links:

This change is strictly for the Gatsby Cloud preview plugin, but not changes should be made. A successful refactoring means that the staging links build with correct side navs (for metadata), pages (for pages), and images (for assets).

[cloud-docs](https://preview-mongodbrayangler.gatsbyjs.io/cloud-docs/gc-test-img/)
[docs](https://preview-mongodbrayangler.gatsbyjs.io/docs/DOP-4063-test/)

### Notes:

* The bulk of this PR is just reorganizing the way data was being consumed from the Snooty Data API so that it was less nested, and easier to maintain. There should be no functional changes other than some additional proactive error handling for incorrect data formats to help improve completeness.

### DIFF
--- a/plugins/gatsby-source-snooty-preview/gatsby-node.js
+++ b/plugins/gatsby-source-snooty-preview/gatsby-node.js
@@ -116,7 +116,9 @@ exports.sourceNodes = async ({
       }
     };
 
-    // Handle data consumed by the Snooty Data API based on data type
+    // Since there's a lot of data incoming from the Snooty Data API, we stream
+    // the data in chunks and parse them as they come instead of fetching everything
+    // as a single JSON response
     const decode = parser();
     decode.on('data', async (_entry) => {
       // Un-nest data
--- a/plugins/gatsby-source-snooty-preview/gatsby-node.js
+++ b/plugins/gatsby-source-snooty-preview/gatsby-node.js
@@ -2,16 +2,13 @@ const { getDataStore } = require('gatsby/dist/datastore');
 const path = require('path');
 const stream = require('stream');
 const { promisify } = require('util');
-const fs = require('fs').promises;
-const { transformBreadcrumbs } = require('../../src/utils/setup/transform-breadcrumbs.js');
 const pipeline = promisify(stream.pipeline);
 const got = require(`got`);
 const { parser } = require(`stream-json/jsonl/Parser`);
-const { saveStaticFiles } = require('../../src/utils/setup/save-asset-files');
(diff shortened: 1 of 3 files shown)


plugins/gatsby-source-snooty-preview/gatsby-node.js
```diff
@@ -120,100 +105,34 @@ exports.sourceNodes = async ({
...
+      await consumeData(entry, {
+        actions,
+        cache,
+        clientAccessToken,
+        createContentDigest,
+        createNodeId,
+        fileWritePromises,
+        getNode,
+        onHandlePage,
+      });
```
rayangler, line 135: I'm not sure how I feel about passing in all of these Gatsby functions and misc. data to this single function. It might be easy to lose track of everything being passed, especially since not every data type handler requires every function. On the other hand, I like that the majority of the logic is now consolidated in a separate file and isn't nested 3 layers deep. 

Happy to bring the logic of `consumeData` back up here if it seems ideal

src/utils/setup/save-asset-files.js
```diff
@@ -37,4 +37,4 @@ const saveStaticFiles = async (staticFiles) => {
   );
 };
 
-module.exports = { saveAssetFiles, saveStaticFiles };
+module.exports = { saveAssetFiles, saveStaticFiles, saveFile };
```
rayangler, line 40: The `saveFile` in `gatsby-node` was the same exact function so I just exported it here. I'm pretty sure they were slightly different at one point 🤷

plugins/gatsby-source-snooty-preview/gatsby-node.js
```diff
@@ -120,100 +105,34 @@ exports.sourceNodes = async ({
     }
     const httpStream = got.stream(url, { headers });
 
+    let pageCount = 0;
+    // Callback function to be ran after a valid page has been found and handled.
+    // Tracks and updates information that spans across several data entries
+    const onHandlePage = (pageTemplate, pageId, pageNodeId) => {
```
rayangler, line 111: This pattern is a little interesting due to `pageCount` and `hasOpenAPIChangelog` being tracked across all pages seen. This is a side effect of having all of the data consumption logic in a separate file

mmeigs (approved):
I cannot find fault. Looks great.

PR #936, opened 2023-10-12
:white_check_mark: DOP-3707 creates additional test for the version dropdown

caesarbell:
### Stories/Links:

DOP-3707

### Current Behavior:

[Prod](https://www.mongodb.com/docs/drivers/node/current/)

### Staging Links:

[Staged Commit](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/c58679996a06128466e9e7f89a1776f36d0cf21a/master/node/caesarbell/DOP-3707/)

### Notes:

This PR adds a focus on the component's behavior rather than its utility.


tests/unit/VersionDropdown.test.js
```diff
@@ -1,66 +1,273 @@
...
-
-      expect(generatePrefix('v2.15', mockSiteMetadata, mockSiteBasePrefix)).toBe(
-        '/bi-connector/docsworker-xlarge/v2.15'
-      );
+      const versionDropdown = screen.queryByRole('button', { name: 'master' });
+      let versionOptionsDropdown = queryElementWithin(versionDropdown, 'listbox');
+      expect(versionOptionsDropdown).not.toBeInTheDocument();
+
+      await act(async () => {
+        await userEvent.click(versionDropdown);
```
seungpark, line 235: minor. don't have to await userEvents. these are non async

tests/unit/VersionDropdown.test.js
```diff
@@ -1,66 +1,273 @@
...
+      mockFetchDocuments.mockClear();
+      mockedFetchDocset.mockClear();
+    });
 
-      expect(generatePrefix('sdk/android/v10.2', mockSiteMetadata, mockSiteBasePrefix)).toBe(
-        '/docs/realm/sdk/android/v10.2'
-      );
+    it('renders correctly', async () => {
+      let tree;
+      await act(() => {
```
seungpark, line 212: minor nitpick. `await act` should be followed with `async` functions (component re-renders on mocked realm functions)

seungpark (approved):
LGTM! minor comments on removing unused `awaits`

tests/unit/VersionDropdown.test.js
caesarbell: In this case you are correct it doesn't need `await`. You will only need to use it when `userEvent` has options. i.e. `options.delay`. Then and only then we must make sure to use `await`.

tests/unit/VersionDropdown.test.js
caesarbell: May I ask why a `async` function need to follow? i.e `async functions (component re-renders on mocked realm functions)`. I ask because inside the `act`, I am calling the one function and the `await act` is what I am waiting on to move to the next step.

seungpark: seems like this is the (undocumented) standard when using  `act` with await. (needs to be a async fn inside)

the mountConsumer has mocked async functions.

so it should be

```
await act( async () => {

})
```

was just getting linting errors because the function inside is assumed to be synchronous:

image

caesarbell: Gotcha. Okay thanks for sharing this insight with me. Definitely undocumented, I could not locate info around that, and I am curious why I wasn't getting linting errors, or maybe I didn't even notice. Thank you for this feedback. I am going to make the change now.

PR #941, opened 2023-10-18
DOP-4077: Create "Only" button for selected subfacets

mayaraman19:
### Stories/Links:

[DOP-4077](https://jira.mongodb.org/browse/DOP-4077)

### Current Behavior:

Currently on the search page there is no "Only" button next to subfacets that, when selected, will ensure that only that subfacet is selected out of its siblings. This PR adds that "Only" button to correspond to the A/C of the ticket.

### Staging Links:

[Staging link](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/maya/DOP-4077/search/)

### Notes:
The style of the "Only" button is a little bit wonky...any advice on making it cleaner and prettier would be appreciated, I'm not super sure how to cleanly line it up with the subfacet title.


src/components/SearchResults/Facets/FacetGroup.js
```diff
@@ -53,7 +54,7 @@ const FacetGroup = ({ facetOption: { name, id, options }, isNested = false }) =>
         glyph: 'ChevronUp',
         text: 'Show less',
       };
-
+  const siblingsSelected = numSelectedChildren > 1 ? true : false;
```
rayangler, line 57: The ternary here might be redundant since we're just returning a boolean equivalent to the condition. It might make sense to simplify

```suggestion
  const siblingsSelected = numSelectedChildren > 1;
```

src/components/SearchResults/Facets/FacetValue.js
```diff
@@ -10,7 +12,40 @@ const checkboxStyle = css`
...
+    align-items: center;
+    color: ${palette.gray.base};
+  }
+  :hover {
+    opacity: 1;
+    border: 0px;
+    box-shadow: none;
+  }
+  :hover + div {
+    border: none;
```
rayangler, line 40: Are most of these styles being added to override the LG Button component's styling? Since the "Only" button doesn't really share the same styling as LG's `Button` component, I wonder if it would make sense to just use a simple `span` element or something. That way, we can just add css to style it similar to the design, instead of needing to override so much. What do you think?

src/components/SearchResults/Facets/FacetValue.js
```diff
@@ -103,19 +142,48 @@ const FacetValue = ({ facetValue: { name, facets, key, id }, isNested = false })
     [handleFacetChange, key, id, nestedSubFacets]
   );
 
+  const onClickHandler = () => {
```
rayangler, line 145: (Nit; non-blocking) Since `onClickHandler` is fairly similar in logic to `onChangeHandler`, but serve different purposes (one is sibling-focused, while the other is children-focused), it might be good to leave a small comment to outline the key difference between the two. Or maybe we can rename the two functions to make what they do clearer?

src/components/SearchResults/Facets/FacetValue.js
```diff
@@ -10,7 +12,40 @@ const checkboxStyle = css`
     font-size: 13px;
     margin-bottom: 8px;
   }
+  :hover + button {
+    opacity: 1;
   }
+  flex: 1;
+`;
+
+const onlyButtonStyle = css`
```
rayangler, line 21: The design has a `font-weight` of 400 for the "Only" text. Might be good to include as part of the style

src/components/SearchResults/Facets/FacetValue.js
```diff
@@ -10,7 +12,40 @@ const checkboxStyle = css`
     font-size: 13px;
     margin-bottom: 8px;
```
rayangler, line 13: > I'm not super sure how to cleanly line it up with the subfacet title.

I think part of the reason that the "Only" text isn't 100% center aligned to the subfacet title is because of this margin. Have you tried including a similar margin to `onlyButtonStyle` to see if that helps? Not sure if you meant something else by how they line up!

rayangler (commented):
Functionality seems good based on what I've seen so far! 

Would it be possible to run `npm ci --legacy-peer-deps` to install the latest dependencies and rebuild the staging link? Looks like the checkboxes on the current staging link may be outdated

src/components/SearchResults/Facets/FacetGroup.js
mayaraman19: New staging link: https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/maya/DOP-4077/index.html (will also update above)

src/components/SearchResults/Facets/FacetValue.js
mayaraman19: Yes, I think that sounds good!

src/components/SearchResults/Facets/FacetValue.js
```diff
@@ -103,19 +130,48 @@ const FacetValue = ({ facetValue: { name, facets, key, id }, isNested = false })
...
+        <Checkbox
+          className={cx(checkboxStyle)}
+          label={name}
+          onChange={updateChildren}
+          checked={isChecked}
+          id={fullFacetId}
+          indeterminate={isIndeterminate}
+        />
+        {isNested && isChecked && siblingsSelected && (
+          <span onClick={updateSiblings} className={onlyButtonStyle}>
```
branberry, line 160: I noticed that the Only button covers some of the facet text when I hover over it
image

I think you can add a `gap` property to the `container` class. I set `gap: 10px` and here's what it looks like:

image

I still don't have it fully centered with this change, but it looks like a step in the right direction

branberry (commented):
Hey Maya! Looks good, just a few small comments

src/components/SearchResults/Facets/FacetValue.js
```diff
@@ -10,7 +11,34 @@ const checkboxStyle = css`
...
+    opacity: 1;
+    cursor: pointer;
+  }
+  line-height: 20px;
+`;
+
+const container = css`
+  div:hover + span {
+    opacity: 1;
+  }
```
rayangler, line 39: (Nit) Is this css still necessary, or is this already handled in `onlyButtonStyle`?

rayangler (approved):
LGTM! Just one nit, but shouldn't block merging

src/components/SearchResults/Facets/FacetValue.js
mayaraman19: yeah, there's some weird behavior without it where the button kinda flickers or won't appear when you want it to

PR #948, opened 2023-10-30
Dop 3700 new

mayaraman19:
### Stories/Links:

[DOP-3700](https://jira.mongodb.org/browse/DOP-3700)

### Current Behavior:

Currently, on every page except https://mongodb.com/docs/ there is a large white space at the bottom of each page of the docs on Firefox. This PR removes one line of styling that fixes the gap issue on Firefox; I believe it doesn't affect any other behavior but I could be wrong.

### Staging Links:

[Staging environment](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/maya/DOP-3700-new/)

### DIFF
--- a/src/components/Header/index.js
+++ b/src/components/Header/index.js
@@ -15,7 +15,6 @@ const StyledHeaderContainer = styled.header(
   grid-area: header;
   top: 0;
   z-index: 1000;
-  ${props.template !== 'landing' || !CHATBOT_ENABLED ? 'position: sticky;' : ''}
   `
 );
 
--- a/src/components/Header/index.js
+++ b/src/components/Header/index.js
@@ -8,15 +8,11 @@ import { isBrowser } from '../../utils/is-browser';
 import useSnootyMetadata from '../../utils/use-snooty-metadata';
 import { useMarianManifests } from '../../hooks/use-marian-manifests';
 
-const CHATBOT_ENABLED = process.env['GATSBY_SHOW_CHATBOT'] === 'true';
-
-const StyledHeaderContainer = styled.header(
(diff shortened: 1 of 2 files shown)


src/components/Header/index.js
```diff
@@ -8,16 +8,11 @@ import { isBrowser } from '../../utils/is-browser';
...
 
-const CHATBOT_ENABLED = process.env['GATSBY_SHOW_CHATBOT'] === 'true';
-
-const StyledHeaderContainer = styled.header(
-  (props) => `
+const StyledHeaderContainer = styled.header(`
   grid-area: header;
   top: 0;
   z-index: 1000;
-  ${props.template !== 'landing' || !CHATBOT_ENABLED ? 'position: sticky;' : ''}
```
seungpark, line 18: this leaves the sticky header on top so we still need this. this change causes a white blank space when scrolling down (see top left):

image

seungpark: seems like the nav component is overflowing the `StyledHeaderContainer` and the grid layout is causing an empty space:

image

my suggestion is to add a `overflow: clip;` to the [global grid](https://github.com/mongodb/snooty/blob/bc9a4d07d72aea42699c88e18c980e9cd9f3ec55/src/layouts/index.js#L69). tested this in browser, but would verify this doesn't have any breaking changes

seungpark (changes requested):

mayaraman19:
I don't see any breaking changes with the `overflow: clip;` change.

seungpark (approved):
LGTM 👍

PR #949, opened 2023-10-31
DOP-4089: Responsive facet filters

seungpark:
### Stories/Links:

DOP-4089

This PR introduces responsive facet filters as [per Spec](https://docs.google.com/document/d/1c9P5_wnlLswoZftJbZG-baqG1Fgs26lO1gv7ExgT3Ss/edit#heading=h.ytznsphw0fe2)

Facet behavior is that each selection should also narrow down the available facet filters. IE each result contains all of the facets selected, and each facet selectable (at any stage) has a search result.

### Current Behavior:

[Current search on master branch with env flag turned on (GATSBY_FEATURE_FACETED_SEARCH=true)
](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/681ce6171672b870647891839397988fc45ad07f/newhomepagewhodis/landing/seung.park/master/search/index.html?q=$and)

In the link above, same master list of facets is always available, and can lead to no results

### Staging Links:

[staged changes with responsive facets](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/2e270932ab29ba09a1a99e6955f126b0323646b7/newhomepagewhodis/landing/seung.park/DOP-4089/search/index.html?q=%24and)

[(env flag turned off for good measure)](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/50d56f3d2bc3497e91dc07d4c8371ec5ebafeed5/newhomepagewhodis/landing/seung.park/DOP-4089/search/index.html?q=%24and&page=1&searchProperty=atlas-cli-upcoming)

### Notes:
- order of facets are not strict from search API server. this should be updated in a follow up ticket to ensure facet list is consistent


mmeigs (commented):
This looks great! I'm wondering about one CSS thing that hasn't been an issue until now. Only because we are re-rendering everything so much for the responsive facets.

I'm wondering if we can set a min-width or something for the div with the search result count. Every render, the filter pills jump from the left side of the page to about 50px or so to the right when the count is finally rendered. If we could lessen this jump it would be great.

Screenshot 2023-11-01 at 9 24 49 AM

rayangler (approved):
LGTM. 

Only caught this in this PR staging link, but it seems like there's missing spacing between the results count and the selected filters on the staging link where the feature flag is turned **off**. I don't think it's related to your current changes though, so I'm down to fix this in a separate PR if that seems more sensible.

Screen Shot 2023-11-02 at 11 57 56 AM

seungpark:
> LGTM.
> 
> Only caught this in this PR staging link, but it seems like there's missing spacing between the results count and the selected filters on the staging link where the feature flag is turned **off**. I don't think it's related to your current changes though, so I'm down to fix this in a separate PR if that seems more sensible.
> 
> Screen Shot 2023-11-02 at 11 57 56 AM

thanks for the catch! this is actually not from this PR, but seems like PROD search page hasn't been deployed in awhile. i pushed [this commit](https://github.com/mongodb/snooty/pull/949/commits/b378aec92291424b7d77f70ea04c1e41cb223299), which changes all styling passed to LG components to be called with lg/emotion css (vs currently emotion/react css)

seungpark:
@rayangler going to merge this as this feature is not yet released, but we should fix the indentation error on master branch for v1 search filters.
there will be a follow up PR to incorporate the changes outlined in the updated JIRA ticket (as well as the [spec](https://docs.google.com/document/d/1c9P5_wnlLswoZftJbZG-baqG1Fgs26lO1gv7ExgT3Ss/edit#heading=h.mbb4punhvs05) and [discussion doc](https://docs.google.com/document/d/1E8yIWbDR3B6GlBxM23Ph-k4mz1mc7BRMQCpeX7_z3Hk/edit#heading=h.gt9cllte775l))

PR #953, opened 2023-11-03
DOP-4081: version dropdown for facet filters

seungpark:
### Stories/Links:

DOP-4081

### Staging Links:

[regression check for existing search with feature flag off `GATSBY_FEATURE_FACETED_SEARCH=false`](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/newhomepagewhodis/landing/seung.park/DOP-4081/search/?q=compass&page=1&searchProperty=atlas-master)

[staging link with test data used ](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/bf468afd223919483e1a071ee959d062620fafb7/newhomepagewhodis/landing/seung.park/DOP-4081/search/index.html?q=%24and&page=1&facets.target_product%3Edocs%3Esub_product=v7.0&facets.target_product=docs)

### Notes:
- This is using test data, as `versions` is not available as facet options yet.
- When a version is selected, parent facet selections are pruned when sending the API request to /search and v2/search/meta since 
- When a parent facet selection is deselected, the version is also deselected from UI and from API requests

### DIFF
--- a/src/components/SearchResults/Facets/useFacets.js
+++ b/src/components/SearchResults/Facets/useFacets.js
@@ -1,6 +1,7 @@
 import { useEffect, useState } from 'react';
-import { assertTrailingSlash } from '../../../utils/assert-trailing-slash';
-import { MARIAN_URL } from '../../../constants';
+// import { assertTrailingSlash } from '../../../utils/assert-trailing-slash';
+import { statusV2 } from '../../../../tests/unit/data/SearchResults.test.json';
+// import { MARIAN_URL } from '../../../constants';
 
 const useFacets = () => {
   const [facets, setFacets] = useState([]);
@@ -8,13 +9,16 @@ const useFacets = () => {
   // Fetch facets
   useEffect(() => {
     const fetchFacets = async () => {
-      try {
-        const result = await fetch(assertTrailingSlash(MARIAN_URL) + 'v2/status');
-        const jsonResult = await result.json();
-        setFacets(jsonResult);
-      } catch (err) {
-        console.error(`Failed to fetch facets: ${err}`);
-      }
+      // DOP-4081
+      // TODO: remove after testing version facets
+      setFacets(statusV2);
+      // try {
+      //   const result = await fetch(assertTrailingSlash(MARIAN_URL) + 'v2/status');
+      //   const jsonResult = await result.json();
+      //   setFacets(jsonResult);
+      // } catch (err) {
+      //   console.error(`Failed to fetch facets: ${err}`);
+      // }
     };
     fetchFacets();
   }, []);
--- a/src/components/SearchResults/Facets/useFacets.js
+++ b/src/components/SearchResults/Facets/useFacets.js
@@ -1,7 +1,6 @@
 import { useEffect, useState } from 'react';
-// import { assertTrailingSlash } from '../../../utils/assert-trailing-slash';
-import { statusV2 } from '../../../../tests/unit/data/SearchResults.test.json';
-// import { MARIAN_URL } from '../../../constants';
+import { assertTrailingSlash } from '../../../utils/assert-trailing-slash';
+import { MARIAN_URL } from '../../../constants';
 
 const useFacets = () => {
   const [facets, setFacets] = useState([]);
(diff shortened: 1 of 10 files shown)


src/components/SearchResults/Facets/useFacets.js
```diff
@@ -1,20 +1,24 @@
...
   useEffect(() => {
     const fetchFacets = async () => {
-      try {
-        const result = await fetch(assertTrailingSlash(MARIAN_URL) + 'v2/status');
-        const jsonResult = await result.json();
-        setFacets(jsonResult);
-      } catch (err) {
-        console.error(`Failed to fetch facets: ${err}`);
-      }
+      // DOP-4081
```
seungpark, line 12: TODO: revert test data before merge

src/components/SearchResults/SearchResults.js
```diff
@@ -16,6 +16,8 @@ import { theme } from '../../theme/docsTheme';
 import { reportAnalytics } from '../../utils/report-analytics';
 import { escapeHtml } from '../../utils/escape-reserved-html-characters';
 import { searchParamsToMetaURL, searchParamsToURL } from '../../utils/search-params-to-url';
+// TODO: remove after testing
+import { statusV2 } from '../../../tests/unit/data/SearchResults.test.json';
```
seungpark, line 20: TODO: revert test data before merge

src/utils/search-params-to-url.js
```diff
@@ -17,6 +18,38 @@ const getFilterParams = (searchParams) => {
...
+ * Removes parent filter selections in search params if any
+ * since existence of child filter selection already implies
+ * the existence of parent filter
+ * ie. target_products>atlas>versions=master
+ *     already implies target_products=atlas
+ *     within the data returned
+ *
+ * @param {URLSearchParams} searchParams
+ */
+function removeParentSelections(searchParams) {
```
rayangler, line 31: Just curious: what is the benefit of removing the parent from the params?

src/utils/search-params-to-url.js
```diff
@@ -1,11 +1,12 @@
 import { MARIAN_URL } from '../constants';
-import { FACETS_KEY_PREFIX } from '../components/SearchResults/SearchContext';
+import { FACETS_KEY_PREFIX, FACETS_LEVEL_KEY } from '../components/SearchResults/SearchContext';
 import { assertTrailingSlash } from './assert-trailing-slash';
 
 const TERM_PARAM = 'q';
 const PAGE_PARAM = 'page';
 const V1_SEARCH_FILTER_PARAM = 'searchProperty';
 const V2_SEARCH_FILTER_PREFIX = FACETS_KEY_PREFIX;
+const SINGLE_SELECT_FIELDS = ['version'];
```
rayangler, line 9: (Nit; non-blocking) It might make sense to use `VERSION_GROUP_ID` in the array so that we have a single reference of the same constant.

It might also be good to bring `VERSION_GROUP_ID` to this file, or have a single util file dedicated to defining these keys/constants. Keeping all defined constants together in a single place would allow us to keep track of which search-related keys are used where more easily, rather than having to go to different files for constants that are used similarly.

What do you think?

src/utils/search-params-to-url.js
seungpark: single select fields for a subfacet should already imply within the data that is tagged, that the parent property exists. that said, the API does an `or` combination of all filters under the same main facet (ie. genre, target product)

so `target_product=atlas` and `target_product>atlas>version=master` will be translated to `target_product = atlas OR target_product>atlas>version=master` which nullifies the second criteria in an OR selection.

to mitigate this, we remove any single select fields' parent facet, since that the parent facet is already implied within the name of the child facet.

src/utils/search-params-to-url.js
seungpark: good call! done! latest commit includes a util file to store all the constants for search server.

src/utils/search-params-to-url.js
rayangler: Oh, gotcha, it facilitates the `OR` logic. Thanks for the example!

rayangler (approved):

PR #955, opened 2023-11-09
DOP-4112: Adjusting size of chatbot loading skeleton to not overflow input bar

mayaraman19:
### Stories/Links:

[DOP-4112](https://jira.mongodb.org/browse/DOP-4112)

### Current Behavior:

Currently, the loading skeleton for the chatbot is large enough to overflow the input bar. This changes the skeleton to have a height of 48px and be properly centered within the input bar.

### Staging Links:

[Staging link](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/maya/DOP-4112/)

### DIFF
--- a/src/components/ChatbotUi.js
+++ b/src/components/ChatbotUi.js
@@ -1,4 +1,4 @@
-import React, { Suspense, lazy, useEffect, useState } from 'react';
+import React, { Suspense, lazy } from 'react';
 import Skeleton from 'react-loading-skeleton';
 import { css } from '@emotion/react';
 import styled from '@emotion/styled';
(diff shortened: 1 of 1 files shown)


src/components/ChatbotUi.js
```diff
@@ -65,6 +66,13 @@ const StyledChatBotUiContainer = styled.div`
     }
   }
 
+  // Styling the chatbot's loading skeleton
+  > span {
+    display: flex;
+    height: 48px;
```
caesarbell, line 72: @mayaraman19 just curious, do we need the height defined on line 72 if we are passing in the height props in the `<Skeleton />` component?

mayaraman19: yes, this is for this container that appears surrounding the loading skeleton which by default is 4px larger than the skeleton itself.

branberry (approved):
LGTM!

PR #966, opened 2023-12-04
DOP-3708: scroll heading into view if hash exists in URL

seungpark:
### Stories/Links:

DOP-3708

### Current Behavior:

[server docs with anchor/hash in URL](https://www.mongodb.com/docs/manual/tutorial/query-documents/#specify-and-conditions)

### Staging Links:

[staged change with anchor/hash in URL](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/seung.park/DOP-3708/tutorial/query-documents/index.html#specify-and-conditions)

### Notes:
- There are considerably two scroll behaviors happening:
- 1 When the browser loads the SSR'ed page and scrolls to the element found (browser native function)
- 2 When the component loads and calls the newly added function

- adding a DO NOT MERGE label as this is a [SPIKE] ticket

### DIFF
--- a/src/hooks/use-hash-anchor.js
+++ b/src/hooks/use-hash-anchor.js
@@ -10,7 +10,6 @@ const useHashAnchor = (id, ref) => {
       return;
     }
     setTimeout(() => {
-      console.log('scrolling');
       window.scrollTo(0, ref.current.offsetTop);
     }, 40);
   }, [hash, id, ref]);


src/components/Heading.js
```diff
@@ -42,6 +44,22 @@ const Heading = ({ sectionDepth, nodeData, page, ...rest }) => {
   const hasSelectors = selectors && Object.keys(selectors).length > 0;
   const shouldShowMobileHeader = !!(isPageTitle && isTabletOrMobile && hasSelectors);
 
+  const { hash } = useLocation();
+  const headingRef = useRef();
+  const bannerOffset = parseInt(
```
rayangler, line 49: We have a `theme.size.stripUnit` function that might accomplish the same logic here. Not sure if that helps!

rayangler (commented):
I think this solution makes sense, given the circumstances. It just seems like a limitation of tabbed content

src/components/Heading.js
seungpark: realized the banner and navbar is already taken into account in the offsetTop so removed! but good find on the `stripUnit` fn thank you

src/components/Heading.js
```diff
@@ -42,6 +44,22 @@ const Heading = ({ sectionDepth, nodeData, page, ...rest }) => {
...
+    theme.header.navbarScrollOffset.slice(0, theme.header.navbarScrollOffset.indexOf('px'))
+  );
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
seungpark: great call! moved to a custom hook and used it first on permalinks (seems the most necessary place to have it)

src/hooks/use-hash-anchor.js
```diff
@@ -0,0 +1,19 @@
...
+const useHashAnchor = (id, ref) => {
+  const { hash } = useLocation();
+
+  useEffect(() => {
+    const hashId = hash?.slice(1);
+    if (!hash || id !== hashId) {
+      return;
+    }
+    setTimeout(() => {
+      console.log('scrolling');
```
rayangler, line 13: We can remove this

rayangler (approved):
Cool, seems like it works well! Thanks for working on this

PR #967, opened 2023-12-06
DOP-4156: Redesign feedback tab button

rayangler:
### Stories/Links:

DOP-4156

### Current Behavior:

[Server docs](https://www.mongodb.com/docs/manual)

### Staging Links:

[Server docs](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/raymund.rodriguez/DOP-4156/index.html)

### Notes:

* Implements the redesign for the feedback tab/button.
* The button is supposed to be very similar to the style of the chatbot button, as seen in the Figma file. However, the button does not seem to be readily available to us as a variant, so I tried to get it as similar as possible for now.
* Moved some rendered text around to a constants file so that both the components and tests can share the same reference easily.

### DIFF
--- a/src/components/Widgets/FeedbackWidget/FeedbackCard.js
+++ b/src/components/Widgets/FeedbackWidget/FeedbackCard.js
@@ -13,17 +13,11 @@ import useNoScroll from './hooks/useNoScroll';
 const FloatingContainer = styled.div`
   position: fixed;
   z-index: 14;
-  right: 16px;
-
-  @media ${theme.screenSize.upToLarge} {
-    top: 40%;
-  }
-
-  @media ${theme.screenSize.largeAndUp} {
-    bottom: 40px;
-  }
+  bottom: ${theme.size.large};
+  right: ${theme.size.large};
 
   @media ${theme.screenSize.upToSmall} {
+    position: fixed;
     right: 0;
     top: ${({ top }) => top};
   }
@@ -44,6 +38,7 @@ const Card = styled(LeafygreenCard)`
     height: calc(100vh - ${({ top }) => top});
     width: 100vw;
     border-radius: 0;
+    border-width: 0px;
     padding-top: 40px;
   }
 `;
--- a/src/components/Widgets/FeedbackWidget/FeedbackTab.js
+++ b/src/components/Widgets/FeedbackWidget/FeedbackTab.js
@@ -4,6 +4,7 @@ import { palette } from '@leafygreen-ui/palette';
 import Icon from '@leafygreen-ui/icon';
 import { theme } from '../../../../src/theme/docsTheme';
 import { useFeedbackContext } from './context';
+import { text } from './constants';
 
 const containerStyle = css`
   display: flex;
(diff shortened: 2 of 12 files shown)


src/components/Widgets/FeedbackWidget/FeedbackButton.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
```
rayangler, line 1: This file was renamed from `FeedbackTab`... again, not sure why the file is renamed in the [commit](https://github.com/mongodb/snooty/pull/967/commits/242e34168fb2aae95b4bec520ac44dfee515fd20), but not in the PR.

src/components/Widgets/FeedbackWidget/FeedbackCard.js
```diff
@@ -13,17 +13,11 @@ import useNoScroll from './hooks/useNoScroll';
...
-  right: 16px;
-
-  @media ${theme.screenSize.upToLarge} {
-    top: 40%;
-  }
-
-  @media ${theme.screenSize.largeAndUp} {
-    bottom: 40px;
-  }
+  bottom: ${theme.size.large};
```
seungpark, line 16: are we doing a feature flag or a feature branch? this will potentially change the current card (?)

src/components/Widgets/FeedbackWidget/FeedbackButton.js
```diff
@@ -0,0 +1,54 @@
...
+
+const starIconStyle = css`
+  color: ${palette.blue.light1};
+`;
+
+const FeedbackButton = () => {
+  const { feedback, initializeFeedback } = useFeedbackContext();
+  return (
+    !feedback && (
+      <div className={cx(containerStyle)} onClick={() => initializeFeedback()}>
```
seungpark, line 46: should we be utilizing lg/button here. will also do away with the custom hover css

seungpark (changes requested):
this mostly LGTM. question about feature branch vs flag below, and a minor change for using LG/button vs custom div

src/components/Widgets/FeedbackWidget/FeedbackButton.js
rayangler: I tried it with the LG button component, and the styling is different enough that it would warrant more css just to restyle everything. LG Button also doesn't support a blue outline variant.

For what it's worth, this is based on an entirely new/different LG component that is specifically for the chatbot, and isn't part of LG's main library. I based this css on their [Figma design file](https://www.figma.com/file/L41rvEwQmJJ8VzO4hCPc7Y/LeafyGreen-Chat-Library?node-id=494%3A34308&mode=dev) and their [source code](https://github.com/10gen/lg-chat/blob/main/packages/fixed-chat-window/src/ChatTrigger/ChatTrigger.tsx).

seungpark: Fair, this would be overwriting the button styling pretty thoroughly. That said, might be a good idea to have a re-usable FAB button (for chatbot UI) in the future

seungpark (approved):

PR #972, opened 2023-12-12
DOP-4193: removed unused footer component

anabellabuckvar:
### Stories/Links:

[DOP-4193](https://jira.mongodb.org/browse/DOP-4193)

### Current Behavior:
[Master Staging link](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/anabella.buckvar/master/index.html)

### Staging Links:

[Staging link](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/anabella.buckvar/DOP-4193/index.html)

### Notes:
No functionality or UI components should be changed, this PR only deletes unused code.
Some of the tests in `FeedbackWidget.test.js` were flawed or improperly described, so those were cleaned up, too.


src/components/Footer.js
```diff
@@ -1,12 +0,0 @@
-import React from 'react';
-import loadable from '@loadable/component';
-import PropTypes from 'prop-types';
-const FeedbackFooter = loadable(() => import('./Widgets/FeedbackWidget/FeedbackFooter'));
```
rayangler, line 4: Is the `FeedbackFooter` a component that we can remove as well? I don't see any other place where the component is used, but please feel free to double-check me on this

anabellabuckvar: You're right, the FeedbackFooter code can also be deleted; I missed  that at first because some of the tests were failing when I deleted it, but I realized those specific tests were unnecessary so I've removed them

tests/unit/FeedbackWidget.test.js
```diff
@@ -75,6 +70,15 @@ describe('FeedbackWidget', () => {
   afterEach(clearMockScreenshotFunctions);
 
   describe('FeedbackTab (Desktop Viewport)', () => {
+    //fix this
```
rayangler, line 73: Just curious what exactly should be fixed? Fwiw, some of these tests are already being altered for the 5-star feedback update

tests/unit/FeedbackWidget.test.js
```diff
@@ -41,13 +36,13 @@ async function mountFormWithFeedbackState(feedbackState = {}, options = {}) {
...
         }}
-        hideHeader={hideHeader}
+        //hideHeader={hideHeader}
       >
         <FeedbackForm />
         <div>
           <FeedbackTab />
           <Heading nodeData={headingData} sectionDepth={1} />
-          <FeedbackFooter />
+          {/* <FeedbackFooter /> */}
```
rayangler, line 45: Is there a reason to keep these two lines commented?

tests/unit/FeedbackWidget.test.js
anabellabuckvar: It was an extraneous comment to myself, just deleted it! I assumed so, but I figured I'd clean it up anyways and make sure the tests all still passed when set up properly

tests/unit/FeedbackWidget.test.js
```diff
@@ -41,13 +36,12 @@ async function mountFormWithFeedbackState(feedbackState = {}, options = {}) {
           url: 'https://docs.mongodb.com/test',
           docs_property: 'test',
         }}
-        hideHeader={hideHeader}
+        //hideHeader={hideHeader}
```
rayangler, line 39: (Nit) Looks like this comment is also a leftover

rayangler (approved):
Should be good with one non-blocking nit to remove a leftover comment. Thanks for working on this!

tests/unit/FeedbackWidget.test.js
anabellabuckvar: I left it in intentionally, as someone will probably want to add tests that the hideHeader parameter works at some point. I can remove it if you think it's unecessary?

tests/unit/FeedbackWidget.test.js
anabellabuckvar: the second one has been removed, see below for my reason behind keeping the above line commented

tests/unit/FeedbackWidget.test.js
rayangler: Hm, I don't see any instances in which `hideHeader` is actually used, and I don't think that prop currently does anything... I'm okay with keeping it for now if we want, and I can handle double-checking as part of 5-star feedback work

PR #977, opened 2023-12-20
DOP-4176: SSG Legacy Dropdown

mayaraman19:
### Stories/Links:

[DOP-4176](https://jira.mongodb.org/browse/DOP-4176)

### Current Behavior:

Currently, our legacy dropdown fetches data during runtime. This PR changes this so that data is sourced during build time in `sourceNodes()` in `gatsby-node.js`, then queried at runtime using `useStaticQuery()`.

The legacy dropdown should look the same as it does on [mongodb.com.docs/legacy](https://mongodb.com/docs/legacy). If you want to test what you're seeing, edit the `displayName` field in `plugins/utils/docsets.js` to be something different, rebuild, and check the dropdown again.

### Staging Links:

[Legacy page on landing](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/maya/DOP-4176/legacy/index.html)

### Notes:

When running `npm run test` there is a small error that I haven't seen before:
```
A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them.
```
However, this doesn't occur when running `npm run test DeprecatedVersionDropdown.test.js` and occurs on my `master` branch of snooty as well so it doesn't have anything to do with this PR, but figured I would note it somewhere.


plugins/utils/docsets.js
```diff
@@ -0,0 +1,22 @@
+const createDocsetNodes = async ({ db, createNode, createNodeId, createContentDigest }) => {
+  // Get all MongoDB products for the sidenav
+  const docsets = await db.realmInterface.fetchDocsets();
+  docsets.forEach((docset) => {
+    createNode({
+      children: [],
+      id: createNodeId(`docsetInfo-${docset.displayName}`),
```
rayangler, line 7: (Super nit; non-blocking) I think it might be better to use the `repoName` or `project` as part of the id instead of `displayName`. We currently have db indexes in place to enforce that those fields are unique

src/components/DeprecatedVersionSelector.js
```diff
@@ -77,16 +77,16 @@ const DeprecatedVersionSelector = ({ metadata: { deprecated_versions: deprecated
...
   useEffect(() => {
     if (reposDatabase) {
-      fetchDocsets(reposDatabase).then((resp) => {
-        const reposBranchesMap = keyBy(resp, 'project');
-        const reposBranchesMapWithOldGen = addOldGenToReposMap(reposBranchesMap);
-        setReposMap(reposBranchesMapWithOldGen);
-      });
+      const reposBranchesMap = keyBy(reposBranches, 'project');
+      const reposBranchesMapWithOldGen = addOldGenToReposMap(reposBranchesMap);
+      setReposMap(reposBranchesMapWithOldGen);
```
rayangler, line 87: (Looking for confirmation on intended behavior) Will this change mean that we are no longer also fetching data client-side? My assumption was that the ticket wanted data from build time to initially populate the options for the dropdown, but to also keep the separate client-side fetch to ensure the latest data/options are present. Otherwise, the landing site would need to be rebuilt whenever there's an update.

rayangler (commented):
I think this looks good overall, but just wanted to confirm the intended behavior for this component in terms of data.

Regarding the error message from testing, the error itself can happen for a variety of reasons, if I remember correctly (leftover database connection that wasn't closed, some async or timed process that wasn't awaited/resolved properly in a test, etc.). It'd be ideal to not have the error if possible, but if this PR didn't cause the issue, then I don't think this should be a blocker

src/components/DeprecatedVersionSelector.js
mayaraman19: oh, rereading the ticket i think you're right. let me look into this again

src/components/DeprecatedVersionSelector.js
```diff
@@ -80,13 +84,17 @@ const DeprecatedVersionSelector = ({ metadata: { deprecated_versions: deprecated
...
-      fetchDocsets(reposDatabase).then((resp) => {
-        const reposBranchesMap = keyBy(resp, 'project');
-        const reposBranchesMapWithOldGen = addOldGenToReposMap(reposBranchesMap);
-        setReposMap(reposBranchesMapWithOldGen);
-      });
+      fetchDocsets(reposDatabase)
+        .then((resp) => {
+          const reposBranchesMap = keyBy(resp, 'project');
+          const reposBranchesMapWithOldGen = addOldGenToReposMap(reposBranchesMap);
+          if (reposBranchesMap.size > 0) setReposMap(reposBranchesMapWithOldGen);
```
mayaraman19, line 91: note: i included this conditional because for some reason the tests on displaying products that have only one legacy version don't work properly without it.

rayangler (commented):
Mostly lgtm! Just one minor comment

src/components/DeprecatedVersionSelector.js
```diff
@@ -80,13 +84,17 @@ const DeprecatedVersionSelector = ({ metadata: { deprecated_versions: deprecated
...
+        })
+        .catch((error) => {
+          setReposMap(reposBranchesBuildDataMapWithOldGen);
+        });
```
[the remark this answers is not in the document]
rayangler: Nevermind, I just realized that the function is already wrapped in a try catch. Having a `console.error` here may potentially be redundant. I'm okay with keeping this as-is for now!

rayangler (approved):
LGTM!

src/components/DeprecatedVersionSelector.js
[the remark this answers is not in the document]
mayaraman19: Actually, after merging it seems like this `setReposMap` call was causing an infinite loop... I think I'll replace it with a `console.error` call.

mayaraman19:
somehow I missed that clicking on the dropdown when relying on build-time data would result in an infinite loop; that has been fixed.

[Staging link (defaulting to build-time data - content should appear in the dropdown + an error should appear in the console)](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/maya/DOP-4176/legacy/index.html)

PR #978, opened 2024-01-03
DOP-4151: Update product drop-down to use displayName value for labels

anabellabuckvar:
### Stories/Links:

[DOP-4151
](https://jira.mongodb.org/browse/DOP-4151)

### Current Behavior:

[Current legacy documentation site](https://www.mongodb.com/docs/legacy/)

### Staging Links:

[Legacy documentation staging site](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/anabella.buckvar/DOP-4151-b/index.html)

### Notes:

The Kubernetes-Operator product is included in the deprecated versions array and has an entry in `pool.repos_branches`, but is not appearing, likely due to some keying error, in the legacy dropdown on staging sites or in production. This should be further investigated and fixed. If desired, I will make a follow up ticket for that issue specifically.

Not all `displayName` fields are accurate. These will be updated manually in the `pool.repos_branches` collection as part of another ticket


src/init/DocumentDatabase.js
```diff
@@ -37,6 +37,10 @@ class RealmInterface {
     return this.realmClient.callFunction('fetchDocset', siteMetadata.reposDatabase, matchConditions);
   }
 
+  async fetchDocsets() {
+    return this.realmClient.callFunction('fetchDocsets', siteMetadata.reposDatabase);
```
mayaraman19, line 41: Unfortunately after my [PR](https://github.com/mongodb/snooty/pull/977) was merged it seems like the page is now broken when relying on build-time data (which can be tested by replacing this return statement with `return Promise.reject()` - but now the page infinitely loads after a product is selected. Happy to work on this too!

mayaraman19: nvm my fault lol

mayaraman19:
looks good to me after my changes are merged - might be worth adding the search object to the `mockedReposBranches` object in `DeprecatedVersionSelector.test.js`?

mayaraman19 (approved):
Looks good to me!

anabellabuckvar:
> looks good to me after my changes are merged - might be worth adding the search object to the `mockedReposBranches` object in `DeprecatedVersionSelector.test.js`?

@mayaraman19 thanks, this would have made sense! However the ticket AC was changed again so now the dropdown will still be using displayName anyways, which you already included in testing

mmeigs (approved):

PR #981, opened 2024-01-10
DOP-4207: use Gatsby Image to pre process images

seungpark:
### Stories/Links:

DOP-4207

### Context:
- Gatsby Image (by default) uses lazy loading to load images. This was actually detrimental to the site score since lazy loading images above the fold results in a longer LCP -> lower site score.
- Instead of using [Gatsby Image component](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-plugin-image/#gatsbyimage) which uses lazy loading, I have resorted to use all the cool transformations and image processing tools by using the plugins [gatsby-transformer-sharp](https://www.gatsbyjs.com/plugins/gatsby-transformer-sharp/) and `gatsby-source-filesystem` to generate lower image res (ie. webp format), while using the existing `` tag. You can inspect the image for sources and the network tab for the lowered sizes.
- Since we do not use specific page components with pre-determined components, created a Image Context from the scope of the page component, that is consumable by any component within the page. The idea is that the images are transformed and fetched during build time, so that there is no client hydration (other than downloading the image)

### Current Behavior:

https://www.mongodb.com/docs/manual/core/csfle/reference/encryption-components/
https://www.mongodb.com/docs/manual/reference/write-concern/write-lifecycle/
https://www.mongodb.com/docs/manual/core/causal-consistency-read-write-concerns/

### Staging Links:

[staged master branch with user width not being respected](https://docs-mongodbcom-staging.corp.mongodb.com/master/docs/seung.park/DOP-4207-use-regular-img/reference/write-concern/write-lifecycle/) ([from this RST/txt](https://github.com/mongodb/docs/blob/master/source/reference/write-concern/write-lifecycle.txt#L21-L23))

[staged branch on server docs - example of image without preprocessed dimensions](https://docs-mongodbcom-staging.corp.mongodb.com/master/docs/seung.park/DOP-4207-use-regular-img/core/csfle/reference/encryption-components/)

[staged branch on server docs - example of lightbox and figure images](https://docs-mongodbcom-staging.corp.mongodb.com/master/docs/seung.park/DOP-4207-use-regular-img/core/causal-consistency-read-write-concerns/)

[staged master branch on server docs - user width now being respected](https://docs-mongodbcom-staging.corp.mongodb.com/master/docs/seung.park/DOP-4207-use-regular-img/reference/write-concern/write-lifecycle/) from this [RST](https://github.com/mongodb/docs/blob/master/source/reference/write-concern/write-lifecycle.txt#L21-L23)

### Notes:
- there was a minor existing bug of not respecting user option for width. this has been address in the example
- Gatsby Image only processes certain file extensions. (png, jpg, avif, and webp). SVG's are not included since they are minimal in size


src/components/Image.js
```diff
@@ -1,75 +1,79 @@
...
+    imgSrc = withPrefix(imgSrc);
+  }
+
+  const userOptionStyle = {};
+  if (widthOption?.endsWith('px')) {
+    userOptionStyle['width'] = widthOption;
+  }
 
-  const { options: { class: directiveClass } = {} } = nodeData;
+  if (loading === 'lazy') {
```
seungpark, line 53: this will have to be an additional layer on the parser side. we can make up rule(s) for when to use lazy, ie. `when inside a tab`, `when it is in a section n > 2`, `when it is the 3rd image in section n <= 2`, etc.

for now, all images will use the regular `` tag, but with the benefit of being able to use gatsby preprocessed images

src/components/Image.js
```diff
@@ -1,75 +1,79 @@
...
-const Image = ({ nodeData, handleImageLoaded, className }) => {
-  const [height, setHeight] = useState(null);
-  const [width, setWidth] = useState(null);
-  const imgRef = useRef();
+const Image = ({ nodeData, className }) => {
+  const scale = (parseInt(getNestedValue(['options', 'scale'], nodeData), 10) || 100) / 100;
+  const widthOption = getNestedValue(['options', 'width'], nodeData);
+  let height = getNestedValue(['options', 'height'], nodeData);
+  const loading = getNestedValue(['options', 'loading'], nodeData);
+  const directiveClass = getNestedValue(['options', 'directiveClass'], nodeData);
```
mmeigs, line 15: So much prettier 😍

src/components/Figure/Lightbox.js
```diff
@@ -51,10 +51,15 @@ const LightboxWrapper = styled('div')`
 const Lightbox = ({ nodeData, ...rest }) => {
   const [open, setOpen] = useState(false);
   const figureWidth = nodeData.options?.figwidth || 'auto';
+  const openModal = useCallback(() => {
+    setOpen(!open);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
mmeigs, line 57: I believe this should have `open` as a dependency to ensure it's always the present value

mmeigs (approved):
Looks great. And I know it's definitely not a 1:1 comparison, but running Lighthouse on your staging pages vs current prod is stunning. 🔥

src/components/Figure/Lightbox.js
seungpark: thanks for the catch! 🙏

src/components/Image.js
```diff
@@ -1,75 +1,79 @@
...
+  const userOptionStyle = {};
+  if (widthOption?.endsWith('px')) {
+    userOptionStyle['width'] = widthOption;
+  }
 
-  const { options: { class: directiveClass } = {} } = nodeData;
+  if (loading === 'lazy') {
+    // loading option comes from parser. if image is to be lazy loaded, use gatsby image
+    return (
+      <GatsbyImage
```
rayangler, line 56: Is using `GatsbyImage` functionally different from just attaching `loading="lazy"` to the `img` element?

gatsby-config.js
```diff
@@ -8,8 +8,20 @@ const pathPrefix = !isPreview ? generatePathPrefix(siteMetadata) : undefined;
...
-const plugins = ['gatsby-plugin-emotion', isPreview ? 'gatsby-source-snooty-preview' : 'gatsby-source-snooty-prod'];
-
+const plugins = [
+  `gatsby-plugin-image`,
+  `gatsby-plugin-sharp`,
+  {
+    resolve: 'gatsby-source-filesystem',
+    options: {
+      name: 'images',
+      path: `${__dirname}/src/images`, // cannot use public as plugins initialize before gatsby-node module
```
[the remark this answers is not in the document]
seungpark: good catch! adding a separate .gitignore file to /images directory instead

plugins/gatsby-source-snooty-prod/gatsby-node.js
```diff
@@ -141,6 +141,24 @@ exports.sourceNodes = async ({ actions, createContentDigest, createNodeId }) =>
...
+        id: createNodeId(`page-images-${slug}`),
+        internal: {
+          type: 'PageImage',
+          contentDigest: createContentDigest(gatsbyImages.map((asset) => asset.key)),
+        },
+        pageAssets: gatsbyImages.map((asset) => asset.key.slice(1)),
+        parent: null,
+        slug: slug,
+      });
+    }
```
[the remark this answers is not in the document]
seungpark: good catch. this should be also happening in the preview plugin as well.

otherwise, we could skip this image optimization and complementary logic. (i suspect it will increase initial build times when processing assets, ie. including the plugin and asset directory logic only on `env.GATSBY_IS_PREVIEW` ) and letting these use regular images

WDYT?

src/components/Image.js
seungpark: there are some benefits, such as transitions as seen on this page (under Compass tab)

https://docs-mongodbcom-staging.corp.mongodb.com/master/docs/seung.park/DOP-4207-always-lazy/tutorial/analyze-query-plan/

this was a test stage site where i returned each component as a lazy gatsby image

plugins/gatsby-source-snooty-prod/gatsby-node.js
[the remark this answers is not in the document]
seungpark: discussed offline. since the transformations and plugins may add time to gatsby cloud preview builds, we will be excluding this image sourcing + transformation logic for preview builds on gatsby cloud, and following up with a SPIKE ticket to either determine this to be detrimental, or to include it with writers' builds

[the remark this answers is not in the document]
seungpark: https://docs-mongodbcom-staging.corp.mongodb.com/1496c56ab0b343fc39bac1406d586e0feb1a442d/master/docs/seung.park/DOP-4207-use-regular-img/docs/DOCSP-28619/

sample staged site of using non pre-processed images. the images are stored in the regular `/public` directory, and they are accessible through the node's original value ie. `images/image-name.png` via the public directory. *disclaimer:* this sample site was intended to mimic gatsby cloud, but using mut to publish to S3

rayangler (approved):
LGTM overall! Thanks for making sure that this doesn't break the preview plugin

tests/unit/__snapshots__/Figure.test.js.snap
```diff
@@ -85,8 +91,10 @@ exports[`renders lightbox correctly when specified as an option 1`] = `
     >
       <img
         alt="/images/firstcluster.png"
-        class="   emotion-2"
+        class="emotion-2"
+        height="NaN"
         src="/images/firstcluster.png"
+        width="NaN"
```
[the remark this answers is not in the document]
seungpark: added a height and width to one test case and updated the snapshots! still will work with NaN's

PR #982, opened 2024-01-10
DOP-4152: Update version drop-down and "view documentation" buttons for new data context 

anabellabuckvar:
### Stories/Links:

[DOP-4152](https://jira.mongodb.org/browse/DOP-4152)

### Current Behavior:

[Current legacy dropdown](https://www.mongodb.com/docs/legacy/)

### Staging Links:
[staging link](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/anabella.buckvar/DOP-4152/legacy/index.html)

### Notes:
- can be merged since [DOP-4150](https://jira.mongodb.org/browse/DOP-4150) is now completed
- some of the links don't work right now because the url slug is not updated correctly in pool_test.repos_branches


src/components/DeprecatedVersionSelector.js
```diff
@@ -81,30 +58,32 @@ const DeprecatedVersionSelector = ({ metadata: { deprecated_versions: deprecated
...
     if (isBrowser) {
       // Extract the value of 'site' query string from the page url to pre-select product
       const { site } = queryString.parse(window.location.search);
-      if (site && Object.keys(deprecatedVersions).includes(site)) {
+      if (site && Object.keys(reposMap).includes(site)) {
         setProduct(site);
       }
     }
-  }, [deprecatedVersions]);
+  }, [reposMap]);
```
anabellabuckvar, line 86: Does the legacy dropdown selector page ever get accessed from a specific product? It seems to me that this code is no longer necessary but would like confirmation before deleting it.

src/components/DeprecatedVersionSelector.js
```diff
@@ -115,31 +94,41 @@ const DeprecatedVersionSelector = ({ metadata: { deprecated_versions: deprecated
...
-        value: version,
-      }))
+  const versionChoices = reposMap[product]?.branches
+    ? reposMap[product]?.branches
+        .map((version) => {
+          //only include versions with an eol_type field
+          if (!!version.eol_type) {
+            return {
+              text: prefixVersion(version.versionSelectorLabel),
+              value: version,
```
anabellabuckvar, line 123: Each version value is set to an object, which is causing errors as shown in the screenshots below. Functionality is not affected but I'm wondering if I should change it and if so to what
Screenshot 2024-01-12 at 2 48 12 PM
Screenshot 2024-01-12 at 2 48 29 PM

mmeigs: Yes, `option` elements should only have `string` values. This also lines up with the `useState` for `version` that is initialized as a `string`. I might have missed something, but I think the only time you use the stored `version` object is simply to access its `urlSlug`. If so, then I would simply have the value be the `urlSlug`! That will get rid of any errors or inconsistencies.

anabellabuckvar: Thanks, it's fixed now! It ended up making more sense to use the `versionSelectorLabel` and not all of the slugs were unique (at least in the staging database), anyways

src/components/DeprecatedVersionSelector.js
```diff
@@ -32,47 +32,24 @@ const prefixVersion = (version) => {
...
-    ...oldGenRepos,
-    ...reposMap,
-  };
-};
-
-const DeprecatedVersionSelector = ({ metadata: { deprecated_versions: deprecatedVersions } }) => {
+const DeprecatedVersionSelector = () => {
   const { reposDatabase } = useSiteMetadata();
-  const reposBranchesBuildData = useAllDocsets();
+  const reposBranchesBuildData = useAllDocsets().filter((project) => !!project.hasEolVersions);
```
schmalliso, line 48: Why are we using the double bang here rather than just `project.hasEolVersions`? (This isn't a socratic method question, I'm legit curious.)

src/components/DeprecatedVersionSelector.js
```diff
@@ -115,33 +94,46 @@ const DeprecatedVersionSelector = ({ metadata: { deprecated_versions: deprecated
...
+              value: version.versionSelectorLabel,
+              urlSlug: version.urlSlug,
+              icon: version.eol_type === 'download' ? <Icon glyph="Download" /> : null,
+            };
+          } else return null;
+        })
+        //Ensure versions set to null are not included and do not break selector
+        .filter((versionChoice) => !!versionChoice)
+        //sort entries alphabetically by text
+        .sort(alphabetize)
```
schmalliso, line 132: Will this alphabetize `v1.0`, `v1.2`, `v1.10` as 
```
1.0
1.10
1.2
```

And if so, do we have a workaround?

schmalliso (commented):
Just a couple of questions! Looks great.

src/components/DeprecatedVersionSelector.js
anabellabuckvar: I included it as an extra precaution in case a project in `reposBranches` doesn't have a `hasEolVersions` field; however, it's not strictly necessary and I could remove use of it if you think that would be better

schmalliso: Nope it's all good, I was just wondering - your rationale is very sound :)

anabellabuckvar: After chatting with @mmeigs, they are truly unnecessary here so I removed double bang instances

src/components/DeprecatedVersionSelector.js
anabellabuckvar: We did not have a workaround. Now we do!

src/components/DeprecatedVersionSelector.js
```diff
@@ -81,32 +58,34 @@ const DeprecatedVersionSelector = ({ metadata: { deprecated_versions: deprecated
...
     }
-  }, [reposDatabase, reposBranchesBuildDataMapWithOldGen]);
+  }, [reposDatabase, reposBranchesBuildDataMap]);
 
   useEffect(() => {
     if (isBrowser) {
       // Extract the value of 'site' query string from the page url to pre-select product
       const { site } = queryString.parse(window.location.search);
-      if (site && Object.keys(deprecatedVersions).includes(site)) {
+      if (site && Object.keys(reposMap).includes(site)) {
```
seungpark, line 82: minor nitpick. instead of `Object.keys(reposMap).includes(site)`, can we use `reposMap[site]` ?

src/components/DeprecatedVersionSelector.js
```diff
@@ -81,32 +58,34 @@ const DeprecatedVersionSelector = ({ metadata: { deprecated_versions: deprecated
...
+            'project'
+          );
+          if (reposBranchesMap.size > 0) setReposMap(reposBranchesMap);
         })
         .catch((error) => {
           console.error(`ERROR: could not access ${reposDatabase} for dropdown data.`);
         });
     }
-  }, [reposDatabase, reposBranchesBuildDataMapWithOldGen]);
+  }, [reposDatabase, reposBranchesBuildDataMap]);
```
mmeigs, line 76: I don't think we need the `reposBranchesBuildDataMap` as a dependency!

src/components/DeprecatedVersionSelector.js
```diff
@@ -128,8 +128,13 @@ const DeprecatedVersionSelector = () => {
         })
         //Ensure versions set to null are not included and do not break selector
         .filter((versionChoice) => !!versionChoice)
-        //sort entries alphabetically by text
-        .sort(alphabetize)
+        //sort versions newest(larger numbers) to oldest(smaller numbers). Assumes there are no more than three digits between/before/after each decimal place
+        .sort((a, b) =>
```
seungpark, line 132: nice sorting! good chance to pull into util and add some unit tests (not blocking!)

src/components/DeprecatedVersionSelector.js
```diff
@@ -115,33 +94,51 @@ const DeprecatedVersionSelector = ({ metadata: { deprecated_versions: deprecated
...
+          a.text
+            .toString()
+            .replace(/\d+/g, (n) => +n + 1000)
+            .localeCompare(b.text.toString().replace(/\d+/g, (n) => +n + 1000))
+        )
     : [];
 
+  const versionChoicesMap = keyBy(versionChoices, 'value');
+
```
mmeigs, line 141: nit: We could wrap both of these variable definitions in a `useMemo` to keep them updating only when necessary

src/components/DeprecatedVersionSelector.js
```diff
@@ -115,33 +94,51 @@ const DeprecatedVersionSelector = ({ metadata: { deprecated_versions: deprecated
 
     // Utilizing hardcoded env because legacy sites are not available on dev/stage
     const hostName = reposMap[product].url.dotcomprd + reposMap[product].prefix.dotcomprd;
-    const versionOptions = deprecatedVersions[product];
-    const versionName = isVersioned(versionOptions) ? version : '';
-    return `${hostName}/${versionName}`;
+    return `${hostName}/${currentVersion?.urlSlug}`;
   };
```
mmeigs, line 98: Because there's a possibility of currentVersion being undefined, maybe we should check for that at the start of this generateUrl function and return `null` if that's the case to make sure that no odd url like `<hostname>/undefined` occurs

src/components/DeprecatedVersionSelector.js
```diff
@@ -115,33 +94,51 @@ const DeprecatedVersionSelector = ({ metadata: { deprecated_versions: deprecated
...
+        })
+        //Ensure versions set to null are not included and do not break selector
+        .filter((versionChoice) => versionChoice)
+        //sort versions newest(larger numbers) to oldest(smaller numbers). Assumes there are no more than three digits between/before/after each decimal place
+        .sort((a, b) =>
+          a.text
+            .toString()
+            .replace(/\d+/g, (n) => +n + 1000)
+            .localeCompare(b.text.toString().replace(/\d+/g, (n) => +n + 1000))
+        )
```
mmeigs, line 137: Yay! Thanks for this

mmeigs (commented):
Just a couple more comments! This is such an exciting change.

src/components/DeprecatedVersionSelector.js
```diff
@@ -160,16 +157,20 @@ const DeprecatedVersionSelector = ({ metadata: { deprecated_versions: deprecated
...
-        View Documentation
+      ></Select>
+      <Button
+        variant="primary"
+        title="View or Download Documentation"
+        rightGlyph={versionChoicesMap[version]?.icon}
+        href={generateUrl(versionChoicesMap[version])}
+        disabled={buttonDisabled}
+      >
+        {version?.eol_type === 'download' ? 'Download Documentation' : 'View Documentation'}
```
seungpark, line 168: think we should be checking `versionChoicesMap[version]?.eol_type` here. seems like `version` is a string CMIIW

tests/unit/DeprecatedVersionSelector.test.js
```diff
@@ -169,13 +170,12 @@ describe('DeprecatedVersionSelector when rendered', () => {
...
+      wrapper = render(<DeprecatedVersionSelector />);
 
       const productDropdown = wrapper.container.querySelectorAll('button')[0];
       userEvent.click(productDropdown);
 
       expect(wrapper.findByText('MongoDB Manual')).toBeTruthy();
-      expect(wrapper.findByText('MongoDB Ops Manager')).toBeTruthy();
       expect(wrapper.findByText('MongoDB Atlas Open Service Broker on Kubernetes')).toBeTruthy();
     });
   });
```
seungpark, line 181: minor. possibly another test for EOL selectors

seungpark (changes requested):
this looks great! a few minor changes below.

also, thanks for the comments! the links are not working but will assume pool_test.repos_branches entries are not cleaned. as for the third comment, not seeing any error on select components

src/components/DeprecatedVersionSelector.js
anabellabuckvar: You're right! good catch, thank you

src/components/DeprecatedVersionSelector.js
anabellabuckvar: Added a test, lmk what you think!

seungpark (approved):
this LGTM. threw me off a little bit because it differs from prod data (by virtue of different sources of truth). 

with that said, would validate that pool_test and pool repos_branches entries are 1) in sync and 2) ready to be tested in preprd so we can promote to prod :)

PR #986, opened 2024-01-19
DOP-4207: check if image exists before calling

seungpark:
### Stories/Links:

DOP-4207

### Notes:
- Minor fix to check for null images. Was getting a build error `  WebpackError: TypeError: Cannot read properties of null (reading 'relativePath')` on cloud-docs /billing page.
- Verified this works as intended with webp images as in this [staging link ](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/mmeigs-move-ako/cloud-docs/seung.park/DOP-4207-safeguard-null-images/billing/index.html)

### DIFF
--- a/src/context/image-context.js
+++ b/src/context/image-context.js
@@ -15,7 +15,9 @@ export default ImageContext;
 const ImageContextProvider = ({ images, children }) => {
   const imageByPath = {};
   for (const image of images) {
-    imageByPath[image.relativePath] = getImage(image);
+    if (image?.relativePath) {
+      imageByPath[image.relativePath] = getImage(image);
+    }
   }
   return <ImageContext.Provider value={{ imageByPath }}>{children} </ImageContext.Provider>;
 };


mmeigs (commented):
Just wondering what's the case of this being null and do we want to simply ignore it?

branberry:
I had the same question as @mmeigs, it looks like we could run into an issue [here](https://github.com/mongodb/snooty/blob/60e49b501df7b6d20c0100a393bcf1864cd51473/src/components/Image.js#L33) with a missing source property, I think...

seungpark:
> Just wondering what's the case of this being null and do we want to simply ignore it?

this has to do with associated graphql nodes [as seen here ](https://github.com/mongodb/snooty/blob/master/plugins/gatsby-source-snooty-prod/gatsby-node.js#L325-L328) and the nodes are created [here](https://github.com/mongodb/snooty/blob/master/plugins/gatsby-source-snooty-prod/gatsby-node.js#L150) (by our code) and [here](https://github.com/mongodb/snooty/blob/master/gatsby-config.js#L18) where it creates nodes on files in src/images directory.

checking graphql, seems like some null file nodes are being tied with the same relativePath:
image

digging into why we are getting nulls in this, but seems like the images that are processed are being sent through

mmeigs:
> > I had the same question as @mmeigs, it looks like we could run into an issue [here](https://github.com/mongodb/snooty/blob/60e49b501df7b6d20c0100a393bcf1864cd51473/src/components/Image.js#L33) with a missing source property, I think...
> 
> 
> 
> this has a safe fallback to go back to the original image as the original prop [in these lines](https://github.com/mongodb/snooty/blob/60e49b501df7b6d20c0100a393bcf1864cd51473/src/components/Image.js#L37-L46)
> 
> 
> 
> Seems like some images are not being processed, specifically ones that are not included in the page, but included in static_assets. Regardless, even if image is not processed, should safely fall back as outlined above

Great. This is what I was mainly wondering about! Thanks!

mmeigs (approved):

PR #988, opened 2024-01-22
DOP-4207: remove leading slash with regex

seungpark:
### Stories/Links:

DOP-4207

The problem for this issue was that we replaced the first char of image argument (the path), without checking if it was leading slash or not. The fix is a simple regex that replaces all leading slashes, just as [Path.normalize](https://nodejs.org/api/path.html#pathnormalizepath) would.

Example of doc content [with](https://github.com/10gen/cloud-docs/blob/master/source/billing.txt#L160) and [without](https://github.com/10gen/cloud-docs/blob/master/source/billing.txt#L279) leading slashes. 

### Current Behavior:

[Previous PR](https://github.com/mongodb/snooty/pull/986) mitigated this error, but would still result in no image being shown (as in [this staging link](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/mmeigs-move-ako/cloud-docs/seung.park/DOP-4207-safeguard-null-images/billing/index.html#usage-by-month-chart))

### Staging Links:

[Removing the leading slash with regex provides good pathing](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/mmeigs-move-ako/cloud-docs/seung.park/DOP-4207-fix-leading-slash/billing/index.html#usage-by-month-chart)


src/utils/replace-leading-slash.js
```diff
@@ -0,0 +1,3 @@
+const replaceLeadingSlash = (filePath) => filePath.replace(/^\/+/, '');
```
mmeigs, line 1: This is so silly, but waffling on whether `removeLeadingSlash` or `replaceLeadingSlash` is better. Up to you. Just putting it out there.

mmeigs (approved):

src/utils/replace-leading-slash.js
rayangler: +1 on remove instead of replace, but not a deal breaker

rayangler (approved):

src/utils/replace-leading-slash.js
seungpark: great catch! done

PR #990, opened 2024-01-24
DOP-3949: Rename the "master" branch to "main"

i80and:
### Stories/Links:

DOP-3949

### Notes:

* Updated documentation to refer to the `main` branch
* Rename `ensure-master.js` to `ensure-main.js`
* Make `ensure-main.js` tolerant of both `master` and `main`. After the branch is officially renamed, this should be changed to just `main`. 

I'm not aware of other repos referring to the branch name. The autobuilder as far as I can see is only tied to specific tags, not branch names.


rayangler (approved):

casthewiz (changes requested):
We should update references to master within https://github.com/mongodb/snooty/blob/master/.github/workflows/test.yml 👀

casthewiz (approved):

PR #997, opened 2024-01-30
DOP-4283: Add buttons for lab drawer

rayangler:
### Stories/Links:

DOP-4283

### Current Behavior:

[Server prod](https://www.mongodb.com/docs/manual/tutorial/getting-started/)

### Staging Links:

[Server staging with feature flag ON](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/raymund.rodriguez/DOP-4283/tutorial/getting-started/index.html) - with drawer
[Server staging with feature flag ON](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/raymund.rodriguez/DOP-4283/reference/method/db.collection.find/index.html#try-it-yourself) - no drawer

The following should load similar to the current behavior:
[Server staging with feature flag OFF](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/drawer-disabled/master/docs/raymund.rodriguez/DOP-4283/tutorial/getting-started/index.html)
[Server staging with feature flag OFF](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/drawer-disabled/master/docs/raymund.rodriguez/DOP-4283/reference/method/db.collection.find/index.html#try-it-yourself)

### Notes:

* Main focus of this PR is to add the buttons needed for the lab drawer, update leftover styling for the drawer, and update tests.
* There were a few changes in plans in both the code and original A/C (see my comments below). This shouldn't affect the implementation too much.
* To test locally, ensure you have the `GATSBY_FEATURE_LAB_DRAWER` frontend feature flag set to true, and pull in the latest changes of the parser to parse a page with an instruqt directive with the drawer option. For example:

```
.. instruqt:: /mongodb-docs/tracks/example?token=example
   :title: Example Lab
   :drawer:
```


src/components/DocumentBody.js
```diff
@@ -141,13 +142,15 @@ const DocumentBody = (props) => {
...
         template={template}
       >
-        <FootnoteContext.Provider value={{ footnotes }}>
-          <Template {...props} useChatbot={useChatbot}>
-            {pageNodes.map((child, index) => (
-              <ComponentFactory key={index} metadata={metadata} nodeData={child} page={page} slug={slug} />
-            ))}
-          </Template>
-        </FootnoteContext.Provider>
+        <InstruqtProvider hasLabDrawer={page?.options?.instruqt}>
```
rayangler, line 145: I had to move this away from `RootProvider` because the `isOpen` state was being preserved when switching between pages. Moving it to `DocumentBody` ensures that values are reset between pages. 

An alternative solution was to use a `useEffect` to set the state back to default whenever the page `slug` prop changed value, but it seemed like unnecessary logic at the root level when the context could have just been brought down to the page level. Please let me know if there's a preference though!

src/components/Instruqt/DrawerButtons.js
```diff
@@ -0,0 +1,50 @@
...
+  };
+
+  const handleDrawerClose = () => {
+    setIsOpen(false);
+  };
+
+  return (
+    <div className={cx(buttonContainerStyle)}>
+      <Icon className={cx(iconStyle)} width={16} height={16} onClick={handleDrawerSize} glyph={drawerSizeGlyph} />
+      <Icon className={cx(iconStyle)} width={24} height={24} onClick={handleMaximizeDrawer} glyph="FullScreenEnter" />
```
rayangler, line 44: Originally, this was supposed to open the Instruqt lab in a new tab, but it turns out I made a mistake during the initial investigation of how lab progress was saved. Tl;dr is that lab progress does not seem to be saved between tabs when the labs are started on different origins (MongoDB vs. Instruqt). Instead, we'll go with the alternative of setting the height of the drawer to its maximum.

I'll go ahead and add a comment to the ticket to reflect these recent findings accordingly

src/components/Instruqt/DrawerButtons.js
```diff
@@ -0,0 +1,50 @@
...
+const DrawerButtons = ({ isMinHeight, targetHeight, setHeight, maxHeight }) => {
+  const { setIsOpen } = useContext(InstruqtContext);
+  const drawerSizeGlyph = isMinHeight ? 'ArrowUp' : 'Minus';
+
+  const handleDrawerSize = () => {
+    setHeight(targetHeight);
+  };
+
+  const handleMaximizeDrawer = () => {
+    setHeight(maxHeight);
```
seungpark, line 34: out of curiosity, should this button act differently when already set to maxHeight?

rayangler: Thanks for flagging. I asked Lara about it and we'll set to the default drawer size when at max height. I've updated the staging link accordingly. 

Designs haven't been updated due to the recentness of the switch from the original plan to have it open to a new tab, to having it maximize instead (but behavior is mentioned in an [annotation](https://www.figma.com/file/7cR6eWtFvKTnJIPL8ZsTaB/Content-Template-Projects?type=design&node-id=708-13171&mode=design&t=1XGETtSk9pv9vuCY-0))

seungpark: sounds good to me! thanks for flagging and updating!

src/components/DocumentBody.js
anabellabuckvar: This makes sense to me!

PR #1000, opened 2024-02-01
DOP-4247: Build Artifacts

mmeigs:
### Stories/Links:

DOP-4247

### Links to Builds:

[Cloud-docs](https://github.com/mongodb/snooty/actions/runs/7744426944)
[Server Docs](https://github.com/mongodb/snooty/actions/runs/7744426952)

^ note above you can download the build artifacts

### Notes:

This PR utilizes the new ([PR](https://github.com/mongodb/docs-worker-actions/pull/77)) Github Action to download latest build data of the desired project, build the docs site, and save the build bundle to an Artifact.

I've set up one workflow for each desired project.

Because of using data from the Snooty Data API, even with mimicking the structure of our manifest files, there were minimal changes necessary to ingest this within Snooty.


.github/workflows/lighthouse-cloud-docs.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Access Build Data
+        uses: mongodb/docs-worker-actions/build-artifact@DOP-4247-build-artifact
+        env: 
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PROJECT_TO_BUILD: cloud-docs
+      - name: Build Snooty
+        env:
+          NPM_BASE_64_AUTH: ${{ secrets.NPM_BASE_64_AUTH }}
+          NPM_EMAIL: ${{ secrets.NPM_EMAIL }}
+          GATSBY_MANIFEST_PATH: snooty-metadata.json
```
rayangler, line 23: I think we might want to include `GATSBY_PARSER_USER: docsworker-xlarge` to match the Autobuilder's setup, and to properly strip away the user from the final page path in [this line](https://github.com/mongodb/snooty/blob/d7ce37c858168e51ed352a10a05ac2d06af09f06/plugins/gatsby-source-snooty-prod/gatsby-node.js#L129)

rayangler (commented):
This is super cool!

src/init/DocumentDatabase.js
```diff
@@ -65,11 +67,22 @@ class ManifestDocumentDatabase {
...
     const result = [];
-    const zipEntries = this.zip.getEntries();
-    for (const entry of zipEntries) {
-      if (entry.entryName.startsWith('documents/')) {
-        const doc = BSON.deserialize(entry.getData());
-        result.push(doc);
+    if (!this.zip) {
+      // Read documents from Gatsby Action download
+      try {
+        const documents = JSON.parse(fs.readFileSync('snooty-documents.json'));
```
rayangler, line 73: (Just curious) Is there a reason we use a separate GH action to fetch and parse the data, instead of fetching from the API directly and keeping them loaded in memory here?

mmeigs: Interesting! It seemed like a central place to do it and to separate the parsing logic. 
I could see this working, too! The only drawback this way would be I believe we would almost definitely have to fetch and parse through all the lines of types twice because of the separation in state between [metadata](https://github.com/mongodb/snooty/blob/d7ce37c858168e51ed352a10a05ac2d06af09f06/src/utils/setup/fetch-manifest-metadata.js#L7) and [assets/documents](https://github.com/mongodb/snooty/blob/d7ce37c858168e51ed352a10a05ac2d06af09f06/src/init/DocumentDatabase.js#L58).

rayangler: I think to avoid double parsing, we could probably add an extra async function to this class (or have some other abstraction) to handle fetching from the API and saving them to respective `this.documents`, `this.metadata`, etc. fields, so that by the time `getDocuments()` (or whatever function) is called, it just returns that data. Something like:

```js
const db = new DocumentDatabase();
await db.init(); // this is where the API is called bc I don't think constructors can be async, OR we could use the existing connect() function? 🤔 
await db.getDocuments(); // this just returns the page documents that were already parsed
await db.getAssets(); // this just returns the metadata that was already parsed
```

The main benefit of allowing the frontend to consume the API data directly is convenience in terms of maintainability and testing. It would also prevent us from needing to transform data from API -> json files -> memory, and instead go from API -> memory. 

That being said, I don't think having the separate GH action is necessarily a blocker for this PR, especially because it already works and I don't know what the complexity or exact limitations are of porting that logic over. I'm okay with whatever people think is best!

branberry, line 73: Not sure how big of an impact this might be, but I think it would preferable to use an `async` version of `fs.readFileSync('snooty-documents.json')`. The reason being that the `sync` methods in node block the main thread, and any concurrent processes have to wait until that sync function completes. [Here is a good article on this topic](https://nodejs.org/en/learn/asynchronous-work/overview-of-blocking-vs-non-blocking).

I believe the Gatsby build process relies on async processing of nodes to do stuff concurrently, so this could interfere with that.

As an alternative, we could do this:

```ts

import fs from 'fs';

const readFileAsync = promisify(fs.readFile);

        const documents = JSON.parse(readFileAsync('snooty-documents.json'));


```

seungpark (commented):
the output looks great! a couple comments below for usage of artifacts between `Access Build Data` to `Build Snooty` steps

also, guessing this is from a previous commit, but this action shows artifact with name `lighthouse-cloud-docs.zip` - where is the `lighthouse` coming from?

src/init/DocumentDatabase.js
mmeigs: Right! I agree with everything you've said completely. It's more the separation between [metadata](https://github.com/mongodb/snooty/blob/d7ce37c858168e51ed352a10a05ac2d06af09f06/src/utils/setup/fetch-manifest-metadata.js#L7) and [assets/documents](https://github.com/mongodb/snooty/blob/d7ce37c858168e51ed352a10a05ac2d06af09f06/src/init/DocumentDatabase.js#L58).

rayangler: Ah okay. Thanks for shedding some more light on the context of this!

mmeigs: Very, very helpful. Thank you for catching this!

.github/workflows/stage-docs-site.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Access Build Data
+        uses: mongodb/docs-worker-actions/build-artifact@DOP-4247-build-artifact
+        env: 
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PROJECT_TO_BUILD: ${{ matrix.project }}
+      - name: Build Snooty
+        env:
+          NPM_BASE_64_AUTH: ${{ secrets.NPM_BASE_64_AUTH }}
+          NPM_EMAIL: ${{ secrets.NPM_EMAIL }}
+          GATSBY_MANIFEST_PATH: snooty-metadata.json
```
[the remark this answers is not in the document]
mmeigs: It's not a zip! This workflow does not build a zipfile. It uses the Snooty Data API to grab the most recent build data, but that is not in a zipped format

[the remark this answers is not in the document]
seungpark: right, but can we make a zip artifact of those streamed data from the data API? ie. [this states artifacts are downloaded as zips from one job to another](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#about-workflow-artifacts), and even the artifact uploaded as the HTML output is a zip file.

if not, then we differentiate between zipped bundles vs json (with different env variables), instead of having the manifest path be `.zip` or `.json` ? WDYT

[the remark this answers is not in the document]
mmeigs: Possible!

mmeigs:
Refactored with all feedback taken into account. Also, for cleanliness, depending on a new env var `GATSBY_BUILD_FROM_JSON` instead of parsing to know if it is a zip file path.

.github/workflows/stage-docs-site.yml
```diff
@@ -0,0 +1,40 @@
...
+    strategy:
+      matrix: 
+        project: ['cloud-docs', 'docs']
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '18.x'
+      - name: Access Build Data
+        uses: mongodb/docs-worker-actions/build-artifact@DOP-4247-build-artifact
```
rayangler, line 21: Will we need to change the branch name?

.github/workflows/stage-docs-site.yml
```diff
@@ -0,0 +1,40 @@
...
+jobs: 
+  staging:
+    permissions: write-all
+    runs-on: ubuntu-latest
+    strategy:
+      matrix: 
+        project: ['cloud-docs', 'docs']
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
```
rayangler, line 17: (Nit) I just saw some warnings in the GH action summary about these having older versions of Node. We can probably bump these up to whatever works

rayangler (approved):
LGTM!

.github/workflows/stage-docs-site.yml
mmeigs: Nice!

PR #1005, opened 2024-02-07
:white_check_mark: DOP-4302 handles test with lazy loading a component

caesarbell:
### Stories/Links:

DOP-4302

### Current Behavior:

[Flaky test](https://github.com/mongodb/snooty/actions/runs/7805040018/job/21288329645)

### Staging Links:

[Passing test](https://github.com/mongodb/snooty/actions/runs/7820424159)

### Notes:

It seems according to Jest's [test](https://jestjs.io/docs/api#testname-fn-timeout) method documentation, there is a way of making a test wait for the promise to resolve before letting the test complete. 

In our case, the solution is to return the promise from the `waitFor` async function from RTL.  Since this test now waits for the promise to resolve, we can remove the optional timeout.

### DIFF
--- a/tests/unit/Presentation.test.js
+++ b/tests/unit/Presentation.test.js
@@ -45,25 +45,19 @@ describe('DocumentBody', () => {
       expect(languageSelector.querySelectorAll('li')).toHaveLength(4);
     }
 
-    await waitFor(
-      () => {
-        const feedbackWidget = screen.getByText(FEEDBACK_BUTTON_TEXT);
-        expect(feedbackWidget).toBeVisible();
-        expect(feedbackWidget).toMatchSnapshot();
+    const feedbackWidget = await screen.getByText(FEEDBACK_BUTTON_TEXT);
+    expect(feedbackWidget).toBeVisible();
+    expect(feedbackWidget).toMatchSnapshot();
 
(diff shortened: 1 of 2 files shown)


caesarbell:
I re-ran the jobs about three times, each one passing, assuming that would be equivalent to pushing up multiple commits to trigger the builds. Each time, the test passes.

tests/unit/Presentation.test.js
```diff
@@ -58,12 +62,7 @@ describe('DocumentBody', () => {
       },
       { timeout: 8000 }
```
mmeigs, line 63: TIL about returning waitFor would be the magic bullet. But wondering if that means we still need the timeout or no?

mmeigs: ^ btw, I'm totally fine with it!! Just wondering if it's necessary.

caesarbell: @mmeigs good point. I know we don't need the `test` method's option timeout param, however, the timeout option is still needed for the `waitFor` method. `waitFor` runs the callback a number of times until the timeout is reached, and the default value is 1000ms. In our use case we need `8000ms`, but we won't need to set the timeout in the test method.

mmeigs (approved):

PR #1007, opened 2024-02-09
DOP-4289: Add readme update check to PR template

mmeigs:
### Stories/Links:

DOP-4289

### Notes:

Screenshot 2024-02-09 at 12 17 35 PM

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -11,3 +11,10 @@ _Put a link to current staging or production behavior, if applicable_
 _Put a link to your staging environment(s), if applicable_
 
 ### Notes:
+
+
+
+
+### README updates
+
+[] This PR introduces changes that should be reflected in the README, and I have made those updates.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -17,4 +17,6 @@ _Put a link to your staging environment(s), if applicable_
 
 ### README updates
 
-[] This PR introduces changes that should be reflected in the README, and I have made those updates.
+ - [ ] This PR introduces changes that should be reflected in the README, and I have made those updates.
+ - [ ] This PR does not introduce changes that should be reflected in the README </br>
+<sub>(check using an 'x')</sub>


.github/pull_request_template.md
```diff
@@ -11,3 +11,12 @@ _Put a link to current staging or production behavior, if applicable_
...
 ### Notes:
+
+
+
+
+### README updates
+
+ - [ ] This PR introduces changes that should be reflected in the README, and I have made those updates.
+ - [ ] This PR does not introduce changes that should be reflected in the README </br>
+<sub>(check using an 'x')</sub>
```
rayangler, line 22: Fwiw, clicking on the checkbox should work just fine as well

branberry (approved):
Looks good!

.github/pull_request_template.md
```diff
@@ -11,3 +11,12 @@ _Put a link to current staging or production behavior, if applicable_
...
 
 ### Notes:
+
+
+
+
+### README updates
+
+ - [ ] This PR introduces changes that should be reflected in the README, and I have made those updates.
+ - [ ] This PR does not introduce changes that should be reflected in the README </br>
```
rayangler, line 21: (Super duper nit; definitely non-blocking) I'm not strictly opposed to having 2 checkboxes, but I feel like having a single checkbox to ensure the PR author has double-checked if this needs a README update would be the most streamlined, since it acts as a sort of checklist. Having 2 checkboxes makes me feel like I have to read both options and think more about selecting the correct one like a questionnaire (although maybe that's actually a good thing). Wdyt?

mmeigs: I mean, I decided to do two for that reason. I feel like if it's just the first one, then as a reviewer I'll wonder if the PR author just forgot to check? I was thinking it was most obvious to force the author to check something.

.github/pull_request_template.md
mmeigs: oh gorgeous. I'll take out the hint then!

.github/pull_request_template.md
rayangler: Makes sense to me! Thanks!

PR #1008, opened 2024-02-09
DOP-4355: Extend grab area of lab drawer handle

rayangler:
### Stories/Links:

DOP-4355

### Current Behavior:

[Server staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/raymund.rodriguez/main/tutorial/getting-started/index.html) - from the bug bash

### Staging Links:

[Server staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/raymund.rodriguez/DOP-4355/tutorial/getting-started/index.html)

### Notes:

* Main change was to wrap the lab handle to allow the grab area to extend past the actual handle div. Please let me know if the range is big enough or not!
* This PR also fixes the annoying interaction in Chrome with the iframe where resizing downwards would become wonky after the mouse starts to hover over the iframe. The change was to add a div overlay only when resizing so that the mouse would technically directly hover the div, and not the iframe itself. Unsure if this is the best way to do this or if this fixes it for everyone, so please feel free to try it out!

### DIFF
--- a/src/components/Instruqt/LabDrawer.js
+++ b/src/components/Instruqt/LabDrawer.js
@@ -32,18 +32,25 @@ const handleContainerStyle = css`
   justify-content: center;
 `;
 
-const handleStyle = css`
+const handleGrabArea = css`
   position: absolute;
-  border-radius: 4px;
   cursor: ns-resize;
   top: 10px;
-  background-color: ${palette.white};
-  width: 50px;
-  height: 4px;
+
+  // Allows element to be bigger in size while maintaining visual size
+  padding: 10px 20px 20px 20px;
+  margin: -10px -20px -20px -20px;
 
   @media ${theme.screenSize.upToMedium} {
     display: none;
   }
+`;
+
+const handleStyle = css`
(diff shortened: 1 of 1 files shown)


anabellabuckvar (approved):
LGTM! My only note is that personally, it seems intuitive to have the grab area extend to pretty much the entire width of the top strip of the lab drawer, but that also might be more of a design decision.

rayangler:
> My only note is that personally, it seems intuitive to have the grab area extend to pretty much the entire width of the top strip of the lab drawer, but that also might be more of a design decision.

Good point. Reached out to Lara to see what the recommendation is

branberry (approved):

rayangler:
@anabellabuckvar I reached out to Lara and she responded with the following:

> That looks good to me! I don't think we need to extend it to the whole width

I'll merge in this PR with the current grab area later today. If we need to update the grab area again in the future, we should be able to just tweak the css a bit

PR #1012, opened 2024-02-14
DOP-4362: Show Chatbot FAB by default and remove Optimizely

rayangler:
### Stories/Links:

DOP-4362

### Current Behavior:

[Server prod](https://www.mongodb.com/docs/manual)

### Staging Links:

[Server staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/raymund.rodriguez/DOP-4362/index.html)
[Landing staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/raymund.rodriguez/DOP-4362/index.html)

### Notes:

* To validate changes, go to different docs pages and ensure Chatbot FAB is shown, except for the `docs-landing` homepage. The network tab of the browser's dev tools should not show any requests to the Optimizely CDN. There should still be a single request present for something related to Optimizely, but it seems like it's an integration tied to some unrelated analytics script outside of our direct control.
* **Bonus:** Adds a new eslint rule to prevent checking test snapshots in async testing functions like `waitFor`.

### README updates

- [ ] This PR introduces changes that should be reflected in the README, and I have made those updates.
- [x] This PR does not introduce changes that should be reflected in the README


tests/unit/Presentation.test.js
```diff
@@ -53,12 +54,9 @@ describe('DocumentBody', () => {
       () => {
         const feedbackWidget = screen.getByText(FEEDBACK_BUTTON_TEXT);
         expect(feedbackWidget).toBeVisible();
-        expect(feedbackWidget).toMatchSnapshot();
 
         const chatbotWidget = screen.getByText(CHATBOT_WIDGET_TEXT);
-        /* NOT to be visible for now, with display:none */
-        expect(chatbotWidget).not.toBeVisible();
-        expect(chatbotWidget).toMatchSnapshot();
```
rayangler, line 61: I'm removing checking for snapshots here because:

1) Turns out that `testing-library` has an [eslint rule](https://github.com/testing-library/eslint-plugin-testing-library/blob/5497dc269825c89c7022febed7efed741826b6b2/docs/rules/no-wait-for-snapshot.md) dedicated to making sure we don't check for snapshots in async testing functions. I've added the rule as part of this PR to ensure we catch this. (I accidentally created 16+ new snapshots without realizing due to a change I made 😅)
2) The snapshots for these elements are pretty small and mostly tied to the text they're searching for, so removing these snapshots seems low risk if implementation changes.

Let me know if there are any preferences on the testing change here, or if the eslint rule is unnecessary!

mmeigs (approved):
Very excited to get this out of there. Thanks!

PR #1016, opened 2024-02-20
DOP 4340/1: Homepage Redesign

mayaraman19:
### Stories/Links:

DOP-4340 + DOP-4341 + DOP-4342 + other small tweaks

New homepage! Now including @anabellabuckvar's tweaks to the introduction header. Built off of [this branch](https://github.com/mongodb/docs-landing/pull/157) of `docs-landing`.

### Staging Links:

[Updated staging link](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOP-4300/landing/maya/DOP-4340-1-homepage/index.html)

[Figma](https://www.figma.com/file/SxKPGkhEJ8ixM9giIPEuJC/Docs-Homepage-Redesign?type=design&node-id=2897-8661&mode=design&t=IKmAcS72dkG5VqCI-0)

### Notes:
- After verifying with design + Sarah Lin, "Docs Home" title was removed from the Sidenav, and nothing other than that should change from the production Sidenav

### README updates


src/components/Card/index.js
```diff
@@ -17,6 +17,28 @@ const cardBaseStyles = css`
...
+  img {
+    width: ${theme.size.xlarge};
+    vertical-align: baseline;
+    height: fit-content;
+  }
+  div {
+    display: flex;
+    flex-direction: column;
+    margin-left: ${theme.size.large};
+    //min-width: 164px;
```
mmeigs, line 33: Should this be removed?

src/components/Card/index.js
```diff
@@ -17,6 +17,28 @@ const cardBaseStyles = css`
   height: 100%;
 `;
 
+// new->old homepage redesign
```
mmeigs, line 20: Not totally sure what this comment means or adds to the code

mmeigs (changes requested):
Okay, these PRs are always so nit-picky to review, so sorry. And definitely check with design before you just take my opinion!!

- The left padding in the main four cards seems different in the design. This could also help for the text to fully stretch out in the max-width possibly. But the design seems to have two different padding-left amounts. Two of them are ~25px and two are ~29px. Your implementation is 32px. 
- I'm not sure if this falls into what @anabellabuckvar is working on, but the secondary heading "More ways to learn" has incorrect styling. (I'm guessing this is a different ticket, but thought I'd mention)
- Comment for docs-landing PR - Check with design and also with Sarah Lin? But the titles on the TOC are different from design. This can be changed in the `.. ia:` section in docs-landing `index.txt`
- The line-height value is incorrect in the text of the small cards at the bottom of the page, ie "Take Free Courses on MongoDB University"
- On mobile, in the larger four cards, the icon and text need to switch to column layout. One on top of the other.
- Oh, and one very very odd thing: in this staging link there is a TOC entry at the top that says "Docs Home"... shouldn't be there, and not really sure how it got there?! lol

seungpark (commented):
minor comments below but overall looks great!

one minor thing i noticed is that the hero and text are not aligned very well. we set the height of the hero at different viewports and the `align-self: end` properties on the heading and paragraph make the margins vary when changing viewport.

for example:

image

image

we can take this in another issue, but perhaps we can make the hero image as a background-image property and have strict margins, not alignments, on the heading and paragraph elements.

src/components/Card/CardGroup.js
```diff
@@ -48,19 +51,31 @@ const StyledGrid = styled('div')`
...
 
   @media ${theme.screenSize.upToSmall} {
-    margin: ${({ isForDrivers }) => getMarginStyles(isForDrivers, 24)};
+    margin: ${({ isForDrivers, isLanding, columns }) => getMarginStyles(isForDrivers, isLanding, columns, 24)};
   }
 
-  @media ${theme.screenSize.upToXLarge} {
-    grid-template-columns: repeat(2, 1fr);
+  ${
+    '' /*want first 4 cards on landing page to stay as one column, and not affect group of 3 cards (so hacky i'm sorry) */
```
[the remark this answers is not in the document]
mayaraman19: without this part, when the viewport is a bit smaller the card groups become 2 x 2, but i want them to stay in one column regardless of viewport size

mayaraman19:
I think I have addressed the comments - as for the header, this got approved by design, we didn't want to touch too many things in case the homepage gets changed again but I'm happy to create a separate ticket because I do think having set margins would be nice.

src/components/Card/CardGroup.js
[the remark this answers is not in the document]
seungpark: oh, would we not want to honor the `columns` options here? seems a bit forced to force 2 columns based on viewport. we can possibly set a default to 2 to prevent regressions but can always favor the `column` option if available

src/components/Card/index.js
```diff
@@ -36,6 +66,13 @@ const cardDriverStyle = css`
   }
 `;
 
+const landingBottomStyling = css`
+  p {
+    font-weight: 500;
+    line-height: 24px;
```
seungpark, line 72: (nit) to use theme here

src/components/Card/index.js
```diff
@@ -17,6 +17,36 @@ const cardBaseStyles = css`
   height: 100%;
 `;
 
+const landingStyles = css`
+  flex-direction: row;
+  padding-left: ${theme.size.medium};
+  img {
+    width: ${theme.size.xlarge};
```
seungpark, line 24: this overrides the width for the first set of cards, but on a minor note, the second icons seem smaller than design as well (design shows roughly 42-46px width):

image

seungpark (changes requested):
largely LGTM! left a comment regarding:
1 - the column option to be respected
and 2 - the second set of card groups to have bigger icons

src/components/Card/CardGroup.js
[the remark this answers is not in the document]
mayaraman19: ok, changed it so on the landing page the column option is respected, but leaving the default for cardgroup as it was before to be 2 columns

seungpark (approved):
this LGTM!

should merge the content change also !

PR #1021, opened 2024-02-22
DOP-4394: save files to public

seungpark:
### Stories/Links:

DOP-4394

### Current Behavior:

[atlas app service page with missing icons](https://www.mongodb.com/docs/atlas/app-services/sync/)

### Staging Links:

[staged atlas app service page](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/atlas-app-services/seung.park/DOP-4394/index.html)

### Notes:
- This is a quick fix to allow directives that use images to work. Images will be stored in both `public` and `src/images` directories. The first is for public access, and the latter is for image processing

### README updates

### DIFF
--- a/src/utils/setup/save-asset-files.js
+++ b/src/utils/setup/save-asset-files.js
@@ -6,8 +6,13 @@ const GATSBY_IMAGE_EXTENSIONS = ['webp', 'png', 'avif'];
 const isPreview = isGatsbyPreview();
 
 const saveFile = async (file, data) => {
-  // if this is preview, always save to 'public'
-  // images saved in /src/images are transformed and processed by gatsby-source-filesystem and gatsby-transformer-sharp
+  // save files both to "public" and "src/images" directories
+  // the first is for public access, and the second is for image processing
+  await fs.mkdir(path.join('public', path.dirname(file)), {
+    recursive: true,
+  });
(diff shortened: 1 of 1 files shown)


src/utils/setup/save-asset-files.js
```diff
@@ -6,8 +6,13 @@ const GATSBY_IMAGE_EXTENSIONS = ['webp', 'png', 'avif'];
 const isPreview = isGatsbyPreview();
 
 const saveFile = async (file, data) => {
-  // if this is preview, always save to 'public'
-  // images saved in /src/images are transformed and processed by gatsby-source-filesystem and gatsby-transformer-sharp
```
mmeigs, line 10: nit: Wondering if we want to keep this comment as I don't think this has changed? Totally non-blocking

mmeigs (approved):
Staging looks good, and change is safe. Not totally sure I understand the entire issue, but I'll look back at the original PR!

src/utils/setup/save-asset-files.js
seungpark: i updated the comment as well! it has a [fallback in the Image component ](https://github.com/mongodb/snooty/blob/main/src/components/Image.js#L57)as well

PR #1022, opened 2024-02-22
DOP-4393: fix broken language selector

anabellabuckvar:
### Stories/Links:

[DOP-4393](https://jira.mongodb.org/browse/DOP-4393)

### Current Behavior:

Clicking the language selector causes the screen to go blank.
example: https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-stage/

### Staging Links:

[Staging Link with working Language Selector](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/anabella.buckvar/DOP-4393/manage-connections-aws-lambda/)

[Staging Link with working Deprecated Version Selector](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/anabella.buckvar/DOP-4393/legacy/index.html)

### README updates

### DIFF
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -83,9 +83,9 @@ const Select = ({
           <Option
             className={cx(optionStyling)}
             key={choice.value}
-            glyph={choice.icon}
             value={choice.value}
             role="option"
+            glyph={choice.glyph ? choice.glyph : null}
           >
             {choice.text}
           </Option>
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -18,6 +18,13 @@ const portalStyle = css`
   position: relative;
 `;
 
+const iconStyling = css`
+  display: inline-block;
+  margin-right: ${theme.size.small};
+  max-height: 20px;
+  width: 30px;
+`;
+
 /* Override LG mobile style of enlarged mobile font */
 const selectStyle = css`
   @media ${theme.screenSize.upToLarge} {
@@ -80,7 +87,14 @@ const Select = ({
         {...props}
       >
         {choices.map((choice) => (
-          <Option className={cx(optionStyling)} key={choice.value} value={choice.value} role="option">
+          <Option
+            className={cx(optionStyling)}
+            key={choice.value}
+            value={choice.value}
+            glyph={choice.icon}
+            role="option"
+          >
+            {choice.tabSelectorIcon && <choice.tabSelectorIcon className={cx(iconStyling)} />}
             {choice.text}
           </Option>
         ))}
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -80,13 +80,7 @@ const Select = ({
         {...props}
       >
         {choices.map((choice) => (
-          <Option
-            className={cx(optionStyling)}
-            key={choice.value}
-            value={choice.value}
-            role="option"
-            // glyph ={choice.icon ? : choice.icon : null}
-          >
+          <Option className={cx(optionStyling)} key={choice.value} value={choice.value} role="option">
             {choice.text}
           </Option>
         ))}
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -85,7 +85,7 @@ const Select = ({
             key={choice.value}
             value={choice.value}
             role="option"
-            glyph={choice.glyph ? choice.glyph : null}
+            // glyph ={choice.icon ? : choice.icon : null}
           >
             {choice.text}
           </Option>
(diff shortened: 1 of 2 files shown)


src/components/Select.js
```diff
@@ -83,9 +83,9 @@ const Select = ({
           <Option
             className={cx(optionStyling)}
             key={choice.value}
-            glyph={choice.icon}
             value={choice.value}
             role="option"
+            // glyph ={choice.icon ? : choice.icon : null}
```
mmeigs, line 88: Confused by this change. Are we just removing glyph completely?

anabellabuckvar: Yes, that's what Allison just instructed.

caesarbell: A bit confused, so `glyph` was breaking the Select component? That is interesting.

anabellabuckvar: Yes.

mmeigs (approved):

src/components/Select.js
seungpark: think we have to use the LG <Icon> in this `glyph` property. the [DeprecatedVersionSelector already has it](https://github.com/mongodb/snooty/blob/main/src/components/DeprecatedVersionSelector.js#L93) in Icon component form, whereas [Tabs have it in regular component with SVG ](https://github.com/mongodb/snooty/blob/main/src/components/icons/C.js#L3)

seungpark (approved):
this LGTM. can we follow up with an Icon component wrapper to pass into glyph (if not already an icon)

src/components/Select.js
rayangler: Something we can do to support both is to just rename the tab selector's icon field to something like `tabSelectorIcon` and conditionally render it in the `Select` component using the old logic that was removed. So in `Select`, we can have:

```js
{choices.map((choice) => (
  <Option
    className={cx(optionStyling)}
    key={choice.value}
    glyph={choice.icon}
    value={choice.value}
    role="option"
>
  {choice.tabSelectorIcon && <choice.tabSelectorIcon className={cx(iconStyling)} />}
  {choice.text}
</Option>
```

But if the consensus is to remove it altogether, that's okay with me

schmalliso: IN AN IDEAL WORLD, we'd keep both, but I don't really care as long as shit's not broken. The glyph on the legacy drop-down is a nice-to-have not essential.

The tab selector icons, as it stands, are odd in that they're JS files not just regular svgs, and we may want to address that in the near future, but for now whatever works works for me.

anabellabuckvar: Just pushed up some new changes that allow icons to appear in the (functioning) Language Selector and in the Deprecated Version Selector!

src/components/Tabs/TabSelectors.js
```diff
@@ -24,7 +24,7 @@ const makeChoices = ({ name, iconMapping, options }) =>
   Object.entries(options).map(([tabId, title]) => ({
     text: getPlaintext(title),
     value: tabId,
-    ...(name === 'drivers' && { icon: iconMapping[tabId] }),
+    ...(name === 'drivers' && { tabSelectorIcon: iconMapping[tabId] }),
```
mmeigs, line 27: Maybe for safety we can also have a check on the existence of the icon mapping?

`...(name === 'drivers' && iconMapping[tabId] && { tabSelectorIcon: iconMapping[tabId] }),`

caesarbell (approved):
LGTM

src/components/Tabs/TabSelectors.js
anabellabuckvar: I don't ~believe it's necessary because the select component checks if choice.tabSelectorIcon is truthy before rendering it, so tabSelectorIcon has a null/undefined value like in the Deprecated Version Selector component, the Select component is still safe and renders correctly! That being said, I'm happy to still include the check for extra safety if you'd like!

src/components/Select.js
```diff
@@ -80,13 +80,7 @@ const Select = ({
...
       >
         {choices.map((choice) => (
-          <Option
-            className={cx(optionStyling)}
-            key={choice.value}
-            glyph={choice.icon}
-            value={choice.value}
-            role="option"
-          >
+          <Option className={cx(optionStyling)} key={choice.value} value={choice.value} role="option">
```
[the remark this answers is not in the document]
anabellabuckvar: yup, the value for choice.icon being passed into each tab selector component isn't LG `Icon`, which is the type that the Option component's `glyph` expects.

rayangler (approved):

src/components/Tabs/TabSelectors.js
mmeigs: Ah wonderful! Thanks!

PR #1024, opened 2024-02-26
DOP-4352: Use custom functions for Netlify deploy events

rayangler:
### Stories/Links:

DOP-4352

### Staging Links:

[Example Netlify site](https://mongodb-prd-jeff-allen-mongo.netlify.app/docs/DOCS-15096/query-api/) - Adding here to show the site doesn't break.
[Example Gatsby Cloud site](https://preview-mongodbrayangler.gatsbyjs.io/docs/DOP-testing/) - Adding here to show the site doesn't break. [Job status](https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=65de0c917111165982c18d25) was successfully updated post-build.

### Notes:
Netlify has support for custom functions that can be triggered when certain events occur, such as when a Netlify deployment is made (see: [docs](https://docs.netlify.com/functions/trigger-on-events/)). This means unlike with Gatsby Cloud, we can call the post-build webhook with custom data (such as the Autobuilder job ID) after the content is actually deployed to the site's CDN, rather than when the plugin finishes the frontend build. This will help reduce the latency between when a user is notified that their job is complete vs. when their changes are actually available on the Netlify site. Not relying on the plugin's `postBuild` step also makes tracking build failures more accurate, as opposed to us having to handle it ourselves within the plugin code.

### README updates

- - [x] This PR introduces changes that should be reflected in the README, and I have made those updates. (I added a README for Netlify, if that counts)


plugins/gatsby-source-snooty-preview/utils/post-build.js
```diff
@@ -13,7 +13,7 @@ const constructSnootyHeader = (payloadString) =>
  * build is finished.
  * @param {object} webhookBody - The webhook body passed to the source plugin to
  * initiate the preview build.
- * @param {string} status - The status of the build, typically "completed" or "failed".
+ * @param {'completed' | 'failed'} status - The status of the build, typically "completed" or "failed".
```
mmeigs, line 16: <3

netlify/utils.js
```diff
@@ -0,0 +1,30 @@
...
+  // This file does not currently exist, but should be created on Netlify builds as part of `npm run build:netlify`.
+  // The INCOMING_HOOK_BODY env var is not automatically passed along to functions, so we use a txt file to save it
+  const relativeFilePath = '../build-hook.txt';
+  const buildHookDataString = fs.readFileSync(path.resolve(__dirname, relativeFilePath), 'utf-8');
+  if (!buildHookDataString) {
+    console.log('No build hook data found.');
+    return;
+  }
+  return JSON.parse(buildHookDataString);
+};
```
mmeigs, line 18: I'm trying to think if we need a try catch here. I want to make sure we always update the job with 'completed' or 'failed' and I'm wondering if reading the file or parsing fails if the thread would make it to the `callPostBuildWebhook`.

rayangler: Good point! I thought about a try-catch here, but I was having a tough time trying to figure out what the ideal behavior would be if we did catch (open to recommendations!). If we fail to read or fetch the payload, we won't have access to the Autobuilder job ID that triggered the Netlify build, making it difficult to know which job ID to send along the post-build webhook anyways. As is, not catching will lead to the error message being propagated up to the logs of the function.

Tangentially, this is a pretty hacky implementation inspired by this [one forum post](https://answers.netlify.com/t/access-build-webhook-payload-in-deploy-succeeded-js/55394/11), since Netlify doesn't have a built-in way to pass along custom data for post-deployment events 😢

mmeigs: Ah, I see! Gosh, there's no way to even pinpoint which job it was to then send off a quick request to Atlas. Ok, sounds acceptable to me! I guess we delete jobs after a while anyway. Thanks for the info! 🙏

mmeigs (approved):

netlify/utils.js
rayangler: Yeah, it's definitely not ideal and I'm assuming this is just one of the limitations for using a third-party platform for our platform's use case. Worst case, if the function fails to parse the job ID to use and call the post-build webhook, the Autobuilder job will hang and eventually time out like it would with Gatsby Cloud builds currently.

PR #1027, opened 2024-02-28
DOP-4379: Modify Figure.test.js to fail if "class" option is not properly accessed

mayaraman19:
### Stories/Links:

DOP-4379

### Current Behavior:

Currently, tests that require the `<Image>` component (such as `Figure.test.js` and `Lightbox.test.js`) do not check if some classes are properly accessed, leading to potential for regressions/class attributes that are not correctly accessed.

This change adds `class`, `scale`, `align`, and `alt` options to the data for `Figure.test.json` and updates relevant snapshots so that if these options are not correctly accessed in `Image.js` the tests will fail.

### Notes:

I wasn't sure if I should add an additional test/data for `Lightbox.test.js` to check that an "alt" attribute is present in the rendered data even if it's not present in the options. If this is necessary I can add it.

### README updates


seungpark (approved):
thanks for making this change!

PR #1034, opened 2024-03-06
DOP-4398: Add hreflang links to every page

rayangler:
### Stories/Links:

DOP-4398

### Current Behavior:

[Server docs on prod](https://www.mongodb.com/docs/manual)
[Server docs on pre-prod](https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/manual/)
[Server docs on main](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/raymund.rodriguez/main/index.html) - Note that the footer at the bottom has access to many languages

### Staging Links:

[Server docs staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/raymund.rodriguez/DOP-4398/index.html)

### Notes:

Alternate links for different languages have been introduced. Note that these currently use the static `siteUrl` config value to ensure that the base URL is defined at build time. This does mean that when built in staging links or pre-prod, the links will always point to the main https://mongodb.com site. Since SEO isn't typically taken into account for non-prod sites, this _should_ be fine (canonical links and sitemaps have similar behavior). 

To validate, either inspect the page or look at the page source (make sure to add `index.html` to the end), and then search for `hreflang` to see all 4 links. Each link for a page should match the pathname of that page.

This PR also fixes a bug regarding languages available through the UnifiedFooter's dropdown. Since we only want to show valid languages, and the footer doesn't currently have a way to pass in available languages, we use client side logic to filter the languages that we want. However, this breaks when the UnifiedFooter is lazy loaded since the client side logic would begin before the component loaded. The solution was to wrap the footer in its own component and move the client side logic there to ensure all of the logic is done at the same time.

### README updates

### DIFF
--- a/src/utils/locale.js
+++ b/src/utils/locale.js
@@ -1,0 +1,28 @@
+import { withPrefix } from 'gatsby';
+import { assertTrailingSlash } from './assert-trailing-slash';
+
+// Update this as more languages are introduced
+export const AVAILABLE_LANGUAGES = [
+  { language: 'English', code: 'en-us' },
+  { language: '简体中文', code: 'zh-cn' },
+  { language: '한국어', code: 'ko-kr' },
+  { language: 'Português', code: 'pt-br' },
+];
+
+/**
+ * Returns a mapping of a page's URL and its equivalent URLs in different languages.
+ * @param {string} slug
+ */
+export const getLocaleMapping = (location, slug) => {
+  // handle the `/` path
+  const slugForUrl = slug === '/' ? `${withPrefix('')}` : `${withPrefix(slug)}`;
+  const localeHrefMap = {};
+
+  AVAILABLE_LANGUAGES.forEach(({ code }) => {
+    const langPrefix = code === 'en-us' ? '' : `/${code}`;
+    const targetUrl = `${location.origin}${langPrefix}${slugForUrl}`;
+    localeHrefMap[code] = assertTrailingSlash(targetUrl);
+  });
+
+  return localeHrefMap;
+};
--- a/src/utils/locale.js
+++ b/src/utils/locale.js
@@ -13,14 +13,14 @@ export const AVAILABLE_LANGUAGES = [
  * Returns a mapping of a page's URL and its equivalent URLs for different languages.
  * @param {string} slug
  */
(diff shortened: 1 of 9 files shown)


src/utils/locale.js
```diff
@@ -0,0 +1,28 @@
...
+export const AVAILABLE_LANGUAGES = [
+  { language: 'English', code: 'en-us' },
+  { language: '简体中文', code: 'zh-cn' },
+  { language: '한국어', code: 'ko-kr' },
+  { language: 'Português', code: 'pt-br' },
+];
+
+/**
+ * Returns a mapping of a page's URL and its equivalent URLs for different languages.
+ * @param {string} slug
```
seungpark, line 14: minor nitpick, missing param for siteUrl here

tests/unit/Head.test.js
```diff
@@ -123,4 +125,21 @@ describe('Head', () => {
...
+  describe('hreflang links', () => {
+    beforeEach(() => {
+      mockStaticQuery({
+        siteUrl: DOTCOM_BASE_URL,
+      });
+      useSnootyMetadata.mockImplementation(() => ({ ...mockEOLSnootyMetadata, eol: false }));
+    });
+
+    it.each([['/'], ['foo']])('renders based on slug', (slug) => {
+      const mockPageContext = { ...mockHeadPageContext, slug };
```
seungpark, line 138: ❤️

src/utils/locale.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+export const getLocaleMapping = (siteUrl, slug) => {
```
seungpark, line 16: could also write a unit test for this util

seungpark (approved):
LGTM! good to move the footer to its own component ➕

anabellabuckvar (approved):

src/utils/locale.js
rayangler: Done! Thanks for the heads up. Forgot to update documentation as the function changed

src/utils/locale.js
rayangler: Done!

PR #1040, opened 2024-03-14
EAI-265: Add Vector Search to chatbot disclaimer text

nlarew:
### Stories/Links:

[EAI-265: Add Vector Search to chatbot disclaimer text](https://jira.mongodb.org/browse/EAI-265)

### Current Behavior:

### Staging Links:

### README updates


src/components/Widgets/ChatbotWidget/PoweredByAtlasSearch.js
```diff
@@ -0,0 +1,39 @@
...
+    return (
+      <Body className={className}>
+        Powered by Atlas Vector Search. <VectorSearchLink>Learn More.</VectorSearchLink>
+      </Body>
+    );
+  }
+  if (linkStyle === 'text') {
+    return (
+      <Body className={className}>
+        Powered by <VectorSearchLink>Atlas Vector Search</VectorSearchLink>
```
mmeigs, line 33: If you'll want to use `className` for styling, you'll need to also use `cx` imported from `@leafygreen-ui/emotion` like in [this file](https://github.com/mongodb/snooty/blob/main/src/components/Paragraph.js).

src/components/Widgets/ChatbotWidget/PoweredByAtlasSearch.js
```diff
@@ -0,0 +1,39 @@
...
+ * Duplicated version of the PoweredByAtlasVectorSearch component from
+ * mongodb-chatbot-ui
+ *
+ * We define our own version here because we faced a rendering issue
+ * when importing the original component from mongodb-chatbot-ui.
+ *   (The issue is that the open modal renders with opacity: 0)
+ * This seems to be a side effect of lazy loading the component.
+ * We have to lazy load the component because mongodb-chatbot-ui
+ * does not currently support SSR.
+ */
```
mmeigs, line 11: Thanks for this

src/components/Widgets/ChatbotWidget/PoweredByAtlasSearch.js
```diff
@@ -0,0 +1,39 @@
...
+
+const url = 'https://www.mongodb.com/products/platform/atlas-vector-search?tck=mongodb_ai_chatbot';
+const VectorSearchLink = ({ children }) => (
+  <Link href={url} hideExternalIcon>
+    {children}
+  </Link>
+);
+
+export function PoweredByAtlasVectorSearch({ className, linkStyle = 'learnMore' }) {
+  if (linkStyle === 'learnMore') {
```
mmeigs, line 23: curious: Just wondering what/when this eventuality is for?

mmeigs (commented):
One barely blocking comment on styling not being applied.

Also, this is out of my jurisdiction, but I'm wondering if we're okay with the chatbot on most screens already being scrolled down when initially opened because of the new additions recently (PoweredByAtlasSearch and adding more default prompts). 

Screenshot 2024-03-18 at 2 53 01 PM

mmeigs:
Here's a staging link! https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/matt.meigs/powered-by-vector-search/index.html

src/components/Widgets/ChatbotWidget/PoweredByAtlasSearch.js
nlarew: We use it in the dropdown menu for the `InputBarTrigger` (i.e. the version of the chatbot that's on the docs landing page)

For that version we're using the `DocsChatbot` component from `mongodb-chatbot-ui` which includes the original version of this component using `linkStyle={learnMore}`.

We don't use `learnMore` in Snooty so this branch is technically unnecessary. I kept it in to stay as close as possible to the original implementation since we should eventually remove these duplicated versions/fix the issue and it'll be helpful to have the same API.

mmeigs: Understandable

nlarew:
@mmeigs good call on the suggested prompts/scroll situation. We're gonna remove one of the existing suggested prompts for now as a replacement.

Updated with that change + your review comment about `cx`

src/components/Widgets/ChatbotWidget/ChatbotFab.js
```diff
@@ -1,8 +1,10 @@
-import { lazy } from 'react';
+import { lazy, Fragment } from 'react';
 import styled from '@emotion/styled';
+import { css } from '@emotion/react';
```
mmeigs, line 3: The `cx` and `css` imports have to be from the same library to work together, ie if you just change this `css` to be imported from '@leafygreen-ui/emotion', then it all works!

mmeigs (commented):
Deleting one prompt makes it look great! 
Okay, ONE last thing which I should have been more clear on.

src/components/Widgets/ChatbotWidget/ChatbotFab.js
nlarew: Oh duh I should've realized this. Good catch!

nlarew: Ok updated to use `css` from `@leafygreen-ui/emotion`

Are we concerned that this page also uses the `styled` API not from leafygreen?

mmeigs: Not concerned. That plays fine together!

mmeigs (approved):

seungpark (approved):

PR #1041, opened 2024-03-14
DOP-4457: Strip away locale code from slugs before localizing them

rayangler:
### Stories/Links:

DOP-4457

### Current Behavior:

[Server prod](https://www.mongodb.com/docs/manual/)

### Staging Links:

[Server staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/raymund.rodriguez/DOP-4457/index.html) - mostly to spot check that things (the version dropdown, hreflang links, and footer language selector) aren't broken. Unfortunately, this will need to be tested in production due to Smartling's production-only logic.

### Notes:

The situation is that Smartling overrides some frontend logic on their end to allow the frontend to properly link to other docs pages within the same locale. We purposely have logic that localizes URLs ourselves, so that we can link to them (for example: the language selector at the footer). This works fine on the English docs, but on the translated docs, Smartling's overridden localized slugs cause our localization of the same slug to result in two locale codes in the URL. A solution would be to remove the locale code from the beginning of a slug (if it exists), before localizing the slug.

### README updates

### DIFF
--- a/src/components/SEO.js
+++ b/src/components/SEO.js
@@ -14,7 +14,7 @@ const SEO = ({ pageTitle, siteTitle, showDocsLandingTitle, canonical, slug }) =>
   const hrefLangLinks = Object.entries(localeHrefMap).map(([langCode, href]) => {
     const hrefLang = langCode === 'en-us' ? 'x-default' : langCode;
     // Do not remove class. This is used to prevent Smartling from potentially overwriting these links
-    const smartlingNoRewriteClass = 'sl_norewrite';
+    const smartlingNoRewriteClass = 'sl_opaque';
     return <link key={hrefLang} className={smartlingNoRewriteClass} rel="alternate" hrefLang={hrefLang} href={href} />;
   });
 
--- a/src/utils/locale.js
+++ b/src/utils/locale.js
@@ -20,7 +20,7 @@ const validateCode = (potentialCode) => !!AVAILABLE_LANGUAGES.find(({ code }) =>
  * @param {string} slug
  * @returns {string}
  */
-export const stripLocale = (slug) => {
+const stripLocale = (slug) => {
   // Smartling has extensive replace logic for URLs and slugs that follow the pattern of "https://www.mongodb.com/docs". However,
   // there are instances where we can't rely on them for certain components
   if (!slug) {
--- a/src/utils/locale.js
+++ b/src/utils/locale.js
@@ -2,6 +2,7 @@ import { withPrefix } from 'gatsby';
 import { assertTrailingSlash } from './assert-trailing-slash';
 import { isBrowser } from './is-browser';
 import { normalizePath } from './normalize-path';
+import { removeLeadingSlash } from './remove-leading-slash';
 
 // Update this as more languages are introduced
 export const AVAILABLE_LANGUAGES = [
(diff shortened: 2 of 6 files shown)


src/components/SEO.js
```diff
@@ -14,7 +14,7 @@ const SEO = ({ pageTitle, siteTitle, showDocsLandingTitle, canonical, slug }) =>
   const hrefLangLinks = Object.entries(localeHrefMap).map(([langCode, href]) => {
     const hrefLang = langCode === 'en-us' ? 'x-default' : langCode;
     // Do not remove class. This is used to prevent Smartling from potentially overwriting these links
-    const smartlingNoRewriteClass = 'sl_norewrite';
+    const smartlingNoRewriteClass = 'sl_opaque';
```
mmeigs, line 17: Curious: what Smartling behavior changes with this?

src/utils/locale.js
```diff
@@ -50,13 +80,14 @@ export const localizePath = (pathname, localeCode) => {
     return '';
   }
 
+  const unlocalizedPath = stripLocale(pathname);
   const code = localeCode && validateCode(localeCode) ? localeCode : getCurrLocale();
   const languagePrefix = code === 'en-us' ? '' : `${code}/`;
-  let newPath = languagePrefix + pathname;
+  let newPath = languagePrefix + unlocalizedPath;
   if (pathname.startsWith('/')) {
```
mmeigs, line 87: Should this conditional use the `unlocalizedPath` or does it matter because it looks like you've tried to replicate the leading character anyway?

src/components/SEO.js
rayangler: Great question! This was the solution provided to me by Arnaud (from Smartling) to prevent the hreflang link from being overwritten. For some reason the `sl_norewrite` class wasn't doing it so he wanted me to try it the opaque class. Here's [a link to the docs](https://help.smartling.com/hc/en-us/articles/360000322554-Preventing-Link-Rewriting), although I'm not sure how different they'd be for this use case.

I should've added this to the description, but I deployed one versions of the Node docs, and you can see in [Simplified Chinese docs](https://www.mongodb.com/zh-cn/docs/drivers/node/v4.8/) that the hreflang link for `x-default` is incorrect, most likely because it's being overwritten.

src/utils/locale.js
```diff
@@ -13,6 +14,35 @@ export const AVAILABLE_LANGUAGES = [
...
+  }
+
+  // Normalize the slug in case any malformed slugs appear like: "//zh-cn/docs"
+  const slugWithSlash = slug.startsWith('/') ? slug : `/${slug}`;
+  const normalizedSlug = normalizePath(slugWithSlash);
+  const firstPathSlug = normalizedSlug.split('/', 2)[1];
+
+  // Replace from the original slug to maintain original form
+  const res = validateCode(firstPathSlug) ? normalizePath(slug.replace(firstPathSlug, '')) : slug;
+  if (res.startsWith('/') && !slug.startsWith('/')) {
```
seungpark, line 37: [not blocking] seems like we are doing a lot of slash handling. would prefer if we could handle it in the exported function and helpers can just do simple string manipulations on input. also not seeing where this is called so may not have to export

src/utils/locale.js
rayangler: I believe it should be the `pathname` to ensure the "original" slug is the source of truth. But I don't think it matters? 🤔

seungpark (approved):
discussed offline weird behavior by Smartling team to overwrite the slug property (written in build time). will have to keep tabs on their changes to see if they submit another fix...

src/utils/locale.js
rayangler: Agreed that there's a lot of slash handling... I've made the function local, but as discussed offline, I'm afraid of how Smartling is handling these slugs, so I want to ensure that the output string doesn't accidentally manipulate the input. Happy to revisit this logic during a less stressful situation to help clean up

PR #1049, opened 2024-04-01
DOP-3457 - update Typography/links to v18

seungpark:
### Stories/Links:

DOP-3457
DOP-3091

### Current Behavior:

[Page from original screenshot in issue](https://www.mongodb.com/docs/charts/bin-data/) - bottom of page shows prev/next page links as described in DOP-3457
[Example of link not line breaking in note in cloud docs](https://www.mongodb.com/docs/atlas/customize-storage/) if you find `WiredTiger storage engine`, you will see the link does not break in the note
[Example app service page ](https://www.mongodb.com/docs/atlas/app-services/rules/examples/)

### Staging Links:

[Staged changes from original JIRA issue screenshot](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/charts/seung.park/DOP-3457/bin-data/index.html#limit-data)
[Addressed line breaking](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/seung.park/DOP-3457/customize-storage/index.html)
[Staged app service page](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/atlas-app-services/seung.park/DOP-3457/rules/examples/index.html) - validating banners here as well

### Notes:
- LG/typography has been updated to update Links since the [latest release](https://github.com/mongodb/leafygreen-ui/commits/%40leafygreen-ui/typography%4018.3.0/) includes the fix for [the line wrapping](https://jira.mongodb.org/browse/LG-2247)Please verify this does not have any unwarranted effects. I didn't find any but could use second eye
- DOP-3457 has been addressed with an update for internalpagenav components
- DOP-3091 has been addressed with an update to LG/Link@v18 but also an overwrite to `display:inline` property to match gatsby (and default) anchor tags

### README updates


seungpark:
> I've been waiting for this change. Looks so good.
> 
> Only thing. Only to do with the "associated" ticket. [This ticket](https://jira.mongodb.org/browse/DOP-3091) actually is referring to another kind of wrapping. I'm not sure how easy or do-able this is for the specific spot or if it can be a general change... but where you see the first two links [here in your staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/atlas-app-services/seung.park/DOP-3457/apps/cicd/index.html), I think removing display: inline-flex fixes it? But then again, I feel like this was the original issue and LG just worked on the underline wrapping rather than this issue.

thanks @mmeigs ! i included the other ticket in this as well since its just a simple overwrite as you said above. verified this with LG team and this may be an update on their end, but we can expect it not in v18^ and overwrite it on our end

src/components/InternalPageNav.js
```diff
@@ -1,62 +1,68 @@
...
 `;
 
-const textAlignLeft = css`
-  text-align: left;
-`;
-
-const textAlignRight = css`
-  text-align: right;
+const titleSpanStyling = css`
+  // display: inline-block;
```
mmeigs, line 30: Do we want to remove?

seungpark: removed, thanks for the catch  !

mmeigs (approved):
LGTM!

PR #1061, opened 2024-04-10
:bug: DOP-4514 using normalizePath as a catch-all for double slashes

caesarbell:
### Stories/Links:

DOP-4514

### Current Behavior:

[Customize Cluster Storage](https://www.mongodb.com/docs/atlas/customize-storage/)

### Staging Links:

[Staged Commit](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/795b0c49fad7fe1754f35c17d2dea1fa72f691f9/master/cloud-docs/caesarbell/DOP-4514/index.html)

### Notes:

This PR does the following; 

-  Uses `normalizePath` as a catch-all for double slashes within the canonical URL.

### README updates

### DIFF
--- a/src/hooks/use-canonical-url.js
+++ b/src/hooks/use-canonical-url.js
@@ -11,7 +11,7 @@ export const useCanonicalUrl = (meta, metadata, slug, repoBranches) => {
   const pathPrefix = generatePrefix(urlSlug, siteMetadata, siteBasePrefix);
 
   // Use default logic assuming there is no canonical provided from the meta directive
-  let canonical = `${siteUrl}${normalizePath(`${pathPrefix}/${slug === '/' ? '' : slug}`)}`;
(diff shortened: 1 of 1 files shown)


caesarbell:
Also, it was hard for me to test this exact case with the `pathPrefix` or `siteBasePrefix` for canonical URLs. I am open to any suggestions to replicate the issue locally.

mmeigs (commented):
I think for full safety's sake, we could use normalizePath on the entire canonical string rather than just the subset? Just my opinion, depending on what others think

caesarbell:
> I think for full safety's sake, we could use normalizePath on the entire canonical string rather than just the subset? Just my opinion, depending on what others think

I initially thought that, but the `SiteUrl` value is `https://www.mongodb.com`, which from what I understand those double slashes would be removed, which we don't want.

mmeigs (approved):

PR #1062, opened 2024-04-12
DOP-4543: Generalize card styles for programming language landing pages

rayangler:
### Stories/Links:

DOP-4543

### Current Behavior:

[Drivers prod](https://www.mongodb.com/docs/drivers/)

### Staging Links:

[Drivers staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOP-4510/drivers/raymund.rodriguez/DOP-4543/index.html)
[Mocked programming language landing page](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOP-4510/drivers/raymund.rodriguez/DOP-4543/php-libraries/index.html)
[Landing staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/raymund.rodriguez/DOP-4543/index.html)

### Notes:

* The programming language landing pages will be using cards in the same format/shape as the cards of the drivers homepage. However, the margins that come with the card group for the drivers homepage are coupled with the `:type: drivers` option.
* Solution: Decouple the margins styles from the card styles such that the margins.
  * Since the margins are only (currently) desired for the drivers homepage, the `drivers-index` template used for that homepage will determine whether the margins are used. To test, change browser screen size.
  * Since the card style will be used for more than just drivers (semantically), we use the style based on whether the card group's `type` is `drivers` (to avoid immediate breaking changes) or when using the `style` `center-content` (potentially preferable as a generalized style option). See this [parser PR](https://github.com/mongodb/snooty-parser/pull/584) to see introduction of new style. Support for the `drivers` `type` can potentially be removed whenever reStructuredText pages no longer rely on that option.
* (Based on discussion below) Also generalizes the styling for cards associated with the bottom cards on the `docs-landing` homepage. This is to allow new programming landing pages to reuse a similar style. After this is merged and released, I should be able to update the styles accordingly. 

To test on a page:

```
.. card-group::
   :columns: 3
   :style: center-content

   .. cards here

.. card-group::
   :columns: 3
   :style: large-icon

   .. cards here
```

### README updates

### DIFF
--- a/src/components/Card/CardGroup.js
+++ b/src/components/Card/CardGroup.js
@@ -104,13 +104,7 @@ const CardGroup = ({
 
   return (
     <StyledGrid
-      className={[
-        'card-group',
-        isCompact ? 'compact' : '',
-        isExtraCompact ? 'extra-compact' : '',
-        isDriversTemplate ? 'drivers' : '',
-        className,
-      ].join(' ')}
+      className={['card-group', className].join(' ')}
       columns={columns}
       noMargin={true}
       isCarousel={isCarousel}
--- a/src/components/Card/CardGroup.js
+++ b/src/components/Card/CardGroup.js
@@ -98,6 +98,7 @@ const CardGroup = ({
   const isCarousel = layout === 'carousel';
   // Keep "type" for backwards compatibility, but it might be good to generalize to "style"
   const isCenterContentStyle = type === 'drivers' || style === 'center-content';
+  const isLargeIconStyle = style === 'large-icon';
   const isDriversTemplate = page?.options?.template === 'drivers-index';
   const isLanding = page?.options?.template === 'landing';
 
@@ -124,6 +125,7 @@ const CardGroup = ({
           isCompact={isCompact}
           isExtraCompact={isExtraCompact}
           isCenterContentStyle={isCenterContentStyle}
+          isLargeIconStyle={isLargeIconStyle}
           page={page}
         />
       ))}
--- a/src/components/Card/CardGroup.js
+++ b/src/components/Card/CardGroup.js
@@ -96,7 +96,9 @@ const CardGroup = ({
   const isCompact = style === 'compact';
   const isExtraCompact = style === 'extra-compact';
   const isCarousel = layout === 'carousel';
-  const isForDrivers = type === 'drivers';
+  // Keep "type" for backwards compatibility, but it might be good to generalize to "style"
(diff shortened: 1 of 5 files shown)


mayaraman19:
perhaps I am misunderstanding the ticket, but I don't see a difference in the margins between the drivers staging and the mockup staging?

mayaraman19:
also (perhaps not in the scope of this ticket, but just want to note somewhere before I forget) it seems a similar decoupling can be done for the "landing-bottom" cards (i.e. cards in the bottom row of the landing page) since cards similar to this style are also in the design for the programming landing pages?

rayangler:
> also (perhaps not in the scope of this ticket, but just want to note somewhere before I forget) it seems a similar decoupling can be done for the "landing-bottom" cards (i.e. cards in the bottom row of the landing page) since cards similar to this style are also in the design for the programming landing pages?

I can see the benefit in decoupling "landing-bottom" cards too. Let me try to see if I can get something reasonable down. Thank you for pointing that out in the designs! I actually didn't realize they shared the same styles as the landing ones.

> perhaps I am misunderstanding the ticket, but I don't see a difference in the margins between the drivers staging and the mockup staging? but i could also see how right now that would be expected behavior, with the ability to add different margins now that margin and drivers have been decoupled.

If you go to the drivers homepage on the staging link or prod link, and then resize the browser to be smaller, the margins should be different from other cards. Let me know if a screenshot would be more beneficial!

src/components/Card/CardGroup.js
```diff
@@ -105,13 +108,13 @@ const CardGroup = ({
         'card-group',
         isCompact ? 'compact' : '',
         isExtraCompact ? 'extra-compact' : '',
-        isForDrivers ? 'drivers' : '',
+        isDriversTemplate ? 'drivers' : '',
```
seungpark, line 111: do we need these classnames - or are these for debugging purposes? seems like its not used for styling

seungpark (approved):
this LGTM. friendly reminder to update docs-landing for largeIconStyle before close

src/components/Card/CardGroup.js
rayangler: Apparently, they were used to help with styling for the `landing` template ([source](https://github.com/mongodb/snooty/blob/d7f41ddb5056bf8661f55031b2e2a5b81398a387/src/templates/landing.js#L47)) once upon a time, but this no longer seems to be the case. I've removed these accordingly. Should be easy to add back if we need them again

PR #1066, opened 2024-04-15
imgs in screenshot bugfix

mmeigs:
### Stories/Links:

DOP-4471

### Current Behavior:

Screenshot 2024-04-15 at 2 27 12 PM

### Staging Behavior:

Screenshot 2024-04-15 at 2 24 21 PM

### Notes:

Some images were not being captured in screenshots. 

This took quite a bit of investigation and experimentation. I did not really find this issue out on the web, but after taking the time to compare which images did and didn't carry over. The fix was to remove the attribute srcset just for the screenshots.

### README updates


seungpark (approved):
ah this may be due to images having a `srcset` attribute without a `sizes` attribute. for now, this is fine but an improvement may be to indicate sizes for these srcSets as well, as outlined [here](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#how_do_you_create_responsive_images)

PR #1067, opened 2024-04-16
DOP-4533: FE removal of the SidenavBackButton

mayaraman19:
### Stories/Links:

DOP-4533

### Current Behavior:

[cloud docs production](https://www.mongodb.com/docs/atlas/security/ip-access-list/)
[manual production](https://www.mongodb.com/docs/manual/administration/change-streams-production-recommendations/)
[MTC production](https://www.mongodb.com/docs/tools-and-connectors/)

### Staging Links:

[cloud-docs staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/maya/DOP-4533/security/ip-access-list/index.html)
[manual staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/v7.0/docs/maya/DOP-4533/administration/change-streams-production-recommendations/)
[MTC staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/maya/DOP-4533/tools-and-connectors/)
[Open API preview page](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/maya/DOP-4533/openapi/preview/)

### Notes:

I did not remove the `SideNavBackButton` from the OpenAPI component, so that should still exist.

### README updates


mmeigs (commented):
This is actually an odd one. So, the OpenAPI component in Snooty is actually only used on an odd OpenAPI preview page in `docs-landing` - not the ones you linked in `cloud-docs`. It's a head-fake, for sure. That back button is actually rendered in our redoc fork, I believe!! I'm currently [asking a channel](https://mongodb.slack.com/archives/C06S703FQPK/p1713383113840109) in Slack if we should change that back button within redoc for the `cloud-docs` OpenAPI pages.

As for the preview page... it doesn't even show a back button right now. I think something in the conditionals is broken. Could be interesting to look into for you! But in the meantime, I'm in favor of completely deleting `SidenavBackButton` as a component. Then, in OpenAPI we can replace that with the new `Docs Home` button. That makes most sense to me.

rayangler (commented):
Current change looks good to me (but will hold off on formally approving based on above discussion).

I agree with the idea of deleting the `SidenavBackButton` component and removing it or updating it from the existing `OpenAPI` component. If we update the OpenAPI preview page to use the new "Docs Home" button, we should consider making it a minimal reusable component. This page should be testable by building `docs-landing` and going to `/openapi/preview`.

I would say this isn't a blocker though and should maybe time box the effort of figuring out the rendering issue of the button on the OpenAPI preview page.

src/components/Sidenav/DocsHomeButton.js
```diff
@@ -0,0 +1,34 @@
...
+
+const homeLinkStyle = LeafyCSS`
+  span {
+    color: ${palette.gray.dark1};
+    font-weight: 400;
+    display: flex;
+    gap: 6px;
+    align-items: text-bottom;
+    svg {
+      height: 17px;
```
mayaraman19, line 19: horrible little way to ensure that the svg and text of home button slightly align, since the svg is a bit bigger than its content so visually it doesn't line up even though the text and icon are bottom-aligned.

src/components/Sidenav/DocsHomeButton.js
```diff
@@ -0,0 +1,34 @@
...
+import { sideNavItemBasePadding } from './styles/sideNavItem';
+import { titleStyle } from './styles/sideNavItem';
+
+const homeLinkStyle = LeafyCSS`
+  span {
+    color: ${palette.gray.dark1};
+    font-weight: 400;
+    display: flex;
+    gap: 6px;
+    align-items: text-bottom;
```
seungpark, line 17: getting an invalid value here for `text-bottom`. we can also remove the height on the svg and set a line-height: ${theme.size.default};

seungpark (commented):
minor css improvement below. otherwise LGTM 👍

src/components/Sidenav/DocsHomeButton.js
mayaraman19: image

removing `align-items: text-bottom`, but I think I need to keep the `height: 17px` because otherwise the icon and title are not very aligned :(

seungpark: i actually feel like its more representative of design with the `line-height: 16px;` on the span vs the height on svg but this is really minor. not a blocker

mayaraman19: i do as well, but when i sent the staging link to design they wanted it to be a bit more aligned bottom-wise

PR #1068, opened 2024-04-17
DOP-4529: Build the breadcrumb data for the current page

anabellabuckvar:
### Stories/Links:

[DOP-4529](https://jira.mongodb.org/browse/DOP-4529)

### Current Behavior:

[Atlas Device SDKs: Install the Web SDK](https://www.mongodb.com/docs/atlas/device-sdks/web/install/)
[Object Models - C++ SDK](https://www.mongodb.com/docs/atlas/device-sdks/sdk/cpp/model-data/object-models/)
[C# Driver](https://www.mongodb.com/docs/drivers/csharp/current/)

### Staging Links:
[Atlas Device SDKs: Install the Web SDK](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/realm/anabella.buckvar/DOP-4529/web/install/index.html)
[Object Models - C++ SDK](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/realm/anabella.buckvar/DOP-4529/sdk/cpp/model-data/object-models/index.html)
[C# Driver](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/csharp/anabella.buckvar/DOP-4529/)

### Notes:
Atlas App Services fetchBreadcrumbs Function

Screenshot 2024-04-24 at 11 06 41 AM

### README updates


src/components/Breadcrumbs/BreadcrumbContainer.js
```diff
@@ -39,10 +39,34 @@ const linkStyling = LeafyCss`
...
 
-const BreadcrumbContainer = ({ homeCrumb, lastCrumb }) => {
-  const { project } = useSnootyMetadata();
-  const parents = useNavigationParents(project);
-  const breadcrumbs = React.useMemo(() => [homeCrumb, ...parents, lastCrumb], [homeCrumb, parents, lastCrumb]);
+const BreadcrumbContainer = ({ homeCrumb, propertyCrumb, slug }) => {
+  const { parentPaths } = useSnootyMetadata();
+
+  //get intermediate breadcrumbs and property Url
+  const queriedCrumbs = useBreadcrumbs()[0];
```
mayaraman19, line 46: In the case that `useBreadcrumbs` returns `[]` this line is going to cause an error. Could we possibly slightly tweak the return type of `useBreadcrumbs` so we don't have to index its result here?

src/components/Breadcrumbs/index.js
```diff
@@ -20,14 +20,14 @@ const Breadcrumbs = ({ homeUrl = null, pageTitle = null, siteTitle, slug }) => {
     url: homeUrl || baseUrl(),
   };
   // If a pageTitle prop is passed, use that as the last breadcrumb instead
```
mayaraman19, line 22: just to prevent future confusion, perhaps change "last" in this comment to "property"?

tests/unit/__snapshots__/Breadcrumbs.test.js.snap
```diff
@@ -172,9 +172,51 @@ exports[`renders correctly with pageTitle 1`] = `
     </span>
     <a
       class="emotion-4 emotion-5"
-      href="/view-analyze/"
+      href="/www.mongodb.comhttps://www.mongodb.com/docs/atlas/"
```
mayaraman19, line 175: hmmm maybe something needs to be double checked with the intermediate(?) breadcrumb here? a similar pattern also shows up when clicking on the intermediate breadcrumb in staging or development.

mayaraman19 (commented):
thank you for doing this! left some minor comments, main thing is that something might need to be checked with the intermediate breadcrumb, maybe we don't need to prepend it with `www.mongodb.com`. also thank you for adding tests!! 

one minor note that is open to discussion - perhaps we should not merge `main` into `feature-breadcrumbs` PRs just to make reviews a little bit easier?

tests/unit/BreadcrumbContainer.test.js
```diff
@@ -4,77 +4,88 @@ import { render } from '@testing-library/react';
...
+  });
 
-    const tree = mountBreadcrumbContainer(mockHomeCrumb, mockLastCrumb, []);
+  it('renders correctly with intermediate breadcrumb. No slug or project parents', () => {
+    const tree = mountBreadcrumbContainer(mockHomeCrumb, mockPropertyCrumb, '');
     expect(tree.asFragment()).toMatchSnapshot();
   });
 });
+
+//can do 2 more tests here for something without intermediate breadcrumbs
```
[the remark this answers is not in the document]
anabellabuckvar: they're there, but I removed the comment, thanks!

src/components/Breadcrumbs/BreadcrumbContainer.js
anabellabuckvar: It's not a significant change but I changed the `useBreadcrumbs()` function so that the result is indexed there instead. Let me know if you'd prefer a different approach

tests/unit/__snapshots__/Breadcrumbs.test.js.snap
anabellabuckvar: Thanks for catching this! Should be fixed now

plugins/utils/breadcrumbs.js
```diff
@@ -0,0 +1,25 @@
...
+
+  return createNode({
+    children: [],
+    id: createNodeId(`Breadcrumbs-${project}`),
+    internal: {
+      contentDigest: createContentDigest(breadcrumbData?.breadcrumbs),
+      type: breadcrumbType,
+    },
+    breadcrumbs: breadcrumbData.breadcrumbs,
+    propertyUrl: breadcrumbData.propertyUrl,
```
rayangler, line 18: I think we'll want to use optional chaining on all of these in case the return data is undefined

plugins/utils/breadcrumbs.js
```diff
@@ -0,0 +1,25 @@
...
+const breadcrumbType = `Breadcrumb`;
+
+const createBreadcrumbNodes = async ({ db, createNode, createNodeId, createContentDigest, getNodesByType }) => {
+  const { database, project } = siteMetadata;
+
+  const breadcrumbData = await db.fetchBreadcrumbs(database, project);
+
+  return createNode({
+    children: [],
+    id: createNodeId(`Breadcrumbs-${project}`),
```
rayangler, line 12: Would it be useful to wrap this in a try-catch block to handle any potential errors that might arise from the network request? The [createOpenAPIChangelogNode](https://github.com/mongodb/snooty/blob/c8d29a4b88ed80bbece684bc725f6f1ceb70989c/plugins/utils/openapi.js#L42) function has a good example of this.

As an edge case, if I try to pass in a project that doesn't exist like `foo`, the site breaks for me

plugins/utils/breadcrumbs.js
```diff
@@ -0,0 +1,25 @@
+const { siteMetadata } = require('../../src/utils/site-metadata');
+
+const breadcrumbType = `Breadcrumb`;
+
+const createBreadcrumbNodes = async ({ db, createNode, createNodeId, createContentDigest, getNodesByType }) => {
```
rayangler, line 5: (Nit) `getNodesByType` doesn't seem to be used. We can probably remove this

mmeigs (commented):
Oh, also I believe now useNavigationParents and the createProjectParentNodes that creates it is not used anymore. So, we can get rid of all the instances of them!

But do NOT delete the App Services function. That is still needed by the main branch and prod. :)

plugins/utils/breadcrumbs.js
rayangler: I think the Atlas App Service function also has a similar issue where if a bad project name is passed to it, an error is thrown when trying to access its properties. Particularly in:

```js
(result) => {
  const propertyUrl = result.slug ? result.baseUrl+result.slug : result.baseUrl;
  return result ? { breadcrumbs: result.breadcrumbs ? result.breadcrumbs: [], propertyUrl:propertyUrl } : [];
}
```

`result` would be null if `project` is a bad value. 

We could probably resolve this by just returning `null` if `!result`, and this will avoid us needing to return `[]` by default

src/components/Breadcrumbs/BreadcrumbContainer.js
```diff
@@ -39,10 +39,34 @@ const linkStyling = LeafyCss`
...
+            return { ...crumb, url: propertyUrl + crumb.path };
+          })
+        : [],
+    [parentPaths, slug, propertyUrl]
+  );
+
+  const breadcrumbs = React.useMemo(
+    () => [homeCrumb, ...intermediateCrumbs, propertyCrumb, ...parents],
+    [homeCrumb, intermediateCrumbs, propertyCrumb, parents]
+  );
```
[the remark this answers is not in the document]
anabellabuckvar: From what I can tell, at the moment we don't generate any breadcrumbs for property homepages, anyways.

https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/realm/anabella.buckvar/DOP-4529/sdk/cpp/index.html

https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/realm/anabella.buckvar/DOP-4529/index.html

src/components/Breadcrumbs/BreadcrumbContainer.js
```diff
@@ -39,10 +39,34 @@ const linkStyling = LeafyCss`
...
-  const parents = useNavigationParents(project);
-  const breadcrumbs = React.useMemo(() => [homeCrumb, ...parents, lastCrumb], [homeCrumb, parents, lastCrumb]);
+const BreadcrumbContainer = ({ homeCrumb, propertyCrumb, slug }) => {
+  const { parentPaths } = useSnootyMetadata();
+
+  //get intermediate breadcrumbs and property Url
+  const queriedCrumbs = useBreadcrumbs()[0];
+  const propertyUrl = queriedCrumbs.propertyUrl;
+  const intermediateCrumbs = queriedCrumbs.breadcrumbs.map((crumb) => {
+    return { ...crumb, url: `www.mongodb.com${crumb.url}` };
```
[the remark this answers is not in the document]
anabellabuckvar: Great idea, thanks! added that in a few places

src/components/Breadcrumbs/BreadcrumbContainer.js
```diff
@@ -39,10 +42,41 @@ const linkStyling = LeafyCss`
...
+  const parents = React.useMemo(
+    () =>
+      parentPaths[slug]
+        ? parentPaths[slug].map((crumb) => {
+            return { ...crumb, url: propertyUrl + removeLeadingSlash(crumb.path) };
+          })
+        : [],
+    [parentPaths, slug, propertyUrl]
+  );
+  console.log(parents);
```
rayangler, line 74: Should we remove this log?

src/components/Breadcrumbs/BreadcrumbContainer.js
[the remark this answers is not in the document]
mmeigs: You are right here: https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/csharp/anabella.buckvar/DOP-4529/index.html

src/components/Breadcrumbs/BreadcrumbContainer.js
```diff
@@ -39,10 +42,41 @@ const linkStyling = LeafyCss`
...
+  const { parentPaths } = useSnootyMetadata();
+
+  //get intermediate breadcrumbs and property Url
+  const queriedCrumbs = useBreadcrumbs();
+  const propertyUrl = assertTrailingSlash(queriedCrumbs?.propertyUrl);
+  const intermediateCrumbs = React.useMemo(
+    () =>
+      queriedCrumbs?.breadcrumbs
+        ? queriedCrumbs.breadcrumbs.map((crumb) => {
+            return { ...crumb, url: `http://www.mongodb.com${assertLeadingSlash(crumb.url)}` };
```
rayangler, line 55: (This is not a blocker) Would we want to assert trailing slashes for every breadcrumb (both `intermediateCrumbs` and `parents`)? Not an expert on how this would affect SEO, but I believe this does technically result in a redirect for the user

rayangler (commented):
Just a couple of minor things from me.

For the Atlas App Service function, I think we'll want to return `null` early if `results` is falsey to avoid any errors when accessing object properties of non-existent results. Otherwise, you may get the error `TypeError: Cannot access member 'slug' of undefined` before ever returning `null`.

Maybe something like:

```js
(result) => {
      if (!result) return null;
      const propertyUrl = result.slug ? result.baseUrl+result.slug : result.baseUrl;
      return { breadcrumbs: result.breadcrumbs ? result.breadcrumbs: [], propertyUrl: propertyUrl };
    },
```

src/components/Breadcrumbs/BreadcrumbContainer.js
anabellabuckvar: Yes sorry that slipped back in when making the changes addressing comments

src/components/Breadcrumbs/BreadcrumbContainer.js
rayangler: To add, I realize I forgot to finish my thought in the comment:

... I believe this does technically result in a redirect for the user if it doesn't have a trailing slash*

Again, not a blocker, but just something to consider

anabellabuckvar: no clue? But happy to add it just in case if you think that would be helpful?

rayangler: Sure, let's do it

plugins/utils/breadcrumbs.js
```diff
@@ -0,0 +1,43 @@
...
+      id: createNodeId(`Breadcrumbs-${project}`),
+      internal: {
+        contentDigest: createContentDigest({}),
+        type: breadcrumbType,
+      },
+      breadcrumbs: null,
+      propertyUrl: '',
+    });
+  }
+};
```
mayaraman19, line 38: minor nitpick but to reduce duplication here you could set `contentDigest`, `breadcrumbs`, `propertyUrl` using the conditional (e.g. `const breadcrumbs = breadcrumbData ? breadcrumbData.breadcrumbs : {}` bla bla bla u get the gist) and then create the node.

rayangler (approved):
I think things generally LGTM. I'd wait to merge until it looks good to everyone else as well!

src/components/Breadcrumbs/BreadcrumbContainer.js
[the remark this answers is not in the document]
mmeigs: @anabellabuckvar just wondering if you've looked into this

mayaraman19 (approved):
i think it looks good, just a reminder to address matt's comment about landing pages

src/components/Breadcrumbs/BreadcrumbContainer.js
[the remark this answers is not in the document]
anabellabuckvar: @mayaraman19 the C# one is not supposed to, that's happening because some of the drivers have incorrect breadcrumbs data in Atlas

plugins/utils/breadcrumbs.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+  const [breadcrumbs, propertyUrl] = breadcrumbData
+    ? [breadcrumbData.breadcrumbs, breadcrumbData.propertyUrl]
+    : [null, ''];
+
+  return createNode({
+    children: [],
+    id: createNodeId(`Breadcrumbs-${project}`),
+    internal: {
+      contentDigest: createContentDigest(breadcrumbData.breadcrumbs),
```
mmeigs, line 21: Make sure to change this `breadcrumbData.breadcrumbs` to the new safely defaulted `breadcrumbs`

src/components/Breadcrumbs/BreadcrumbContainer.js
```diff
@@ -39,10 +42,42 @@ const linkStyling = LeafyCss`
...
+  //get intermediate breadcrumbs and property Url
+  const queriedCrumbs = useBreadcrumbs();
+  const propertyUrl = assertTrailingSlash(queriedCrumbs?.propertyUrl);
+  const intermediateCrumbs = React.useMemo(
+    () =>
+      queriedCrumbs?.breadcrumbs
+        ? queriedCrumbs.breadcrumbs.map((crumb) => {
+            return { ...crumb, url: assertTrailingSlash(`http://www.mongodb.com${assertLeadingSlash(crumb.url)}`) };
+          })
+        : [],
```
mmeigs, line 57: nit: but for clarity/readability: instead of using a ternary operator, could we do 

```
React.useMemo(() => (queriedCrumbs?.breadcrumbs ?? []).map((.....
```

This makes it one statement with a kind of fallback rather than having to follow the logic of the ternary.

src/components/Breadcrumbs/BreadcrumbContainer.js
```diff
@@ -39,10 +42,42 @@ const linkStyling = LeafyCss`
...
+    [queriedCrumbs]
+  );
+
+  if (propertyCrumb) {
+    propertyCrumb.url = propertyUrl;
+  }
+
```
mmeigs, line 64: Could we leave a comment to explain this, please :)

src/components/Breadcrumbs/BreadcrumbContainer.js
```diff
@@ -39,10 +42,42 @@ const linkStyling = LeafyCss`
...
+  //add respective url to each direct parent crumb
+  const parents = React.useMemo(
+    () =>
+      parentPaths[slug]
+        ? parentPaths[slug].map((crumb) => {
+            return { ...crumb, url: assertTrailingSlash(propertyUrl + removeLeadingSlash(crumb.path)) };
+          })
+        : [],
+    [parentPaths, slug, propertyUrl]
+  );
```
mmeigs, line 75: nit: you could do the same thing with this as above `(parentPaths[slug] ?? []).map....`

src/components/Breadcrumbs/BreadcrumbContainer.js
mmeigs: Oh, and I don't think we need the conditional? Isn't `propertyCrumb` always defined?

mmeigs (commented):
A few comments! And although I'm totally fine with not doing it until your next ticket, I thought we were going to put the logic of the breadcrumb construction into a util. Fine to leave out for now though!

src/components/Breadcrumbs/BreadcrumbContainer.js
anabellabuckvar: Yup I added a comment and changed the conditional to check for `propertyCrumb.length`; `propertyCrumb` is set to an empty array when the current site is the property homepage. It's a little counterintuitive to have this set in both files but I was thinking I would finish cleaning up the logic in the next PR because that's where I already separated the breadcrumb creation logic into a different util

mmeigs (approved):

PR #1072, opened 2024-04-23
DOP-4530: Add IndividualBreadcrumb and CollapsedBreadcrumb components

i80and:
### Stories/Links:

DOP-4539

### Notes:

* `@leafygreen/tooltip` package updated to support the `baseFontSize` option.

### README updates


src/components/Breadcrumbs/IndividualBreadcrumb.js
```diff
@@ -0,0 +1,85 @@
...
+
+  :hover,
+  :focus {
+    text-decoration: none;
+  }
+`;
+
+// This should be graphemes using the Intl.Segmenter API, but browser support is too new
+// at time of writing. For now, this is in code points. Which is wrong. But fast and easy.
+const MAX_CRUMB_LENGTH = 21;
```
seungpark, line 36: the discussion on the figma was truncating to 21 characters but i feel like this wouldn't be consistent enough  (either overly compensative and short, or bleed to second line. as far as I understand, we do not want this to wrap)

my suggestion is to try the whole BreadCrumbContainer as a flexbox and have each Breadcrumb in the middle have these properties:

```
    text-overflow: ellipsis;
    overflow: hidden;
    text-wrap: nowrap;
```

the first and last should have `min-width: max-content;` to not truncate. 

going from
image


to

image

*note: for some reason the last link had to be adjusted `line-height` property to stay in line with the rest of the content

seungpark (commented):
some comments below! i think its pretty important to keep it to one line so had a style suggestion vs a truncating logic.

also tested spreading the `projectParents` in the [graph node creation](https://github.com/mongodb/snooty/blob/main/plugins/utils/project-parents.js#L21) to produce the screenshots in comments

src/components/Breadcrumbs/IndividualBreadcrumb.js
```diff
@@ -0,0 +1,85 @@
...
+  );
+
+  if (truncated) {
+    result = (
+      <Tooltip
+        // To get the tooltip above the topnav we have to pop up the z-index
+        popoverZIndex={9001}
+        baseFontSize={13}
+        triggerEvent="hover"
+        open={true}
```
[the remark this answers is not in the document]
i80and: Ah, crap, yeah, this was from testing. Thank you!

i80and:
image
image
image

src/components/Breadcrumbs/BreadcrumbContainer.js
```diff
@@ -1,69 +1,53 @@
...
 const BreadcrumbContainer = ({ homeCrumb, lastCrumb }) => {
   const { project } = useSnootyMetadata();
   const parents = useNavigationParents(project);
-  const breadcrumbs = React.useMemo(() => [homeCrumb, ...parents, lastCrumb], [homeCrumb, parents, lastCrumb]);
+
+  // Our breadcrumbs representation is an array of crumbObjectShape || (array of crumbObjectShape)
+  // The latter indicates a collapsed series of breadcrumbs.
+  const breadcrumbs = React.useMemo(() => {
+    const crumbs = [homeCrumb, ...parents, lastCrumb];
+    if (crumbs.length > 5) {
```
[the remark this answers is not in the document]
i80and: The PR is now permitting 3 or 5 breadcrumbs, depending on `(useScreenSize).isMobile()` as shown in new screenshots!

src/components/Breadcrumbs/IndividualBreadcrumb.js
i80and: Done! The race condition I alluded to doesn't seem to be happening today? And I THINK the result is better?

src/components/Breadcrumbs/BreadcrumbContainer.js
```diff
@@ -1,73 +1,67 @@
...
+  const breadcrumbs = React.useMemo(() => {
+    const crumbs = [homeCrumb, ...parents, lastCrumb];
+    if (crumbs.length > maxCrumbs) {
+      // A maximum of maxCrumbs breadcrumbs may be shown, so we collapse the first run of internal
+      // crumbs into a single "…" crumb
+      const collapsedCrumbs = crumbs.splice(1, crumbs.length - maxCrumbs + 1, []);
+      crumbs[1] = collapsedCrumbs;
+    }
+    return crumbs;
+  }, [maxCrumbs, homeCrumb, parents, lastCrumb]);
```
mmeigs, line 40: Oh ok, I think we're going to need to sync as a group to reformulate how we do this. There's a lot more complexity to this that needs to occur.

mmeigs: Basically, as everything gets too truncated, one-by-one they should be subsumed into the CollapsedBreadcrumbs as seen in the figma. 

I think this will require passing down a function or ref to the IndividualBreadcrumb component which will trigger whenever one of the breadcrumbs has been truncated to less than 21 characters width. (this will take a calculation similar to what you're doing already. 

ie `(ref.current?.scrollWidth ?? 0) > (ref.current?.clientWidth ?? 0) && (ref.current?.clientWidth <= 125px)`

The 125px is my quick eye on the width that contains about 21 characters.

mmeigs (commented):
This looks really good overall. Some small things. But mainly we need to fix the way the truncation happens from a larger perspective, sadly. It is the most annoying part of this.

As soon as a breadcrumb is truncated AND smaller than 21 characters, the CollapsedBreadcrumb ellipses should subsume its closest breadcrumb to give the rest more space. There aren't going to be exact numbers on maxCrumb. More of a calculation.

src/components/Breadcrumbs/IndividualBreadcrumb.js
```diff
@@ -0,0 +1,92 @@
...
+`;
+
+const linkWrapperLayoutStyling = LeafyCss`
+  text-overflow: ellipsis;
+  overflow: hidden;
+  text-wrap: nowrap;
+
+  :first-child, :last-child {
+    min-width: max-content;
+  }
```
mmeigs, line 25: This also *can* rarely cause an issue when on mobile and the last crumb is long it cuts out without an ellipsis. We can fix this at least on mobile (should be the only time this happens) by not giving this property to the last-child.

src/components/Breadcrumbs/IndividualBreadcrumb.js
```diff
@@ -31,35 +15,58 @@ const linkStyling = LeafyCss`
...
+    </span>
   );
 
-  if (truncated) {
+  if (isTruncated) {
     result = (
       <Tooltip
         // To get the tooltip above the topnav we have to pop up the z-index
         popoverZIndex={9001}
         baseFontSize={13}
```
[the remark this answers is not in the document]
i80and: Neat!

The only values the `Tooltip` component accepts are 13 and 16 -- would this potentially be brittle? Or is it not a concern?

src/components/Breadcrumbs/BreadcrumbContainer.js
```diff
@@ -1,73 +1,67 @@
...
-
-const StyledArrow = styled('span')`
+const StyledSlash = styled('span')`
   cursor: default;
-
-  :last-of-type {
-    ${activeColor}
-  }
+  padding-left: 4px;
+  padding-right: 4px;
```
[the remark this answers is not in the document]
i80and: Ah! I thought it was 4px on each side. My error!

src/components/Breadcrumbs/IndividualBreadcrumb.js
```diff
@@ -0,0 +1,92 @@
...
+import Link from '../Link';
+import { formatText } from '../../utils/format-text';
+import { theme } from '../../theme/docsTheme';
+
+const linkStyling = LeafyCss`
+  font-size: ${theme.fontSize.small};
+
+  :hover,
+  :focus {
+    text-decoration: none;
```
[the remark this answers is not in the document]
i80and: Good eye 👀 thanks!

src/components/Breadcrumbs/BreadcrumbContainer.js
i80and: Can this happen in a follow-up ticket, given that this is going out on Thursday?

i80and: Also, would this introduce a flash of layout changes? To get the ref's layout width, we would have to do the layout, then belatedly make a decision on whether to collapse breadcrumbs after the nodes have already been out in the wilds of the DOM

src/components/Breadcrumbs/BreadcrumbContainer.js
```diff
@@ -1,73 +1,67 @@
...
-  :focus {
-    text-decoration: none;
-
-    :not(:last-of-type) {
-      ${activeColor}
-    }
-  }
+const Flexbox = styled('span')`
+  display: flex;
 `;
```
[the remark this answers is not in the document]
mmeigs: Gotcha! I honestly don't understand why it was a <Body> component as the parent in the first place. We can change that to a div! This is a collection of different elements, and I don't agree the choice from the past to use a p tag when this is more of a section of many elements than a block of mostly text.

src/components/Breadcrumbs/BreadcrumbContainer.js
mmeigs: It would end up producing a layout shift. We can ask design if they'd like a loading skeleton that might be preferable.

I am nervous to push that off. This is the base of what was asked for the project.

mmeigs: Because this will happen.
Screenshot 2024-04-30 at 3 25 40 PM

src/components/Breadcrumbs/BreadcrumbContainer.js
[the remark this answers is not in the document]
i80and: Cool! I'm on board with using a div instead of a the LG typography `<Body>` component then. It seems to look fine.

src/components/Breadcrumbs/BreadcrumbContainer.js
i80and: Got it. I'll try to make this change happen tonight then

src/components/Breadcrumbs/IndividualBreadcrumb.js
i80and: Addressed! :last-child is now in a @media query

i80and:
The exact behavior we want is a fun little optimization problem without well-defined parameters. The behavior of this PR as it now stands using the heuristic Matt suggested is a greedy implementation: collapsing breadcrumbs is mostly always chosen if it's available as an option over breadcrumb truncation. But truncation does happen.

Behavior with a synthetic dataset shown.

[Screencast_20240430_234634.webm](https://github.com/mongodb/snooty/assets/334192/fe9bf879-08e0-4f52-a9bf-21bab75a7c86)

mmeigs (approved):
Looks fantastic!! Thank you! 
Now, the fun part of merging in the state from the feature branch 😅

PR #1078, opened 2024-04-26
DOP-4540: Simplify versioned prefix generation regardless of build type

rayangler:
### Stories/Links:

DOP-NNNN

### Current Behavior:

### Staging Links:

### README updates

### DIFF
--- a/src/context/version-context.js
+++ b/src/context/version-context.js
@@ -241,7 +241,7 @@ const VersionContextProvider = ({ repoBranches, slug, children }) => {
         gitBranchName === LEGACY_GIT_BRANCH
           ? gitBranchName
           : targetBranch.urlSlug || targetBranch.urlAliases[0] || targetBranch.gitBranchName;
-      const urlTarget = getUrl(target, metadata.project, metadata, repoBranches?.siteBasePrefix, slug);
+      const urlTarget = getUrl(target, metadata.project, repoBranches?.siteBasePrefix, slug);
       navigate(urlTarget);
     },
     [availableVersions, metadata, repoBranches, slug]
--- a/tests/unit/Head.test.js
+++ b/tests/unit/Head.test.js
@@ -3,10 +3,10 @@ import { render, screen } from '@testing-library/react';
 import { Head } from '../../src/components/DocumentBody';
 import mockStaticQuery from '../utils/mockStaticQuery';
 import { useSiteMetadata } from '../../src/hooks/use-site-metadata';
-import { generatePrefix } from '../../src/components/VersionDropdown/utils';
 import useSnootyMetadata from '../../src/utils/use-snooty-metadata';
 import { AVAILABLE_LANGUAGES } from '../../src/utils/locale';
 import { DOTCOM_BASE_URL } from '../../src/utils/base-url';
+import { generateVersionedPrefix } from '../../src/utils/generate-versioned-prefix';
 import mockCompleteEOLPageContext from './data/CompleteEOLPageContext.json';
 import mockEOLSnootyMetadata from './data/EOLSnootyMetadata.json';
 import mockHeadPageContext from './data/HeadPageContext.test.json';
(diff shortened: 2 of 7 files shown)


src/context/version-context.js
```diff
@@ -241,7 +241,7 @@ const VersionContextProvider = ({ repoBranches, slug, children }) => {
         gitBranchName === LEGACY_GIT_BRANCH
           ? gitBranchName
           : targetBranch.urlSlug || targetBranch.urlAliases[0] || targetBranch.gitBranchName;
-      const urlTarget = getUrl(target, metadata.project, metadata, repoBranches?.siteBasePrefix, slug);
+      const urlTarget = getUrl(target, metadata.project, repoBranches?.siteBasePrefix, slug);
```
seungpark, line 244: seems like this wont get the correct url in dev/staging. is this intended to be targeted to the same env+project?

ie from staging site with dev env link https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/default-dev/master/docs/raymund.rodriguez/DOP-4540/ goes to `/docs/v6.0` when changing the version selector

tests/unit/Head.test.js
```diff
@@ -62,23 +66,91 @@ describe('Head', () => {
...
-      const siteMetadata = useSiteMetadata();
-      const { siteUrl, parserBranch } = siteMetadata;
-      const urlSlug = mockHeadPageContext.repoBranches.branches.find(
-        (branch) => branch.gitBranchName === parserBranch
-      )?.urlSlug;
-      const pathPrefix = generatePrefix(urlSlug, siteMetadata);
-      const slug = mockHeadPageContext.slug;
 
-      const canonical = `${siteUrl}${pathPrefix}/${slug === '/' ? '' : slug}`;
+      const expectedCanonical = 'https://www.mongodb.com/docs/mongocli/undefined/';
```
seungpark, line 110: should this be the expected behavior here? ie. how should test doc branches' (why named parserBranch 🤯) canonicals be handled?

seungpark (commented):
question on changing versions on staged versions as well as test doc branches below! otherwise very clean! thanks for the clean up

src/context/version-context.js
rayangler: That's if you click the `v6.0` version, right? Do you have an expectation of where it would go for dev? `/docs-qa/v6.0`?

seungpark: i was imagining it to be an extension of the current base url and replacing the version, ie `https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/default-dev/v6.0/docs/raymund.rodriguez/DOP-4540/` but open to suggestions here

tests/unit/Head.test.js
seungpark: i would think to just get the parserBranch info and tag it on. we can log an error for missing branch, but the canonical should be still the "valid" right?

src/context/version-context.js
rayangler: I thought we originally wanted to simplify the url generation by not having to worry about different envs and varying code paths. Wouldn't something like this lead us back to the original confusion in code paths that warranted the ticket?

seungpark: ha i guess i was imagining the most ideal situation - no diff in code implementation but the target is still correct. but i think this is still a regression from previous implementation, see [example here](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/matt.meigs/DOP-4548-cta-banner/aggregation/index.html). this still replaces the the correct portion of url (master -> new version) when changing the version selector vs replacing the whole after domain.

rayangler: The example linked was built with `SNOOTY_ENV === dotcomprd`, so the ideal behavior has less to do with the environment, and more to do with whether a `pathPrefix` exists in the build or not.

So just to confirm before I make changes:

We want this function to rely on the existence of whether there's a path prefix defined or not. If there's a path prefix, we just replace the version slug with the target. If there's no path prefix, we replace the parser version. This will occur regardless of what env is used to build the site. Same goes with the canonical since the function is used for both

rayangler: For posterity: Spoke about this with @seungpark offline. To help simplify logic regarding path prefixes agnostic of environment, we may need to make some local Makefile changes to have the slug order conform closer to prod deployments. Will look into this a bit

PR #1080, opened 2024-04-29
DOP-4566: set init values for tabs and tab selectors during build time

seungpark:
### Stories/Links:

DOP-4566

### Current Behavior:

[cloud docs page with tab selectors](https://www.mongodb.com/docs/atlas/atlas-search/tutorial/facet-tutorial/)

### Staging Links:

[cloud docs page](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/seung.park/DOP-4566/atlas-search/tutorial/facet-tutorial/index.html) with both `.. tabs-selector` and `.. tabs` directives

### Notes:
- This PR should not have a change in user experience - the values are being set in context during build time, but content is not rendered until local storage is read, to prevent CLS and content shifting
- PR also removes the need for useEffect that corrects bad/outdated local storage data - this quality check is moved to the init function.

### README updates


src/components/Tabs/index.js
```diff
@@ -89,27 +89,24 @@ const Tabs = ({ nodeData: { children, options = {} }, page, ...rest }) => {
...
+    return activeTabIdx;
+  });
 
   const scrollAnchorRef = useRef();
-  const previousTabsetChoice = activeTabs[tabsetName];
   // Hide tabset if it includes the :hidden: option, or if it is controlled by a dropdown selector
   const isHidden = options.hidden || Object.keys(selectors).includes(tabsetName);
   const isProductLanding = page?.options?.template === 'product-landing';
 
-  useEffect(() => {
```
seungpark, line 100: remove this effect, took the same logic and moved it to initial value function above (vs having it as a post render effect)

src/components/Tabs/tab-context.js
```diff
@@ -64,8 +72,54 @@ const reducer = (prevState, { name, value }) => {
   };
 };
 
+const createInitialTabs = (choicesPerSelector, localActiveTabs) => {
```
seungpark, line 75: added this get-initial-value function as it uses a combination of default values (ie. specifically for language tab selectors) and local storage when available in the browser

src/components/RootProvider.js
```diff
@@ -7,7 +7,8 @@ import { SidenavContextProvider } from './Sidenav';
 import { TabProvider } from './Tabs/tab-context';
 import { ContentsProvider } from './Contents/contents-context';
 
-const RootProvider = ({ children, headingNodes, selectors, slug, repoBranches, remoteMetadata, project }) => {
+const RootProvider = ({ children, headingNodes, selectors, slug, repoBranches, remoteMetadata }) => {
+  console.log('check slug', slug);
```
seungpark, line 11: TODO: remove logs before merge

src/components/Tabs/index.js
```diff
@@ -89,27 +89,24 @@ const Tabs = ({ nodeData: { children, options = {} }, page, ...rest }) => {
   const { activeTabs, selectors, setActiveTab } = useContext(TabContext);
   const tabIds = children.map((child) => getTabId(child));
   const tabsetName = options.tabset || generateAnonymousTabsetName(tabIds);
-  const [activeTab, setActiveTabIndex] = useState(0);
+  const [activeTab, setActiveTabIndex] = useState(() => {
+    const activeTabIdx = tabIds.indexOf(activeTabs?.[tabsetName]);
+    console.log(
```
seungpark, line 94: TODO: remove logs before merge

src/components/Tabs/tab-context.js
```diff
@@ -64,8 +73,60 @@ const reducer = (prevState, { name, value }) => {
...
+    return {};
+  }
+  // get default tabs based on availability
+  const defaultRes = Object.keys(choicesPerSelector || {}).reduce((res, selectorKey) => {
+    const nodeOptionIdx = choicesPerSelector[selectorKey].findIndex((tab) => tab.value === 'nodejs');
+    // TODO: find default choicesPerSelector for other tab types
+    if (selectorKey === 'drivers' && nodeOptionIdx > -1) {
+      res[selectorKey] = 'nodejs';
+    } else {
+      res[selectorKey] = choicesPerSelector[selectorKey][0].value;
```
rayangler, line 90: (In incognito) Looks like the selector will not have a selected choice when going from this [staging page](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/seung.park/DOP-4566/atlas-search/tutorial/cross-collection-tutorials/index.html) to the one above it in the TOC. Accessing [the page](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/seung.park/DOP-4566/atlas-search/tutorial/partial-match/index.html) directly works as expected though. 

Is this an expected change in behavior?

src/components/Tabs/index.js
mmeigs: Would love this to stick around as a comment, though!

src/components/Tabs/tab-context.js
```diff
@@ -64,8 +73,60 @@ const reducer = (prevState, { name, value }) => {
...
+  // all tabbed content is read from browser local storage
+  // if there is no browser, wait for client side local storage
+  // hidden content should be handled from tab components
+  if (!isBrowser) {
+    return {};
+  }
+  // get default tabs based on availability
+  const defaultRes = Object.keys(choicesPerSelector || {}).reduce((res, selectorKey) => {
+    const nodeOptionIdx = choicesPerSelector[selectorKey].findIndex((tab) => tab.value === 'nodejs');
+    // TODO: find default choicesPerSelector for other tab types
```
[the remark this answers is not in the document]
seungpark: i just left it as a comment but glad to remove. was verifying if there was any other `default tabs` we should manage

src/components/Tabs/tab-context.js
seungpark: thanks for catching the page navigation! seems like we need to keep the default check from tab component (vs tab context) as the context only inits once within each docs site.

also seems [this](https://github.com/mongodb/snooty/pull/1080/files#r1583583391) needs to be re-inserted as i am getting an error for `Cannot update a component (`TabProvider`) while rendering a different component (`Tabs`)` if i put it in the init function vs an effect post render..

rayangler: If we want to keep a context pattern, we can consider moving the tab context to a be at a page-level rather than a root-level. Might be more intuitive since tabs aren't always expected to persist across pages anyways.

This is just an idea and I'm not necessarily strongly for or against it

seungpark: done with [this commit ](https://github.com/mongodb/snooty/pull/1080/commits/b64457c7e8baad0897210e29fdc2e9f5a053896f)

great catch again!

src/components/Tabs/index.js
seungpark: added! <3

src/components/DocumentBody.js
```diff
@@ -103,28 +104,30 @@ const DocumentBody = (props) => {
...
-                  {pageNodes.map((child, index) => (
-                    <ComponentFactory key={index} metadata={metadata} nodeData={child} page={page} slug={slug} />
-                  ))}
-                </Template>
-              </div>
-            </FootnoteContext.Provider>
-          </ImageContextProvider>
-        </Widgets>
-      </InstruqtProvider>
+      <TabProvider selectors={page?.options?.selectors}>
```
rayangler, line 107: Will we need to move this to `DocumentBodyPreview` as well for parity?

src/components/Tabs/tab-context.js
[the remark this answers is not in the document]
rayangler: Gotcha. I'm not aware of any other defaults we need to be aware of at this time. If we want to leave the comment to make it clear we're overriding defaults, I would suggest just removing the "TODO" part as it suggests there's something incomplete about this.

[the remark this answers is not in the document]
seungpark: fair point! removing TODO and converting it to a NOTE

src/components/DocumentBody.js
seungpark: you are a savant <3 done!

src/layouts/index.js
```diff
@@ -103,7 +103,6 @@ const DefaultLayout = ({ children, data: { page }, pageContext: { slug, repoBran
         slug={slug}
         repoBranches={repoBranches}
         headingNodes={page?.ast?.options?.headings}
-        selectors={page?.ast?.options?.selectors}
```
rayangler, line 106: (Nit) I should've caught this in a previous review, but if it's not needed, let's remove it from [preview-layout](https://github.com/mongodb/snooty/blob/fef90df72e16f5063983f2da8e0fde99f6ab3f28/src/layouts/preview-layout.js#L156) as well.

rayangler (approved):
I think changes more or less lgtm

src/layouts/index.js
seungpark: done!

PR #1081, opened 2024-04-30
:bug: DOP-4572 Unused objects/props are included in reference links

caesarbell:
### Stories/Links:

DOP-4572

### Current Behavior:

[Current Behavior](https://www.mongodb.com/docs/manual/query-api/)

### Staging Links:

[New Behavior](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/7e656a70bc3127a7327a0223343def3e99e3af38/master/docs/caesarbell/DOP-4572/query-api/index.html)

### Notes:

This PR fixes the **Unused objects/props are included in reference links** issue by providing a validation filter for props that will become attached to an HTML element, in this case, it filters out the invalid props before being spread to GatsbyLink or LGLink which in turn passes those props into an anchor tag i.e. `<a></a>`

### Resources:
- [Reach Router Link API](https://reach.tech/router/api/Link)
- [Gatsby Link API](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-link/)
- [Global attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes)
- [Anchor attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attributes)

### README updates

### DIFF
--- a/src/components/Link.js
+++ b/src/components/Link.js
@@ -9,6 +9,7 @@ import ArrowRightIcon from '@leafygreen-ui/icon/dist/ArrowRight';
 import { isRelativeUrl } from '../utils/is-relative-url';
 import { joinClassNames } from '../utils/join-class-names';
 import { isGatsbyPreview } from '../utils/is-gatsby-preview';
+import { validateHTMAttributes } from '../utils/validate-element-attributes';
 
 /*
  * Note: This component is not suitable for internal page navigation:
@@ -66,6 +67,8 @@ const Link = ({
   if (!to) to = '';
   const anchor = to.startsWith('#');
 
+  const anchorProps = validateHTMAttributes('anchor', other);
+
   //used instead of LG showLinkArrow prop for consistency between LGLinks and GatsbyLinks(GatsbyLinks don't have that prop)
   const decoration = showLinkArrow ? (
     <span>
@@ -103,7 +106,7 @@ const Link = ({
         activeClassName={activeClassName}
         partiallyActive={partiallyActive}
         to={to}
-        {...other}
+        {...anchorProps}
       >
         {children}
         {decoration}
@@ -122,7 +125,7 @@ const Link = ({
       href={to}
       hideExternalIcon={!showExtIcon}
       target={target}
-      {...other}
+      {...anchorProps}
     >
       {children}
       {decoration}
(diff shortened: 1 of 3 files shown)


src/components/Link.js
```diff
@@ -66,6 +67,8 @@ const Link = ({
   if (!to) to = '';
   const anchor = to.startsWith('#');
 
+  const anchorProps = validateHTMAttributes('anchor', other);
```
rayangler, line 70: (Non-blocking; just forward-looking) I think it would benefit us to determine what props are actually included in `other` and determine if it's worth spreading and keeping around. With our growing catalogue of components, it would be good to be more explicit to avoid needing to guess.

That said, I think the current solution works fine for now.

rayangler (approved):
Thanks for working on this. Just one nit and one non-blocking thought

src/utils/validate-element-attributes.js
```diff
@@ -0,0 +1,167 @@
...
+];
+
+export const elements = {
+  anchor: anchorTagAttributes,
+};
+
+/**
+ * Filter props based on validAttributes and regexPatterns
+ * @param {string} elementType
+ * @param {any} props // potential suitor for HTML attributes
```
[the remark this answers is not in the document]
caesarbell: @rayangler  good call on that. I will make that change.

src/components/Link.js
caesarbell: That is a good idea. This would be a great approach as our catalog of components continues to grow.

src/utils/validate-element-attributes.js
```diff
@@ -0,0 +1,167 @@
...
+  'onwaiting',
+];
+
+const globalHTMLAttributes = [
+  'accesskey',
+  'autocapitalize',
+  'autofocus',
+  'classname', // Handle React cases
+  'contenteditable',
+  'data-*', // Any attribute that starts with 'data-' is a global attribute
```
seungpark, line 72: minor. seems like this is handled in `dataAndRoleBasedAttributes`

seungpark (approved):
minor comment on using a `*` in string[] match, but otherwise LGTM!

PR #1093, opened 2024-05-09
DOP-4594: Breadcrumbs and Internal Navigation buttons use TOC labels

mmeigs:
### Stories/Links:

DOP-4594

### Current Behavior:

[Atlas](https://www.mongodb.com/docs/atlas/tutorial/create-atlas-account/)
- if you look at the bottom of the page for the "back button", the label now in prod is Getting Started with Atlas, rather than just Getting Started

### Staging Links:

[Atlas](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/matt.meigs/DOP-4594-slug-breadcrumblabel/tutorial/create-atlas-account/index.html)
- You can now see that the breadcrumb is Getting Started and the back button at the bottom of the page is as well!
- NOTE: the TOC is still incorrect lol I dropped a ticket in for that. You'll see that the TOC says "Getting Started with Atlas" but in prod it is "Getting Started".

### Notes

This depends on a parser change which is merged yet not released!

### README updates


src/utils/setup/transform-breadcrumbs.js
```diff
@@ -1,13 +1,11 @@
...
+const transformBreadcrumbs = (breadcrumbs, slugToBreadcrumbLabel) => {
   Object.entries(breadcrumbs).forEach(([slug, breadcrumbList]) => {
     breadcrumbs[slug] = breadcrumbList.map((path) => {
-      const title = slugToTitle?.[path] || [];
+      const title = slugToBreadcrumbLabel?.[path] || '';
       return {
         path,
         title,
-        plaintext: getPlaintext(title),
+        plaintext: title,
```
seungpark, line 8: do we need both `title` and `plaintext` fields here? seems like we could do away with plaintext if we are using the same string for both

seungpark (commented):
minor point to not double title and plaintext if they are both strings

seungpark (approved):
this LGTM 👍

PR #1101, opened 2024-05-29
DOP-4660: Applies dark mode to the Vertical Procedures and Steps Component

caesarbell:
### Stories/Links:

DOP-4660

### Current Behavior:

[Atlas: Get Started](https://www.mongodb.com/docs/atlas/getting-started/)
Click on the  Atlas UI tab and verify the Vertical Procedures Component

[Atlas: Connect to your Atlas Cluster](https://www.mongodb.com/docs/atlas/tutorial/connect-to-your-cluster/#connect-to-your-atlas-cluster)
Verify the Steps Component

### Staging Links:

[Staged Atlas: Get Started](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/caesarbell/DOP-4660/getting-started/index.html)
Click on the  Atlas UI tab and verify the Vertical Procedures Component

[Staged Altas: Connect to your Atlas Cluster](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/caesarbell/DOP-4660/tutorial/connect-to-your-cluster/index.html#connect-to-your-atlas-cluster)
Verify the Steps Component

### Notes:

- Dark mode is enabled for both the Vertical Procedures Component and the Steps Component.

### README updates


mmeigs (commented):
This looks pretty good. 
I do think we all should probably decide on our preferred approach.

Do we use the context in DocumentBody and pass the prop through everything? Or should we use the context in each (parent) component and only pass down if there are children that need it (meaning, not children that pass through the ComponentFactory)?

Or some other pattern. Pros and cons, etc.

caesarbell:
> This looks pretty good. I do think we all should probably decide on our preferred approach.
> 
> Do we use the context in DocumentBody and pass the prop through everything? Or should we use the context in each (parent) component and only pass down if there are children that need it (meaning, not children that pass through the ComponentFactory)?
> 
> Or some other pattern. Pros and cons, etc.

@mmeigs this is a good point. I think you are correct, this is probably a good spot to have that discussion. 

In one case we call the context in the top level and pass it down to the ComponentFactory Component and access the boolean value for the components we need, and in those cases when there are children that do not need it, it will just have a false value by default with no objects and invalid DOM attributes rendering in the browser. 

The alternative would be calling the context in all the parent components that will need it including their children, and when thinking about Typography, I feel this could be a bit overkill (CMIIW), but I am curious what you think @mmeigs and other members of the team.

caesarbell:
Another example that might shine some light is the Header Component. This also calls the `ComponentFactory`, and if we remove the `darkMode` props from the `....rest` object, it won't be passed down into the `ComponentFactory`  which has `darkMode` set to false.

caesarbell:
I have updated this PR, I believe it makes more sense to leverage the `useDarkMode` hook which is provided by LeafyGreen. Which also aligns with @mmeigs concerns. This allows us to select the components that need overriding styles from what LG gives us, or for Snooty Components that don't use LG.

mmeigs (approved):

PR #1110, opened 2024-06-04
DOP-4627: Darkmode - Code components

mmeigs:
### Stories/Links:

DOP-4627

### Current Behavior:

[Page with Code & caption block](https://www.mongodb.com/docs/atlas/api/migrate-to-new-version/)
[Page with Code and also Input/Output Codeblock](https://www.mongodb.com/docs/atlas/atlas-search/compound/)

### Staging Links:

[Page with Code & caption block](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/matt.meigs/DOP-4627-darkmode-code/api/migrate-to-new-version/index.html)
[Page with Code and also Input/Output Codeblock](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/matt.meigs/DOP-4627-darkmode-code/atlas-search/compound/index.html)

### Notes:

Updated the Code package version to the latest release.
Changes Snooty styles to match desired dark mode state.

This affects the `code-block` and `io-code-block` directives.

### README updates


src/components/Code/Code.js
```diff
@@ -29,10 +22,11 @@ const sourceCodeStyle = css`
 
 const Code = ({
   nodeData: { caption, copyable, emphasize_lines: emphasizeLines, lang, linenos, value, source, lineno_start },
-  darkMode,
+  darkMode: localDarkMode,
```
mayaraman19, line 25: looks pretty good - out of curiosity, why do we need the `darkMode` prop since the LG Code component seems to [automatically detect dark mode](https://github.com/mongodb/leafygreen-ui/blob/main/packages/code/src/Code/Code.tsx#L80)? Same with [IconButton](https://github.com/mongodb/leafygreen-ui/blob/af2082609d1e049b50adc8c43a4b30ebf678c4b8/packages/icon-button/src/IconButton.tsx#L37), unless there is something I am misunderstanding!

mmeigs: Ah, I guess I can update IconButton in this PR to get it to auto detect. I've been on the fence about how much to separate all these tickets between components, but I think you're right.

The Code prop... this is an odd one. This was already in our code. I have not seen an instance yet where this would be used... but I am worried to take it out because I feel like I must be unaware of a way that it's used **while in light mode** in our docs currently. So, I've kept the use of it as before. Basically, the logic is that we will use darkMode Code if this prop is passed in OR if the user's preference is set to darkmode.

src/components/Code/Output.js
```diff
@@ -20,7 +22,16 @@ const Output = ({ nodeData: { children }, ...rest }) => {
...
+
+        /* Ensures no double border */
+        ${darkMode &&
+        `
+          > div {
+            border-top: var(--code-border-top);
+          }
+        `}
       `}
+      style={{ '--code-border-top': 'none' }}
```
mayaraman19, line 34: might be a slight bit redundant to have `darkMode` and the CSS variable here? could we maybe either do 

```
${darkMode &&
        `
          > div {
            border-top: none;
          }
        `}
```
or
```
/* Ensures no double border */
        `
          > div {
            border-top: var(--code-border-top);
          }
        `}
      `}
      style={{ '--code-border-top': darkMode ? 'none' : 'auto' }}
```
or something similar?

mmeigs: oh damn, that second option is what I meant to do but I was going too quickly. Thank you!!

src/components/Code/Code.js
```diff
@@ -101,7 +100,7 @@ const Code = ({
         highlightLines={emphasizeLines}
         language={language}
         languageOptions={languageOptions}
-        darkMode={darkMode}
+        darkMode={darkModeProp ?? darkMode}
```
seungpark, line 103: does this need to be passed down ? seems like lg/code already consumes [here ](https://github.com/mongodb/leafygreen-ui/blob/main/packages/code/src/Code/Code.tsx#L80)

src/components/Code/Output.js
```diff
@@ -20,7 +22,16 @@ const Output = ({ nodeData: { children }, ...rest }) => {
           border-bottom-right-radius: 12px;
           border-bottom-left-radius: 12px;
         }
+
```
seungpark, line 26: can we convert this to be a styled div to gain consistency?

seungpark (commented):
minor comments below but overall looks good!

can we add to the PR description which ReST directives are used for these components for fast testing! 🙏

src/components/Code/Output.js
mmeigs: 🤷 sure! Clean up snooty while we go!

src/components/Code/Code.js
mmeigs: This was already here! I am keeping it to make sure that if there's a place in Snooty where we are specifically having dark mode code in a light mode setting, it can still happen.

src/components/Code/Output.js
mmeigs: Fixed! And I found a weird border bug while fixing this, so thank you times two!

src/components/Code/Output.js
```diff
@@ -20,7 +23,29 @@ const Output = ({ nodeData: { children }, ...rest }) => {
...
       `}
+      style={
+        darkMode
+          ? {
+              '--code-container-border': 'none',
+              '--code-pre-border': `1px solid ${palette.gray.dark2}`,
+              '--code-pre-border-top': 'none',
+            }
+          : {
+              '--code-container-border': 'auto',
```
seungpark, line 44: what does this `border: auto` value do? seems like an invalid arg for border

mmeigs: ah yes, I should use `initial`

src/components/Code/Output.js
```diff
@@ -1,26 +1,52 @@
...
-          border-bottom-left-radius: 12px;
-        }
-      `}
+    <OutputContainer
+      style={
+        darkMode
+          ? {
+              '--code-container-border': 'none',
+              '--code-pre-border': `1px solid ${palette.gray.dark2}`,
+              '--code-pre-border-top': 'none',
```
mayaraman19, line 42: do we need `--code-pre-border-top` as a variable if it's set to `"none"` in both places?

src/components/Code/Code.js
```diff
@@ -29,10 +22,11 @@ const sourceCodeStyle = css`
 
 const Code = ({
   nodeData: { caption, copyable, emphasize_lines: emphasizeLines, lang, linenos, value, source, lineno_start },
-  darkMode,
+  darkMode: darkModeProp,
```
caesarbell, line 25: I could be missing something. Where does the `darkMode` prop get passed in from? Isn't the `Code` component used within the ComponentFactory?

caesarbell: nvm, I saw a previous comment about this.

src/components/Code/Output.js
```diff
@@ -1,26 +1,50 @@
 import React from 'react';
-import { css } from '@emotion/react';
+import styled from '@emotion/styled';
```
caesarbell, line 2: ❤️

caesarbell (approved):
LGTM

seungpark (approved):

PR #1115, opened 2024-06-05
DOP-4741 💄  updates Steps & Procedure Component to use style props for dynamic styles

caesarbell:
### Stories/Links:

DOP-4741

### Current Behavior:

[MongoDB Atlas](https://www.mongodb.com/docs/atlas/)

[MongoDB Atlas Connect to your cluster](https://www.mongodb.com/docs/atlas/tutorial/connect-to-your-cluster/#connect-to-your-atlas-cluster)

### Staging Links:

[Staged: MongoDB Atlas](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/caesarbell/DOP-4741/index.html)

[Staged: MongoDB Atlas Connect to your cluster](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/caesarbell/DOP-4741/tutorial/connect-to-your-cluster/index.html#connect-to-your-atlas-cluster)

### Notes:

These changes come after agreeing on using the style props with `css` var for handling dynamic styling.

### README updates

- - [ ] This PR introduces changes that should be reflected in the README, and I have made those updates.
- - [x] This PR does not introduce changes that should be reflected in the README

PR #1121, opened 2024-06-07
DOP-4731: Toggle dark mode for ListTable

rayangler:
### Stories/Links:

DOP-4731

### Current Behavior:

[Server docs](https://www.mongodb.com/docs/manual/reference/command/analyzeShardKey/) - tables that include zebra striping
[Server docs](https://www.mongodb.com/docs/manual/core/causal-consistency-read-write-concerns/) - tables that are also in callouts

### Staging Links:

[Server docs](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/raymund.rodriguez/DOP-4731/reference/command/analyzeShardKey/index.html) - tables that include zebra striping
[Server docs](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/raymund.rodriguez/DOP-4731/core/causal-consistency-read-write-concerns/index.html) - tables that are also in callouts

### Notes:

As discussed in planning, upgrading the LG Table is out of scope for this ticket due to a lot of breaking changes and restructuring we'd need to do to accommodate the latest version of the table component. For this reason, I went ahead and overwrote the dark mode colors for just the rows for the table, and it matches the site theme well now. Otherwise, the style would not match the current dark mode theme:

Screen Shot 2024-06-07 at 2 18 06 PM

We should consider upgrading the component as part of DOP-3614 relatively soon though.

### README updates


mayaraman19 (commented):
Thank you for doing this, the zebra stripes look great! I have a couple comments:
- we probably should have given this more visibility, but offline we decided that our convention for changing CSS styles in dark mode would be to use CSS variables that are passed in via the `style` attribute - check out [this change](https://github.com/mongodb/snooty/pull/1115/files#diff-7db929129a1305400bc4592bbd7b7ff13941a701690efac636463f1b320c5b25L51-R51) in one of Caesar's PRs, hopefully that makes the convention more clear or also reach out if you have questions ofc
- Also, maybe this is out of the scope of this PR, but the font for some tables/parts of tables changes in Dark Mode? The table under [this section](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/raymund.rodriguez/DOP-4731/reference/command/analyzeShardKey/#keycharacteristics) for example

src/components/ListTable.js
```diff
@@ -17,14 +18,19 @@ const align = (key) => {
...
   ${customAlign && `text-align: ${align(customAlign)}`};
   ${customWidth && `width: ${customWidth}`};
   margin: ${theme.size.medium} 0;
 
   table & {
     margin: 0;
   }
+
+  tbody > tr:nth-of-type(2n) {
+    // Need this since we're overriding background color for all rows for dark mode
```
seungpark, line 31: is this valid comment? seems like its for every 2 rows vs comment states for all rows. each row seems like its alternating also:

image


image

[from this page](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/raymund.rodriguez/DOP-4731/reference/command/analyzeShardKey/index.html#output)

seungpark (commented):
looks great! just a minor question below concerning alternating rows

src/components/ListTable.js
rayangler: Sorry, I might have phrased the comment awkwardly. "we're overriding background color for all rows in dark mode" refers to the styling [here](https://github.com/mongodb/snooty/pull/1121/files#diff-c0635850a29537b5e9333b6e82a9306b70f036a6c42f9c129a903418a00ac4ebR47). Applying that color prevents zebra striping from appearing in dark mode, so we need to override it again with `!important` here to restore zebra striping again.

I can rewrite comment to something like "Overriding bg color for dark mode rows prevents zebra striping. Adding this important rule to restore zebra striping" if that makes it clearer

seungpark: ah that makes sense!  thanks for clarification

rayangler: Updated the comment 👍

rayangler:
@mayaraman19 I've gone ahead and addressed the 2 things you mentioned. The style prop does make it a bit awkward with needing to redefine certain default styles we're overriding, but hopefully it still looks sane. Let me know what you think!

I've also gone ahead and updated staging links with a couple more examples as well

mayaraman19 (approved):
👍 thank you for doing this!

PR #1130, opened 2024-06-12
:sparkles: DOP-4662 adds community pill to cards for when using tags

caesarbell:
### Stories/Links:

DOP-4662

### Current Behavior:

[Python Landing](https://www.mongodb.com/docs/languages/python/)

[Docs Landing](https://www.mongodb.com/docs/)

### Staging Links:

[Staged: Python Landing](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/caesarbell/DOP-4662/languages/python/index.html)

[Staged: Docs Landing](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/forked/dark-icons/landing/caesarbell/DOP-4662/index.html) (built from a forked repo)

### Notes:

- For this PR, I modified the community pill component to be reusable and swapped out the Tag component to use the new community pill component.

### README updates

### DIFF
--- a/src/components/CommunityPillLink.js
+++ b/src/components/CommunityPillLink.js
@@ -13,18 +13,6 @@ const communityPillVariants = {
   green: Variant.Green,
 };
 
-/**
- * CommunityPillLink component
- *
- * @param {Object} param0 - The parameters object
- * @param {Object} param0.nodeData - Data for the node
- * @param {String} param0.nodeData.argument - Text to display for Link
- * @param {Object} [param0.nodeData.options] - Options for the node data
- * @param {string} [param0.nodeData.options.url] - URL in the options
- * @param {string} param0.variant - Variant of the community pill link
- * @param {string} [param0.text='community built'] - Text to display, defaults to 'community built'
- * @returns {JSX.Element} The rendered CommunityPillLink component
- */
 const CommunityPillLink = ({ nodeData, variant, text = 'community built' }) => {
   const { argument, options: { url } = {} } = nodeData || {};
 
--- a/src/components/CommunityPillLink.js
+++ b/src/components/CommunityPillLink.js
@@ -13,13 +13,13 @@ const communityPillVariants = {
   green: Variant.Green,
 };
 
-const CommunityPillLink = ({ nodeData, variant, text = 'community built' }) => {
+const CommunityPillLink = ({ nodeData, variant = 'lightGray', text = 'community built' }) => {
   const { argument, options: { url } = {} } = nodeData || {};
 
   return (
     <div>
       {nodeData && argument && url && <Link to={url}>{getPlaintext(argument)}</Link>}
-      <Badge variant={communityPillVariants[variant] || Variant.LightGray}>{text}</Badge>
+      <Badge variant={communityPillVariants[variant]}>{text}</Badge>
     </div>
   );
 };
--- a/src/components/CommunityPillLink.js
+++ b/src/components/CommunityPillLink.js
@@ -1,19 +1,37 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import Badge from '@leafygreen-ui/badge';
+import Badge, { Variant } from '@leafygreen-ui/badge';
 import { getPlaintext } from '../utils/get-plaintext';
 import Link from './Link';
 
-const CommunityPillLink = ({
-  nodeData: {
-    argument,
-    options: { url },
-  },
-}) => {
+const communityPillVariants = {
+  darkGray: Variant.DarkGray,
+  lightGray: Variant.LightGray,
+  red: Variant.Red,
+  yellow: Variant.Yellow,
+  blue: Variant.Blue,
+  green: Variant.Green,
+};
+
+/**
+ * CommunityPillLink component
+ *
+ * @param {Object} param0 - The parameters object
+ * @param {Object} param0.nodeData - Data for the node
+ * @param {String} param0.nodeData.argument - Text to display for Link
+ * @param {Object} [param0.nodeData.options] - Options for the node data
+ * @param {string} [param0.nodeData.options.url] - URL in the options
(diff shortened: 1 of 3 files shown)


src/components/CommunityPillLink.js
```diff
@@ -1,19 +1,37 @@
...
+ * @param {Object} param0 - The parameters object
+ * @param {Object} param0.nodeData - Data for the node
+ * @param {String} param0.nodeData.argument - Text to display for Link
+ * @param {Object} [param0.nodeData.options] - Options for the node data
+ * @param {string} [param0.nodeData.options.url] - URL in the options
+ * @param {string} param0.variant - Variant of the community pill link
+ * @param {string} [param0.text='community built'] - Text to display, defaults to 'community built'
+ * @returns {JSX.Element} The rendered CommunityPillLink component
+ */
+const CommunityPillLink = ({ nodeData, variant, text = 'community built' }) => {
```
mmeigs, line 28: Does variant need to default to "lightgray"?

src/components/CommunityPillLink.js
```diff
@@ -1,19 +1,37 @@
...
+/**
+ * CommunityPillLink component
+ *
+ * @param {Object} param0 - The parameters object
+ * @param {Object} param0.nodeData - Data for the node
+ * @param {String} param0.nodeData.argument - Text to display for Link
+ * @param {Object} [param0.nodeData.options] - Options for the node data
+ * @param {string} [param0.nodeData.options.url] - URL in the options
+ * @param {string} param0.variant - Variant of the community pill link
+ * @param {string} [param0.text='community built'] - Text to display, defaults to 'community built'
```
mayaraman19, line 25: this is a little bit difficult to understand because of the repeated references to `param0`, `nodeData`, etc. Might be better to remove + ensure the code is straightforward to understand in order to reduce confusion?

src/components/CommunityPillLink.js
caesarbell: @mmeigs good question, in this case, yes the variant prop should default to `lightgray` to support the original use case of the community pill. 

I am doing so within the Badge component i.e.

```js
<Badge variant={communityPillVariants[variant] || Variant.LightGray}>{text}</Badge>
```

caesarbell: I could re-write this to be more like 

```js
const CommunityPillLink = ({ nodeData, variant = 'lightgray', text = 'community built' }) => {
  const { argument, options: { url } = {} } = nodeData || {};

  return (
    <div>
      {nodeData && argument && url && <Link to={url}>{getPlaintext(argument)}</Link>}
      <Badge variant={communityPillVariants[variant]}>{text}</Badge>
    </div>
  );
};
```
if you prefer.

src/components/CommunityPillLink.js
```diff
@@ -1,19 +1,37 @@
...
+ * @returns {JSX.Element} The rendered CommunityPillLink component
+ */
+const CommunityPillLink = ({ nodeData, variant, text = 'community built' }) => {
+  const { argument, options: { url } = {} } = nodeData || {};
```
mayaraman19, line 29: is there a way to define these as optional using the `nodeData` prop instead i.e.

```
const CommunityPillLink = ({ nodeData: {argument?, options?: {url}}, variant, text = 'community built' }) => {
...
```
or something like that. I'm forgetting my Javascript currently though so idk if this is possible

src/components/CommunityPillLink.js
mayaraman19: i like this!

src/components/CommunityPillLink.js
caesarbell: @mayaraman19 I can remove this, it was to help when hovering your mouse over the component when called in another file.

src/components/CommunityPillLink.js
caesarbell: @mayaraman19 just out of curiosity, what are you trying to achieve, readability? I am asking so I can better understand the suggestion/comment.

caesarbell: @mayaraman19 also, from what I understand that might not be doable in vanilla JavaScript. That is more of a Typescript approach.

src/components/CommunityPillLink.js
caesarbell: @mayaraman19 the former or latter?

mmeigs: Wow, my brain. Yep - you're so right. I guess it would be nice to have the defaults be declared the same way! Either way is fine with me.

caesarbell: @mmeigs no worries, it's the middle of the week and our brains have been on overdrive. I will update the code to declare it the same way.

mmeigs (approved):

src/components/CommunityPillLink.js
mayaraman19: i guess the intent was to use the convention we typically use for getting values from `nodeData` as well as reduce the need to check for nodeData [here](https://github.com/mongodb/snooty/pull/1130/files#diff-6854794849b1ee1c01317a7f35f873920c41fd61c9d3477cd7575e5a26ff5cf7R21).

src/components/CommunityPillLink.js
mayaraman19: i see! that is a nice touch. I guess the confusion lies in naming `param0` since it's not mentioned by name in the code, as well as I don't understand why some have brackets `[]` around them, since they don't seem to be array values? minor & non-blocking though, i trust your judgement!

mayaraman19 (approved):

src/components/CommunityPillLink.js
caesarbell: @mayaraman19 I went with your feedback, no need to add confusion when it's not needed. The brackets indicated the default value.

PR #1132, opened 2024-06-13
DOP-4641 / DOP-4740: Adding Language Logos

mayaraman19:
### Stories/Links:

DOP-4641 and DOP-4740 - basically the same

### Current Behavior:

[Current Cloud Docs driver-connection Page](https://www.mongodb.com/docs/atlas/driver-connection/)

### Staging Links:

[Cloud docs driver-connection Staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/maya/DOP-4641/driver-connection/index.html)

### Notes:

Adds Dark Mode versions of necessary icons (Rust, Node, Java) locally in Snooty and enables the Tab Selector to use these. Also updates the Java light mode logo to match what's used in Dark Mode, the drivers landing page and the IA on the docs landing page.

Need to also update the language selector on Code components to reflect Dark Mode icons.

### README updates


src/components/Tabs/tab-context.js
```diff
@@ -61,9 +64,42 @@ const DRIVER_ICON_MAP = {
...
+  rust: IconRustDark,
+  'rust-sync': IconRustDark,
+  'rust-async': IconRustDark,
+  scala: IconScala,
+  shell: IconShell,
+  swift: IconSwift,
+  'swift-async': IconSwift,
+  'swift-sync': IconSwift,
+  typescript: IconTypescript,
+};
```
mayaraman19, line 97: unfortunate to have so much duplication here but because these mappings are created at build time I wasn't sure what the best way to do things is - I'll try to look into sourcing them at runtime instead perhaps

mmeigs (commented):
Looks good! I'm wondering if you could get the svg code for the new Dark images rather than putting an href for a png inside an svg?

caesarbell:
@mayaraman19  Is it beneficial to provide a staging link with the dark mode feature flag set to false for regression testing?

src/components/Code/Code.js
```diff
@@ -11,15 +11,30 @@ import { useDarkMode } from '@leafygreen-ui/leafygreen-provider';
...
 const sourceCodeStyle = css`
   display: flex;
   align-items: center;
   justify-content: center;
 `;
 
+// Returns the icon associated with the driver language that would be
+// shown on our TabSelector component
+const getDriverImage = (driver, driverIconMap) => {
+  const DriverIcon = driverIconMap?.[driver];
```
caesarbell, line 27: @mayaraman19 super nit, but CMIIW, `driverIconMap` will not return `undefined` or `null`, if that is correct why are we conditionally chaining it? Once again if I am misunderstanding or misreading it feel free to ignore it.

src/components/Code/Code.js
```diff
@@ -11,15 +11,30 @@ import { useDarkMode } from '@leafygreen-ui/leafygreen-provider';
...
+
+  // Use LG File icon as our default placeholder for images. This overwrites
+  // LG's Language Switcher current default icon. See:
+  // https://github.com/mongodb/leafygreen-ui/blob/6041b89bf5f9dc1e5ea76018bc2cd84bc1fd6faf/packages/code/src/LanguageSwitcher.tsx#L135-L136
+  return <Icon glyph="File" />;
+};
+
 const Code = ({
   nodeData: { caption, copyable, emphasize_lines: emphasizeLines, lang, linenos, value, source, lineno_start },
   darkMode: darkModeProp,
```
caesarbell, line 40: (for learning purposes) I am just curious, is there ever a case where `darkMode` will passed down as a prop? I was under the impression it was called within the `ComponentFactory`.

caesarbell (approved):
LGTM - I just left a small suggestion, and one question for learning purposes.

src/components/icons/Java.js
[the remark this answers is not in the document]
mayaraman19: i'll look into this! sounds interesting

src/components/Code/Code.js
mayaraman19: @caesarbell this was previous behavior that I am preserving, so I'm not entirely sure either. I can remove it.

src/components/Code/Code.js
mayaraman19: Also previous behavior that is being preserved. I'm assuming that sometimes there are, for some reason, specific code blocks that are supposed to be dark mode even in a light mode setting.

src/components/Code/Code.js
caesarbell: @mayaraman19 gotcha, that makes sense. I guess if we can safely remove it, we should. Thoughts?

seungpark (approved):

PR #1134, opened 2024-06-17
Remove old lighthouse workflow

mmeigs:
### Stories/Links:

No ticket. Just clean up.

### Notes:

I have deleted the old Lighthouse github workflow and all references/files that were a part of it.

My open PR will re-add some of these back, but they will all be current and working pieces rather than a constantly failing action.

Also: I am working on that open PR to change back the tsconfig.

### README updates

- - [x] This PR introduces changes that should be reflected in the README, and I have made those updates.


package-lock.json
```diff
@@ -12,7 +12,7 @@
         "@emotion/eslint-plugin": "^11.7.0",
         "@emotion/react": "^11.7.1",
         "@emotion/styled": "^11.0.0",
-        "@leafygreen-ui/badge": "^5.0.1",
+        "@leafygreen-ui/badge": "^8.1.3",
```
seungpark, line 15: weird diff here but this looks correct

seungpark (approved):

package-lock.json
mmeigs: so odd...

PR #1141, opened 2024-06-20
DOP-4761: update LG/tabs to v12 to render content in HTML

seungpark:
### Stories/Links:

DOP-4761

### Current Behavior:

[Tab behavior on homepage](https://www.mongodb.com/docs/atlas/)
[Tabs in inner content page](https://www.mongodb.com/docs/atlas/access-tracking/)

### Staging Links:

[Tabs on a landing page](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/seung.park/DOP-4761/index.html)
[Tabs in inner content page](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/seung.park/DOP-4761/access-tracking/index.html)

### Notes:
- Styling and user interaction should see no change
- Tabbed content is visible in the HTML produced (as well as the page HTML). You can verify this by going to the [S3 bucket](https://us-east-2.console.aws.amazon.com/s3/object/docs-mongodb-org-stg?region=us-east-2&bucketType=general&prefix=master/cloud-docs/seung.park/DOP-4761/index.html) and downloading the HTML file and seeing the tabbed content within

### README updates

### DIFF
--- a/src/components/Tabs/index.js
+++ b/src/components/Tabs/index.js
@@ -1,6 +1,5 @@
 import React, { useCallback, useContext, useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
-import styled from '@emotion/styled';
 import { css, cx } from '@leafygreen-ui/emotion';
 import { Tabs as LeafyTabs, Tab as LeafyTab } from '@leafygreen-ui/tabs';
 import { CodeProvider } from '../Code/code-context';
@@ -64,11 +63,12 @@ const getTabsStyling = ({ isHidden, isProductLanding }) => css`
   ${isProductLanding && landingTabsStyling};
 `;
 
-const TabContent = styled('div')`
+const tabContentStyling = css`
   margin-top: ${theme.size.medium};
-  ${(props) =>
-    props.isProductLanding &&
-    `  display: grid;
+`;
+
+const productLandingTabContentStyling = css`
+  display: grid;
   column-gap: ${theme.size.medium};
   grid-template-columns: repeat(2, 1fr);
 
@@ -81,7 +81,7 @@ const TabContent = styled('div')`
 
   @media ${theme.screenSize.upToLarge} {
     display: block;
-  }`}
+  }
 `;
 
 const Tabs = ({ nodeData: { children, options = {} }, page, ...rest }) => {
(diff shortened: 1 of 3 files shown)


src/components/Tabs/index.js
```diff
@@ -1,5 +1,6 @@
 import React, { useCallback, useContext, useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
```
rayangler, line 3: If we want to eventually stick to a singular way of defining CSS styling, I would suggest LG's emotion/css package. If we do want to use `styled`, let's use Emotion's `styled` instead of `styled-components`. (We only need `styled-components` as a dependency for Redoc for now, but we can hopefully remove it in the future)

src/components/Tabs/index.js
```diff
@@ -80,9 +82,25 @@ const landingTabStyling = css`
...
+    border-radius: ${theme.size.small};
+    grid-column: 2;
+    margin-top: 0px;
+    display: block;
+  }
+
+  @media ${theme.screenSize.upToLarge} {
+    display: block;
+  }`
+      : ''}
```
rayangler, line 103: (Nit) If we don't plan to render a style if the condition is false, I think we can just do something like

```js
${(props) => props.isProductLanding && whatever}
```

This might make it more readable.

rayangler: Also, this looks almost the same as `landingTabStyling`. Is this something we want to consolidate?

src/components/Tabs/index.js
seungpark: great catch! auto finish with imports

seungpark: done. removed the styled component and replaced with div with a css className, conditional on landing 👍

caesarbell (approved):
Thanks for fixing these so quickly. LGTM

src/components/Tabs/index.js
seungpark: removed landingTabStyling as this was consolidated and applied to the tab content

src/components/Tabs/index.js
```diff
@@ -152,10 +152,15 @@ const Tabs = ({ nodeData: { children, options = {} }, page, ...rest }) => {
...
 
             return (
-              <LeafyTab className={cx(getTabStyling({ isProductLanding }))} key={tabId} name={tabTitle}>
-                {tab.children.map((child, i) => (
-                  <ComponentFactory {...rest} key={`${tabId}-${i}`} nodeData={child} />
-                ))}
+              <LeafyTab key={tabId} name={tabTitle}>
+                <div
+                  className={cx(tabContentStyling, isProductLanding ? productLandingTabContentStyling : '')}
+                  isProductLanding={isProductLanding}
```
rayangler, line 158: (Nit) Is this attribute being used anywhere? If not, we should remove it

rayangler (approved):

src/components/Tabs/index.js
seungpark: good catch! removed

PR #1149, opened 2024-06-25
:art: DOP-4780 overrides the weblink-link any styles for dark mode

caesarbell:
### Stories/Links:

DOP-4780

### Current Behavior:

[Docs Manual](https://www.mongodb.com/docs/manual/tutorial/install-mongodb-enterprise-on-red-hat-tarball/#platform-support)

### Staging Links:

[Staged: Docs Manaual](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/caesarbell/DOP-4780/tutorial/install-mongodb-enterprise-on-red-hat-tarball/index.html#platform-support)

### Notes:

- overrides the `-webkit-any-link` color and color for visited styles for dark mode

### README updates


src/components/Footnote/FootnoteReference.js
```diff
@@ -1,23 +1,44 @@
 import React, { useContext } from 'react';
+import { useDarkMode } from '@leafygreen-ui/leafygreen-provider';
+import { css } from '@emotion/react';
+import { palette } from '@leafygreen-ui/palette';
 import PropTypes from 'prop-types';
 import { getNestedValue } from '../../utils/get-nested-value';
 import FootnoteContext from './footnote-context';
 
+const refStyles = (darkMode) => css`
```
caesarbell, line 9: I opted for applying the dynamic styles within the static styles since we might not use the style props.

If this was premature, I can update the code to use the style props with CSS vars.

src/components/Footnote/FootnoteReference.js
```diff
@@ -1,23 +1,44 @@
 import React, { useContext } from 'react';
+import { useDarkMode } from '@leafygreen-ui/leafygreen-provider';
+import { css } from '@emotion/react';
```
seungpark, line 3: lets use the css / cx from '@leafygreen-ui/emotion' and move into classNames for consistency

seungpark (changes requested):
minor change below for using LG / emotion

src/components/Footnote/FootnoteReference.js
caesarbell: @seungpark CMIIW, but `css / cx from '@leafygreen-ui/emotion'` doesn't work on non-LG components, so we would need to use `css @emotion/react` for non-LG components.

caesarbell: If you want to use the classnames we would probably need to do something like this 

```js
import { ClassNames } from '@emotion/react'
<ClassNames>
      {({ css, cx }) => (
        <div
          className={cx(
            'some-class',
            css`
              color: yellow;
            `
          )}
        />
      )}
    </ClassNames>
```

seungpark: i believe this works with className prop (not css prop) 

```
import { css, cx } from '@leafygreen-ui/emotion';
...



className={cx("footnote-reference header-buffer", refStyles(darkMode))}`
```

src/components/Footnote/FootnoteReference.js
```diff
@@ -1,23 +1,40 @@
...
   // anonymous footnoteReferences originating from included files -- docutils
   // appears to assign IDs within the included files before they are collated
 
   const ref = refname || id.replace('id', '');
   const uid = refname ? `${refname}-${id}` : id;
+
   return (
-    <a className="footnote-reference header-buffer" href={`#footnote-${ref}`} id={`ref-${uid}`}>
+    <a
+      className={cx('footnote-reference header-buffer', darkMode && refStyles)}
```
mmeigs, line 34: If you make this a ternary instead, emotion won't have to create a className for light mode.

``` 
className={cx('footnote-reference header-buffer', darkMode ? refStyles: '')}
```

mmeigs: You know what - I might be wrong. And anyway, this doesn't really matter. Leave it how it is!

mmeigs (approved):

PR #1151, opened 2024-06-26
:art: DOP-4635 using static css approach

caesarbell:
### Stories/Links:

DOP-4635

### Current Behavior:

[Previous Staged: Docs Manual](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/caesarbell/DOP-4635/tutorial/install-mongodb-enterprise-on-red-hat-tarball/index.html#footnote-oracle-linux)

### Staging Links:

[New Staged without Style Prop: Docs Manual](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/caesarbell/DOP-4635-ClassName/tutorial/install-mongodb-enterprise-on-red-hat-tarball/index.html#platform-support)

### Notes:

- Same styles applied, but now using only `classnames` rather than using the style props. 

### README updates

### DIFF
--- a/src/components/Footnote/index.js
+++ b/src/components/Footnote/index.js
@@ -22,12 +22,12 @@ const tableStyling = (darkMode) => css`
   ${darkMode &&
   `
       tbody tr td a {
-        color: var(--color);
+        color: ${palette.blue.light1};
       }
     `}
 
   :target {
-    background-color: var(--background-color);
+    background-color: ${darkMode ? palette.gray.dark2 : '#ffa'};
   }
 `;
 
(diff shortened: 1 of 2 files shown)


seungpark (approved):

src/components/Footnote/index.js
```diff
@@ -49,25 +49,8 @@ const Footnote = ({ nodeData: { children, id, name }, ...rest }) => {
...
   return (
-    <table
-      className="header-buffer"
-      css={tableStyling(darkMode)}
-      style={footnoteDynamicStyles}
-      frame="void"
-      id={`footnote-${ref}`}
-      rules="none"
-    >
+    <table className="header-buffer" css={tableStyling(darkMode)} frame="void" id={`footnote-${ref}`} rules="none">
```
mmeigs, line 53: We're going to use className instead of css, right?

caesarbell: @mmeigs CMIIW, to be able to use classNames for non-LG, we would need to do something like 

```js
import { ClassNames } from '@emotion/react'
 <ClassNames>
      {({ css, cx }) => (
        <div
          className={cx(
            'some-class',
            css`
              color: yellow;
            `
          )}
        />
      )}
    </ClassNames>
```

or unless we can do something like 

```js
<table className="header-buffer" className={tableStyling(darkMode)} frame="void" id={`footnote-${ref}`} rules="none">
</table>
````

mmeigs (approved):

PR #1169, opened 2024-07-17
DOP-4835: Fix meta robots not appearing due to missing page data

rayangler:
### Stories/Links:

DOP-4835

### Current Behavior:

[Atlas prod](https://www.mongodb.com/docs/atlas/atlas-data-processing/) - `noindex` does not appear in a meta tag

### Staging Links:

[Atlas staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/raymund.rodriguez/DOP-4835/atlas-data-processing/index.html) - `noindex` appears in a meta tag.

### Notes:

* Re-introduces page data into Gatsby Head. This was missing due to page context no longer having page data.
* Throws an error if page AST is missing from Gatsby Head again. This should hopefully stop the build with an error to emphasize that this is not expected.
* Adjusts existing data sources for tests for the `Head` component to split between page context and data. This should better mock how data is being passed into the component.

### README updates

### DIFF
--- a/tests/unit/Head.test.js
+++ b/tests/unit/Head.test.js
@@ -27,7 +27,7 @@ describe('Head', () => {
       useSnootyMetadata.mockImplementation(() => mockEOLSnootyMetadata);
     });
     it('renders the canonical tag from the snooty.toml', () => {
-      render(<Head pageContext={mockCompleteEOLPageContext} />);
+      render(<Head pageContext={mockCompleteEOLPageContext.pageContext} data={mockCompleteEOLPageContext.data} />);
       const _canonical = mockEOLSnootyMetadata.canonical;
       const canonicalTag = screen.getByTestId('canonical');
       expect(canonicalTag).toBeInTheDocument();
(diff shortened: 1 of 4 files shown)


tests/unit/Head.test.js
```diff
@@ -178,19 +178,21 @@ describe('Head', () => {
...
     };
 
-    let mockPageContext = mockHeadPageContext;
     it('renders the canonical tag from directive rather than pulling from snooty.toml', () => {
-      mockPageContext = mockCompleteEOLPageContext;
-      mockPageContext.page.children.push(metaCanonical);
-      render(<Head pageContext={mockPageContext} />);
+      const mockPageContext = { ...mockHeadPageContext.pageContext };
+      const mockData = { ...mockCompleteEOLPageContext.data };
+      mockData.page.ast.children.push(metaCanonical);
```
rayangler, line 184: Spreading the data instead of reusing local variable due to potential data leaking in between tests due to object modifications for the last 2 unit tests below. Spreading here will help keep copies of data separate to promote independent tests

mmeigs (approved):

rayangler:
Realized that we can throw the error instead of just printing it out, so I changed that in the code. Might be more appropriate given the importance of having the AST in this component at the moment

caesarbell:
> Realized that we can throw the error instead of just printing it out, so I changed that in the code. Might be more appropriate given the importance of having the AST in this component at the moment

Good call, I was just about to ask about that, and if we could cause the build to fail since it seem important to have.

PR #1175, opened 2024-07-23
DOP-4825: Hide Sitewide CTA banners on Scroll

mayaraman19:
### Stories/Links:

DOP-4825

### Current Behavior:

[Docs landing](https://www.mongodb.com/docs/)
[Manual](https://www.mongodb.com/docs/manual/)
[Cloud-docs](https://www.mongodb.com/docs/atlas/getting-started/)

### Staging Links:

[Docs landing](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/maya/DOP-4825/index.html) (behavior should be unchanged)
[Manual](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/v7.0/docs/maya/DOP-4825/index.html)
[Cloud docs](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/maya/DOP-4825/index.html)

[Example without banner](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/no-banner/v7.0/docs/maya/DOP-4825/index.html)
[Example with dark mode/action bar enabled](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/dark-mode-enable/master/cloud-docs/maya/DOP-4825/index.html)

### Notes:

- Not sure how useful this PR is because the new action bar will be what automatically sticks to the top after it's released. 
- Also, I left `preview-header.js` untouched because I'm not sure how to test it and, although I could probably just copy what I did in `Header/index.js` I don't want to accidentally mess anything up, although any advice on this would be appreciated!

### README updates

### DIFF
--- a/src/components/Header/index.js
+++ b/src/components/Header/index.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
 import { UnifiedNav } from '@mdb/consistent-nav';
@@ -6,13 +6,13 @@ import { SidenavMobileMenuDropdown } from '../Sidenav';
 import SiteBanner from '../Banner/SiteBanner';
 import { theme } from '../../theme/docsTheme';
 import { AVAILABLE_LANGUAGES, getCurrLocale, onSelectLocale } from '../../utils/locale';
+import { HeaderContext } from './header-context';
 
 const StyledHeaderContainer = styled.header(
   (props) => `
   grid-area: header;
   top: 0;
-  // TODO : CHANGE THIS
-  margin-top: 40px;
+  margin-top: ${props.hasBanner ? theme.header.bannerHeight : '0px'};
   z-index: ${theme.zIndexes.header};
   ${
     props.template === 'landing' || props.template === 'errorpage' || process.env['GATSBY_ENABLE_DARK_MODE'] === 'true'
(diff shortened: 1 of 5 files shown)


src/components/Header/index.js
```diff
@@ -24,24 +26,28 @@ const Header = ({ sidenav, eol, template }) => {
   const unifiedNavProperty = 'DOCS';
 
   const enabledLocales = AVAILABLE_LANGUAGES.map((language) => language.localeCode);
+  const { bannerContent } = useContext(HeaderContext);
+
   return (
-    <StyledHeaderContainer template={template}>
+    <>
       <SiteBanner />
```
seungpark, line 33: [minor nit]. can we position this within `StyledHeaderContainer` and remove the absolute positioning

seungpark (commented):
minor comment below but not a blocker. also can we get a staging link without the dark mode flag for sanity check. thank you!

src/components/Header/index.js
mayaraman19: I removed the banner from `StyledHeaderContainer` because I think sticky items can only "stick" within their container, so it must be removed if I wanted the banner to scroll but the header to remain sticky - otherwise whole container was sticky or none of it at all

I added the absolute positioning because otherwise the banner would show up at the bottom of the page, but I can explore changing this.

seungpark: ah i see. this is due to the GlobalGrid etc. this works for me as well as we already have banner height set in the theme and setting margins accordingly 👍

seungpark (approved):

PR #1178, opened 2024-07-29
DOP-4862: Dark Mode toggle in Sidenav

mmeigs:
### Stories/Links:

DOP-4862

### Current Behavior:

[Atlas](https://www.mongodb.com/docs/atlas/)
[Landing](https://www.mongodb.com/docs/)

### Staging Links:

[Atlas](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/matt.meigs/DOP-4862-dark-mode-toggle/index.html)
[Landing](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/matt.meigs/DOP-4862-dark-mode-toggle/index.html)

### Notes:

This PR adds a Dark Mode IconButton toggle to the Sidenav.

It also removes the ActionBar from dark mode.

### README updates


seungpark:
CMIIW but i understood this as a toggle and not a menu - the staging links are using the same dropdown menu as we used previously.

src/components/ActionBar/DarkModeDropdown.js
```diff
@@ -36,7 +36,7 @@ const DarkModeDropdown = ({ ...props }) => {
   return (
     <Menu
       className={cx(menuStyling)}
-      usePortal={false}
+      usePortal={true}
```
rayangler, line 39: I'm running into an issue on the staging links where scrolling down while clicking on the dropdown causes the page to scroll back up. The dropdown menu is then only present when scrolling back down. Looks like the portal might be causing this.

https://github.com/user-attachments/assets/1ca71a4a-5aec-41e1-be42-57f3f4c04694

mmeigs:
Spoke to Liz, and @seungpark was right! I assumed we were moving the ActionBar dropdown for this. But they discussed offline that this will be a simple IconButton "toggle".

src/components/ActionBar/DarkModeDropdown.js
mmeigs: Not an issue anymore with removing the `Menu` component, but thank you for catching this!

src/components/Sidenav/DarkModeToggle.js
```diff
@@ -0,0 +1,32 @@
+import React, { useContext } from 'react';
+import { css, cx } from '@leafygreen-ui/emotion';
+import Icon from '@leafygreen-ui/icon';
+import IconButton from '@leafygreen-ui/icon-button';
+import { useDarkMode } from '@leafygreen-ui/leafygreen-provider';
+import { DARK_THEME_CLASSNAME, DarkModeContext, LIGHT_THEME_CLASSNAME } from '../../context/dark-mode-context';
+
+const toggleStyle = css`
+  margin: 0 16px;
```
seungpark, line 9: [minor] to use theme value here

src/components/Sidenav/DarkModeToggle.js
```diff
@@ -0,0 +1,32 @@
...
+  const onToggle = () => {
+    setDarkModePref(darkMode ? LIGHT_THEME_CLASSNAME : DARK_THEME_CLASSNAME);
+  };
+
+  return (
+    <IconButton
+      className={cx(toggleStyle)}
+      onClick={onToggle}
+      aria-label="Dark Mode Toggle"
+      aria-labelledby="Dark Mode Toggle"
```
rayangler, line 25: (Nit) I know this is based on pre-existing code, but I wonder if we can remove `aria-labelledby`. `aria-label` seems more appropriate since `aria-labelledby` isn't referencing existing text from another element

seungpark (commented):
not sure why but seems like Header component has `position: sticky` removed from your staging links !

rayangler (approved):
LGTM!

rayangler:
> not sure why but seems like Header component has position: sticky removed from your staging links !

@seungpark I think that's because the sticky position was removed when the dark mode flag is enabled, originally to make room for the action bar. Is the plan to only remove the sticky position when the action bar is present?

seungpark:
> > not sure why but seems like Header component has position: sticky removed from your staging links !
> 
> @seungpark I think that's because the sticky position was removed when the dark mode flag is enabled, originally to make room for the action bar. Is the plan to only remove the sticky position when the action bar is present?

Ah good point. Think we would want to separate that to the feature flag for the new Action bar project branch cc @mmeigs

src/layouts/index.js
```diff
@@ -114,6 +113,10 @@ const DefaultLayout = ({ children, data: { page }, pageContext: { slug, repoBran
       >
         <GlobalGrid isInPresentationMode={isInPresentationMode}>
           {!isInPresentationMode ? <Header sidenav={sidenav} eol={eol} slug={slug} template={template} /> : <div />}
+          <StyledContentContainer>
+            <ContentTransition slug={slug}>{children}</ContentTransition>
+          </StyledContentContainer>
+          {/* Sidenav is after the Content to ensure DOM hierarchy without using z-index */}
```
mayaraman19, line 119: I think this might not be good for accessibility since the visual order (left-to-right, top-to-bottom) of the page will not match the DOM order. for example, if you try tabbing through the page, the order feels a little bit off. However tabbing through the sidenav on our site already isn't a great experience so idk lol, this might be fine 👍

mmeigs: Huh, alright - damn. I'll put the z-indexes back for now.

seungpark (approved):
LGTM ! now we can release dark mode without blockers <3

PR #1193, opened 2024-08-05
DOP-4879: use feedback widget in action bar (desktop)

seungpark:
### Stories/Links:

DOP-4879

### Current Behavior:

[Current prod feedback uses FAB for feedback](https://www.mongodb.com/docs/atlas/)

### Staging Links:

[Staging link with feedback enabled in action bar](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/seung.park/DOP-4879/index.html)
[Gatsby preview build test](https://preview-mongodbseungpark.gatsbyjs.io/drivers/DOP-4817/)

### Notes:
- This PR updates the Feedback Module to be in the Action Bar (for Desktop)
- Feedback behavior should not change as existing, but there are small design changes that are noticeable 
    - caption to indicate rating
    - update padding and sizing of feedback card
    - update copy based on selected feedback rating

### README updates

### DIFF
--- a/src/components/Widgets/FeedbackWidget/FeedbackCard.js
+++ b/src/components/Widgets/FeedbackWidget/FeedbackCard.js
@@ -16,11 +16,6 @@ import { useFeedbackContext } from './context';
 import useNoScroll from './hooks/useNoScroll';
 
 const FloatingContainer = styled.div`
-  position: fixed;
-  z-index: 14;
-  bottom: ${({ hasOpenLabDrawer }) => (hasOpenLabDrawer ? '70px' : theme.size.large)};
-  right: ${theme.size.large};
-
   @media ${theme.screenSize.upToSmall} {
     padding-top: ${({ top }) => `${top}`};
     right: 0;
--- a/src/components/Widgets/FeedbackWidget/FeedbackCard.js
+++ b/src/components/Widgets/FeedbackWidget/FeedbackCard.js
@@ -27,7 +27,8 @@ const FloatingContainer = styled.div`
 
 const Card = styled(LeafygreenCard)`
   /* Card Size */
-  width: 234px;
+  width: 290px;
+  padding: ${theme.size.medium} ${theme.size.default};
   display: flex;
   flex-direction: column;
   position: relative;
--- a/src/components/Widgets/FeedbackWidget/FeedbackCard.js
+++ b/src/components/Widgets/FeedbackWidget/FeedbackCard.js
@@ -3,12 +3,10 @@ import styled from '@emotion/styled';
 import LeafygreenCard from '@leafygreen-ui/card';
 import { useDarkMode } from '@leafygreen-ui/leafygreen-provider';
 import { palette } from '@leafygreen-ui/palette';
-import { feedbackId } from '../FeedbackWidget/FeedbackForm';
 import { theme } from '../../../../src/theme/docsTheme';
 import useScreenSize from '../../../hooks/useScreenSize';
 import useStickyTopValues from '../../../hooks/useStickyTopValues';
 import { InstruqtContext } from '../../Instruqt/instruqt-context';
-import { elementZIndex } from '../../../utils/dynamically-set-z-index';
 import { HeaderContext } from '../../Header/header-context';
 import ProgressBar from './components/PageIndicators';
 import CloseButton from './components/CloseButton';
@@ -61,13 +59,11 @@ const FeedbackCard = ({ isOpen, children }) => {
 
   const onClose = () => {
     abandon();
-    // reset the z-index set by the screenshot button in ScreenshotButton.js
-    elementZIndex.setZIndex('.widgets', theme.zIndexes.header);
   };
 
   return (
     isOpen && (
-      <FloatingContainer darkMode={darkMode} top={topBuffer} id={feedbackId} hasOpenLabDrawer={isLabOpen}>
+      <FloatingContainer darkMode={darkMode} top={topBuffer} hasOpenLabDrawer={isLabOpen}>
         <Card>
           <CloseButton onClick={onClose} />
           <ProgressBar />
--- a/src/components/Widgets/FeedbackWidget/FeedbackCard.js
+++ b/src/components/Widgets/FeedbackWidget/FeedbackCard.js
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useMemo, useRef } from 'react';
+import React, { useContext, useMemo } from 'react';
 import styled from '@emotion/styled';
 import LeafygreenCard from '@leafygreen-ui/card';
 import { useDarkMode } from '@leafygreen-ui/leafygreen-provider';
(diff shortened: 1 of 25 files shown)


src/components/Widgets/FeedbackWidget/context.js
```diff
@@ -126,6 +126,13 @@ export function FeedbackProvider({ page, test = {}, ...props }) {
     selectInitialRating,
   };
 
+  // reset feedback when route changes
+  useEffect(() => {
+    return globalHistory.listen((location, action) => {
+      abandon();
+    });
+  }, [abandon]);
```
rayangler, line 134: Is there a benefit to using this listener here over checking if `page.url` changes in the dependency array? Doesn't seem like `location` or `action` are being used so I'm just wondering if it would be simpler to check if the page url changes

src/styles/global-dark-mode.css
```diff
@@ -31,9 +21,20 @@ body {
   --link-color-primary: var(--blue-light1);
   --link-font-weight: 700;
 
-  --background-color-primary: var(--gray-dark3);
+  /* TODO: check this is correct color */
```
rayangler, line 24: (Just checking) Is this something we still need to do?

rayangler (commented):
Thanks for working on this!

I know that this PR is for the desktop, so please let me know if this is intentional for now: Noticed on smaller screen sizes, after the feedback button wraps to the next line, the feedback card ends up being cut off on the side due to positioning (see screenshot below). On mobile screen sizes, the button causes the feedback form to appear but it seems to be hidden.

On tablet screen sizes:
Screen Shot 2024-08-06 at 3 55 48 PM

src/styles/global-dark-mode.css
seungpark: good catch! gonna remove this comment. wanted to confirm we getting the colors right, but black (#001E2B) and white(#FFFFFF). 

also wanted to note that we could use this in the `--wayfinding-option-bg-color` https://github.com/mongodb/snooty/pull/1192/files#diff-a17fe66e417a0ed00243067e348c14d5f769b6cd74713d9df92c6e15c135e182R27

since it is the same as `--background-color-primary`

src/components/Widgets/FeedbackWidget/context.js
seungpark: hm how do you feel about using `  const { href } = useLocation();` vs prop dependency? seems more convenient than prop drilling

seungpark:
> Thanks for working on this!
> 
> I know that this PR is for the desktop, so please let me know if this is intentional for now: Noticed on smaller screen sizes, after the feedback button wraps to the next line, the feedback card ends up being cut off on the side due to positioning (see screenshot below). On mobile screen sizes, the button causes the feedback form to appear but it seems to be hidden.
> 

Will follow up for mobile! Dropped in a ticket for the [mobile action bar](https://jira.mongodb.org/browse/DOP-4908) specifically, but will file another one for the Feedback full viewport view

src/components/Widgets/FeedbackWidget/components/ScreenshotButton.js
```diff
@@ -200,20 +197,18 @@ const ScreenshotButton = ({ size = 'default', ...props }) => {
     setIsScreenshotButtonClicked(false);
     setCurrElemState(null);
     setScreenshotTaken(false);
-    elementZIndex.resetZIndex('.widgets');
   };
 
   const handleDOMElementClick = (e) => {
     e.preventDefault();
 
     //conditionally set the z-index on the widget card when the screenshot is clicked
```
mmeigs, line 205: Remove comment if the line of code is removed?

src/components/Widgets/FeedbackWidget/components/StarRating.js
```diff
@@ -166,7 +181,14 @@ const StarRating = ({ className, handleRatingSelection = () => {}, editable = tr
           );
         })}
       </Layout>
-    )
+      {showCaption && (
+        <Body baseFontSize={13} className={cx(captionStyling)}>
+          Poor
+          <StyledArrow>&#10230;</StyledArrow>
```
mmeigs, line 187: Although I like yours, the Figma's arrow seems to be a lot bigger.

mmeigs (commented):
I know it's annoying, but can you also change layouts/preview-layout.js to match this so the Gatsby sites will also have the ActionBar?

src/components/Widgets/FeedbackWidget/components/ScreenshotButton.js
seungpark: good catch!

seungpark:
> I know it's annoying, but can you also change layouts/preview-layout.js to match this so the Gatsby sites will also have the ActionBar?

https://preview-mongodbseungpark.gatsbyjs.io/drivers/DOP-4817/
pushed up a few changes to move the feedback button!

src/components/Widgets/FeedbackWidget/components/StarRating.js
seungpark: also got approved for the unicode 👍

src/components/Widgets/FeedbackWidget/components/StarRating.js
```diff
@@ -170,15 +133,7 @@ const StarRating = ({ className, handleRatingSelection = () => {}, editable = tr
...
-          return (
-            <Star
-              key={ratingValue}
-              ratingValue={ratingValue}
-              isHighlighted={isHighlighted}
-              triggerEnabled={editable}
-              {...eventProps}
-            />
-          );
+          return <Star key={ratingValue} isHighlighted={isHighlighted} triggerEnabled={editable} {...eventProps} />;
```
rayangler, line 136: We should be able to remove `triggerEnabled` here as well

src/components/Widgets/FeedbackWidget/FeedbackCard.js
```diff
@@ -61,15 +61,33 @@ const FeedbackCard = ({ isOpen, children }) => {
     abandon();
   };
 
+  const ref = useRef(null);
+
+  // Effect to detect click outside of FeedbackCard
+  useEffect(() => {
```
rayangler, line 67: (Non-blocking suggestion) We have an unused hook called [useClickOutside](https://github.com/mongodb/snooty/blob/e1619ae1a89c66fbdafe488e4a5cc9f714bd10d1/src/hooks/use-click-outside.js). Do you think it's worth repurposing the hook for this component by passing in `ref` and `abandon`?

(Behavior question) I noticed that if you click the feedback button, and click it again, this `useEffect` causes the card to disappear but then quickly reappears due to the button click, causing flashing. Is that behavior intentional or do we want to keep the previous behavior of closing the widget card?

rayangler:
Wanted to flag this gap with the side nav. Is this something we should fix here or is that work slated for a different PR?

Screen Shot 2024-08-07 at 4 04 36 PM

seungpark:
> Wanted to flag this gap with the side nav. Is this something we should fix here or is that work slated for a different PR?
> 
> Screen Shot 2024-08-07 at 4 04 36 PM

this will be captured [here](https://jira.mongodb.org/browse/DOP-4897)! yay for feature branches. the target branch should be merged once all the issues in the epic are closed

src/components/Widgets/FeedbackWidget/FeedbackCard.js
seungpark: good catch on both! resorting to make a container which has both form and button and calls useClickOutside to detect outside, and keep the button to both open close. incoming push

rayangler: Sorry, more context in case it helps with debugging/refactoring: With this click outside hook, seems like the screenshot functionality breaks (due to intentionally clicking outside 😅). Unsure if this requires a more custom implementation of clicking outside than the existing hook, so feel free to ignore the suggestion

seungpark: another good catch! moved the `screenshotTaken` state into the context so we can use it in the container. the container can check for this before calling the abandon fn!

src/components/Widgets/FeedbackWidget/context.js
rayangler: Yup, that's fine with me. Feel free to let me know if `globalHistory.listen` is better for the use case though! I was genuinely curious about it 🙂

rayangler (approved):

PR #1208, opened 2024-08-27
DOP-4686: Implement Method Selector

rayangler:
### Stories/Links:

DOP-4686
[Example rST](https://github.com/rayangler/cloud-docs/blob/ace163b51c2f157f38a6d044d2a5faea63b287d3/source/testing-method-selector.txt)
[Designs](https://www.figma.com/design/7cR6eWtFvKTnJIPL8ZsTaB/Content-Template-Projects?node-id=3473-22389&t=RNnOtvg64NtvtUHi-0)
[Related parser PR](https://github.com/mongodb/snooty-parser/pull/614)

### Current Behavior:

[Atlas docs](https://www.mongodb.com/docs/atlas/) (majority of changes are net new)

### Staging Links:

* [Example with 6 options](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/sandbox/cloud-docs/raymund.rodriguez/DOP-4686/testing-method-selector/index.html)
* [Example with 3 options](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/sandbox/cloud-docs/raymund.rodriguez/DOP-4686/testing-method-selector-3-options/index.html)

### Notes:

* When the side nav is open at around 1025px, the content column is too narrow and causes the method options to wrap. After speaking to design, we've agreed that allowing the options to wrap while conditionally having the green line below it present only when the options are in a single row was okay to have.
* Padding and widths of options have been reduced from the original designs to accommodate options in desktop screen size.
* **Known issue:** The "On This Page" component isn't populated despite having headings present on the page. This seems like a bug with the parser, and I'll create a ticket to handle fixing this separately. I don't think this should be a blocker for this PR considering the "On This Page" component will have the same issue with showing headings from unselected tabs, which is a less than ideal UX either way.

### README updates


src/templates/document.js
```diff
@@ -10,16 +10,17 @@ import TabSelectors from '../components/Tabs/TabSelectors';
 import useSnootyMetadata from '../utils/use-snooty-metadata';
 import AssociatedVersionSelector from '../components/AssociatedVersionSelector';
 
+const MAX_CONTENT_WIDTH = '775px';
+
 const DocumentContainer = styled('div')`
   display: grid;
   grid-template-areas: 'main right';
-  grid-template-columns: minmax(0, auto) 1fr;
+  grid-template-columns: minmax(0, ${MAX_CONTENT_WIDTH}) 1fr;
```
rayangler, line 18: Changing this because `auto` was causing the main content column to shrink when the content itself wasn't wide enough. Using the same max width here allows the content to try being the max possible width when possible by default.

This doesn't change much on a typical docs page since they're typically full of content, but this is useful when developing new components that depend on as much content width as possible (such as the Method Selector and its options)

src/templates/document.js
```diff
@@ -10,16 +10,17 @@ import TabSelectors from '../components/Tabs/TabSelectors';
...
 `;
 
 const StyledMainColumn = styled(MainColumn)`
   grid-area: main;
-  max-width: 775px;
-  overflow-x: auto;
```
rayangler, line 22: Changing this because `overflow-x` was causing the box shadow of the radio boxes to appear cut off. (Also, I don't think it's generally good practice to have main page content be horizontally scrollable)

src/components/MethodSelector/MethodSelector.js
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * Helper function to determine if options will wrap or not
+ * @param {number} count
+ * @returns {string}
+ */
+const whenWrappingDesktop = (count) => {
+  // Hardcoding this will make it easier to have accurate media queries, without relying on client-side JS and potentially
+  // creating content layout shifts.
+  // Only wraps when there's 4+ options are found
```
rayangler, line 21: I'm a bit iffy about this implementation because it's not very resilient in the event that we change our site's widths, padding, etc., but it was the only way I can think of to easily get the green line below the options to appear based on wrapping without creating an unnecessary content layout shift or flash on initial load.

I'm thinking we can keep this for now and revisit later, if needed. Thoughts on this would be much appreciated!

src/templates/document.js
seungpark: this is fine for me, since its defining a min/max range. wondering when the `content itself wasn't wide enough`, can you give an example here?

src/components/MethodSelector/MethodOptionContent.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { css, cx } from '@leafygreen-ui/emotion';
+import ComponentFactory from '../ComponentFactory';
+import { theme } from '../../theme/docsTheme';
+import MethodDescription from './MethodDescription';
+
+const METHOD_DESCRIPTION_NAME = 'method-description';
+
+const displayStyle = (isSelectedOption) => css`
+  ${!isSelectedOption && 'display: none;'}
```
seungpark, line 10: yay for translations and SEO 👍

src/components/MethodSelector/MethodSelector.js
seungpark: at the moment i'm not sure how to do this purely with css (without a useEffect and JS that appears after initial load) but it seems a little inconsistent with rows and number of elements. 
ie. if i delete the DOM for one of the options and shrink the page, it seems to not register for all widths:
image

seungpark: thinking off the top of my head - how about a Grid <container for radio boxes to maintain the box sizing> inside a Flexbox <container for said Grid and the line <hr>> . 

the rows after first of grid can have a bottom margin, and the overflow of flexbox can be hidden. WDYT?

seungpark (commented):
this looks great! brainstorming a way for the underline below, we can discuss more in depth if needed!

src/templates/document.js
rayangler: Turns out the new method selector component actually circumvents this with its usage of `max-width: 100vw;`, but without it during development, I was running into issues where changing methods with varying content widths would cause the main content column to change width. This is most noticeable when the content is only a few words or is broken up into smaller widths.

I can look to see if we have more realistic examples on prod somewhere and add them to this thread if they exist, but here's a video with `max-width` unset in the selector, and `auto` used in the `document` template. Notice how the widths change because of the widths of the content, despite there being more than enough space for the content column itself.

https://github.com/user-attachments/assets/7675f752-a818-4e9d-a34d-0d21f5d53167

src/components/MethodSelector/MethodSelector.js
rayangler: Thanks for the inspiration! I've done something similar to this and updated the staging links accordingly. 

For posterity, the triangle needs a lot more calculations to ensure that it's grouped together with the `hr` line, but it seems more flexible in general. I also had to add a few pixels of padding around the grid of options to ensure the box shadow of the selected option doesn't get cut off.

Let me know if this seems like a better alternative

src/components/MethodSelector/MethodSelector.js
```diff
@@ -62,53 +56,56 @@ const radioBoxStyle = (isSelected, count, isSidenavCollapsed) => css`
...
+ * starts from the lefthand side, relative to an element that is the full width of the content column.
+ * @param {number} count
+ * @param {number} selectedIdx
+ * @returns {string}
+ */
+const trianglePosCalc = (count, selectedIdx) => {
+  const numGaps = `${count - 1} * ${OPTIONS_GAP}`;
+  const fullHorizontalPadding = `${OPTIONS_GRID_PADDING} * 2`;
+  const widthOfOneOption = `(100% - ${fullHorizontalPadding} - ${numGaps}) / ${count}`;
+  return `calc(${OPTIONS_GRID_PADDING} + (${widthOfOneOption} / 2) + ${selectedIdx} * (${OPTIONS_GAP} + ${widthOfOneOption}))`;
```
seungpark, line 80: [minor nitpick] since we are going from left -> right, can we order the elements like so. ie:
`return `calc(${OPTIONS_GRID_PADDING} + ${WIDTHS_OF_PRECED_OPTIONS} + ${widthOfOneOption / 2})`;

seungpark (approved):
this LGTM! left two minor comments below regarding method selector description's margin and var naming for caret position calculation

src/components/MethodSelector/MethodDescription.js
```diff
@@ -0,0 +1,36 @@
...
+import React from 'react';
+import { theme } from '../../theme/docsTheme';
+import ComponentFactory from '../ComponentFactory';
+import TabSelectors from '../Tabs/TabSelectors';
+
+const containerStyle = css`
+  font-size: ${theme.fontSize.small};
+  margin-bottom: ${theme.size.large};
+
+  * {
```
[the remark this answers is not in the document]
rayangler: Thanks for pointing this out. The top margin is good to stay as is. Lara requested this [in thread](https://mongodb.slack.com/archives/C07DW1R200Z/p1724785491992759?thread_ts=1724703601.490769&cid=C07DW1R200Z). It looks like the language selector is expected to have a 32px though, so I'll look into that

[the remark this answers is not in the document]
rayangler: Updated the language selector accordingly.

src/components/MethodSelector/MethodSelector.js
rayangler: Done

src/templates/document.js
rayangler: After taking another look at this, seems like the margins are taken into account with this minmax issue. This causes other pages' content to not wrap at 775px. Gonna spend some time trying to ensure the method selector can help stretch out the max width of the content column

mmeigs (approved):
Looks really good!

PR #1189, opened 2024-08-02
DOP-4616: Search on every page

seungpark:
[DOP-4616](https://jira.mongodb.org/browse/DOP-4616)

Feature branch for dark mode + universal search.
This branch should only have merge commits from PR's that scope out smaller tasks.

These have been staged with the env flag `GATSBY_ENABLE_DARK_MODE=true`:

[cloud docs](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/seung.park/DOP-4616/index.html)
[landing](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/seung.park/DOP-4616/index.html)
[404](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs-404/seung.park/DOP-4616/index.html)
[app services](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/atlas-app-services/seung.park/DOP-4616/index.html)
[server](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/v7.3/docs/seung.park/DOP-4616/index.html)

Refer to [this spreadsheet](https://docs.google.com/spreadsheets/d/1S9YtG_mrZUPEjFEocRZYKIoCGxzN5FNz4wGK7XVa7vo/edit?gid=1022791337#gid=1022791337) for bug tracking

PR #1209, opened 2024-08-28
DOP-4909: feedback form works on mobile with new action bar

seungpark:
### Stories/Links:

DOP-4909

- Marking the destination branch as DOP-4897 as mobile behavior was updated on this branch
- Feedback card should take up the whole screen on mobile as per [figma design](https://www.figma.com/design/7cR6eWtFvKTnJIPL8ZsTaB/Content-Template-Projects?node-id=7488-6545&t=ZSe2b6fwRzBrQhC8-0)

### Staging Links:

[landing staged](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/seung.park/DOP-4909/index.html)

### Notes:
- Did not stage other projects as it seems pretty standard on all pages. Feel free to request other staging links if needed

### README updates


src/components/Widgets/FeedbackWidget/FeedbackCard.js
```diff
@@ -16,9 +16,7 @@ import useNoScroll from './hooks/useNoScroll';
 const CardContainer = styled.div`
   @media ${theme.screenSize.upToSmall} {
     padding-top: ${({ top }) => `${top}`};
-    right: 0;
-    top: 0;
-    bottom: 60px;
+    height: 100%;
```
rayangler, line 19: Looks like this height is causing the card container to extend well beyond the viewport (maybe because it's in a portal? not sure). This makes it so that scrolling further down the page and clicking the feedback menu results in a blank screen, with the actual feedback items further up

rayangler (commented):
Wanted to flag a couple of things:
* (Nit) Should the PR title say "from" or "form"?
* Opening the mobile side nav and then clicking on a rating causes it to disappear immediately. Seems like the form is still in the HTML but it's hidden due to the CSS. Is this something to be addressed in this PR or separately?

src/components/Widgets/FeedbackWidget/FeedbackCard.js
seungpark: thanks for this catch! changing to fixed position

seungpark:
> Wanted to flag a couple of things:
> 
> * (Nit) Should the PR title say "from" or "form"?
> * Opening the mobile side nav and then clicking on a rating causes it to disappear immediately. Seems like the form is still in the HTML but it's hidden due to the CSS. Is this something to be addressed in this PR or separately?

Thanks for the behavioral catch. This was due to a [useClickOutside](https://github.com/mongodb/snooty/pull/1189/files#diff-a4d0df6952218a4f6afbca3fa05ba9db1228e7a4a03391d9aec21b40352a870dR14) previously added for closing feedback when user clicks outside of the desktop version. fixed and addressed! 

And updated the title :)

src/components/Widgets/FeedbackWidget/FeedbackForm.js
```diff
@@ -23,6 +23,14 @@ const formStyle = css`
   position: absolute;
   right: 0;
   margin-top: ${theme.size.tiny};
+
+  @media ${theme.screenSize.upToSmall} {
+    position: fixed;
+    top: 0;
+    bottom: 0;
+    left: 0;
+    right: 0;
```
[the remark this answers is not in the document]
seungpark: for people with square windows! done

PR #1215, opened 2024-08-30
No ticket: Realm product-landing template headers - more specific

mmeigs:
### Context

After fixing the header text color on the Realm landing page, I realized that there are many uses of the `product-landing` template within the Realm repo. This means that it was changing all the text colors in those pages rather than just the one with the light color hero on it. 

This PR makes this change more specific than just being in the Realm repo and product-landing template. It requires a hard-coded fileid to identify which page has the light color hero image.

In the future, if we have the time maybe we add an option to the image directive in the parser to specify which hero images are light colored? This is not a wide-spread problem, but I also hate the hard-coded solution long-term. 

### Current Behavior:

[Old staging with issue (check any of the sdk pages to see the issue)](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOP-4942-dark-icons-realm-master/realm/matt.meigs/realm-dark-mode-hero-text/index.html)

### Staging Links:

[Fixed staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOP-4942-dark-icons-realm-master/realm/matt.meigs/fix-realm-headers/index.html)

### README updates

### DIFF
--- a/src/templates/product-landing.js
+++ b/src/templates/product-landing.js
@@ -38,18 +38,22 @@ const Wrapper = styled('main')`
   }
 
   // realm PLP has full width p
-  ${({ isRealm }) =>
+  ${({ isRealm, hasLightHero }) =>
     isRealm &&
     `
+    ${
+      hasLightHero &&
+      `
+      @media ${theme.screenSize.mediumAndUp} {
+        h1 {
+          color: ${palette.black};
+        }
 
-    @media ${theme.screenSize.mediumAndUp} {
-      h1 {
-        color: ${palette.black};
+        .introduction > p:first-of-type {
+          color: ${palette.black};
+        }
       }
-
-      .introduction > p:first-of-type {
-        color: ${palette.black};
-      }
+    `
     }
 
     section > p {
@@ -176,6 +180,8 @@ const Wrapper = styled('main')`
   }
 `;
 
+const LIGHT_HERO_PAGES = ['index.txt'];
+
 const ProductLanding = ({ children, data: { page } }) => {
   const { project } = useSnootyMetadata();
   const useHero = ['guides', 'realm'].includes(project);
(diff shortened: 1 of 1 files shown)


src/templates/product-landing.js
```diff
@@ -176,13 +180,16 @@ const Wrapper = styled('main')`
...
+const LIGHT_HERO_PAGES = ['index.txt'];
+
 const ProductLanding = ({ children, data: { page } }) => {
   const { project } = useSnootyMetadata();
   const useHero = ['guides', 'realm'].includes(project);
   const isGuides = project === 'guides';
   const isRealm = project === 'realm';
   const pageOptions = page?.ast?.options;
   const hasMaxWidthParagraphs = ['', 'true'].includes(pageOptions?.['pl-max-width-paragraphs']);
+  const hasLightHero = LIGHT_HERO_PAGES.includes(page?.ast?.fileid);
```
mayaraman19, line 192: cmiiw, but this would be true even for index pages that don't have light hero images right (i.e. if they come from another project)? which wouldn't end up getting used outside of realm anyway but can be a bit confusing, do we want to filter by realm beforehand and/or somehow combine fileid and project so this is true only when the correct file from the correct project is selected?

mayaraman19 (commented):
good catch! also wondering if this is something we could talk to other teams about (design and perhaps content idk) for a standard for hero images? like if all hero images were dark? idk...

src/templates/product-landing.js
mmeigs: You are very right. I defined this badly because I was filtering later on. Fixing now.

rayangler (approved):
Looks good!

> maybe we add an option to the image directive in the parser to specify which hero images are light colored?

This sounds like a good long-term solution, especially if we end up having more instances of this.

mmeigs:
> good catch! also wondering if this is something we could talk to other teams about (design and perhaps content idk) for a standard for hero images? like if all hero images were dark? idk...

Really good thing to keep in mind. This should be the only one and design this time just mentioned not wanting to change the image back and forth. But that would be nice...

PR #1216, opened 2024-09-03
DOP-4955: OpenAPI Preview page - disable Dark Mode

mmeigs:
### Stories/Links:

DOP-4955

### Current Behavior:

[Preview Page](https://www.mongodb.com/docs/openapi/preview/)

### Staging Links:

[Preview Page](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOP-4873-dark-icons/landing/matt.meigs/DOP-4955-dark-openapi-preview/openapi/preview/index.html)

### Notes:

This PR disables Dark Mode on the OpenAPI preview page. It does it in such a way that it does not affect localStorage or the user's saved preference at all so that when navigating between the preview page and any other docs page, the user's dark mode preference will not be affected by the switch. 

Also, I have cleaned up some unused props.

### README updates

### DIFF
--- a/src/components/OpenAPI/index.js
+++ b/src/components/OpenAPI/index.js
@@ -128,7 +128,7 @@ const LoadingWidget = ({ className }) => (
 const MenuTitleContainer = ({ pageTitle }) => {
   return (
     <>
-      <DocsHomeButton />
+      <DocsHomeButton hideDarkModeToggle={true} />
       <Border />
       <MenuTitle>{pageTitle}</MenuTitle>
     </>
--- a/src/components/RootProvider.js
+++ b/src/components/RootProvider.js
@@ -9,7 +9,7 @@ import { ContentsProvider } from './Contents/contents-context';
 
 const RootProvider = ({ children, headingNodes, slug, repoBranches, remoteMetadata }) => {
   return (
-    <DarkModeContextProvider>
+    <DarkModeContextProvider slug={slug}>
       <ContentsProvider headingNodes={headingNodes}>
         <HeaderContextProvider>
           <VersionContextProvider repoBranches={repoBranches} slug={slug}>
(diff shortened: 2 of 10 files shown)


seungpark:
it seems like the [HEAD script is still running](https://github.com/mongodb/snooty/blob/main/gatsby-ssr.js#L62) and therefore the page is loading in dark mode if dark mode was previously set. visiting the staging link while having dark mode on in a different staging link leads to the staging link with dark mode on

src/components/RootProvider.js
```diff
@@ -9,7 +9,7 @@ import { ContentsProvider } from './Contents/contents-context';
 
 const RootProvider = ({ children, headingNodes, slug, repoBranches, remoteMetadata }) => {
   return (
-    <DarkModeContextProvider>
+    <DarkModeContextProvider slug={slug}>
```
mayaraman19, line 12: if you want to avoid prop drilling, you can instead use `PageContext` in `DarkModeContextProvider`

src/components/OpenAPI/index.js
```diff
@@ -128,7 +128,7 @@ const LoadingWidget = ({ className }) => (
 const MenuTitleContainer = ({ pageTitle }) => {
   return (
     <>
-      <DocsHomeButton />
+      <DocsHomeButton hideDarkModeToggle={true} />
```
mayaraman19, line 131: also, how will disabling dark mode work with the action bar? or is that not going to appear on this site?

mmeigs: The dark mode toggle is hidden on the action bar for this page - you can see one file above!

src/components/RootProvider.js
mmeigs: Ah, sadly the PageContextProvider is not available here. It's a child of this.

mmeigs:
> it seems like the [HEAD script is still running](https://github.com/mongodb/snooty/blob/main/gatsby-ssr.js#L62) and therefore the page is loading in dark mode if dark mode was previously set. visiting the staging link while having dark mode on in a different staging link leads to the staging link with dark mode on

Offline discussion ended with adding a check within the gatsby-ssr head script to further enforce light mode on the OpenAPI preview page

seungpark (approved):
🕶️

mayaraman19 (approved):

PR #1242, opened 2024-09-18
0px if not browser

mayaraman19:
### Stories/Links:

DOP-5000

Unfortunately I can't really replicate this behavior in local dev, but in preprd you see that `scrollY = undefined` on initial load which causes the bottom 3 links to not show up until you scroll. This PR sets it to 0 when it's being built so it is not undefined on initial load. Check out the `<div>` with id `docs-side-nav-container`. 

### Current Behavior:

[Landing preprd](https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/)

### Staging Links:

[Landing Staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOP-4683/landing/maya/DOP-5000/index.html)

### README updates

### DIFF
--- a/src/hooks/useViewport.js
+++ b/src/hooks/useViewport.js
@@ -10,7 +10,10 @@ export function getViewport() {
         scrollY: Math.max(window.pageYOffset, document.documentElement.scrollTop, document.body.scrollTop || 0),
         scrollX: Math.max(window.pageXOffset, document.documentElement.scrollLeft, document.body.scrollLeft || 0),
       }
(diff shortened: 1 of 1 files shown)


seungpark (approved):
this works for me :+1: i was thinking in the sidenav where [we call this](https://github.com/mongodb/snooty/blob/main/src/components/Sidenav/Sidenav.js#L255) but this is actually better 👍 thank you!

PR #1254, opened 2024-09-24
DOP-4883: Menu not closing after selection on Mobile 

biancalaube:
### Stories/Links:

DOP-4883

### Current Behavior:

[Current Behavior](https://www.mongodb.com/docs/atlas/tutorial/create-atlas-account/)

* If you are on mobile you have to select the top docs menu button to get rid of the menu/sidenav once you select a site

### Staging Links:

[Stagging Behavior](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/bianca.laube/DOP-4883/index.html)
* Now when you select on a link (that is not a twisty) it should automatically get rid of the menu/sidenav (on mobile)

### README updates


mayaraman19 (changes requested):
Nice work, thanks for picking this up! I think there are a couple things that need a bit of tweaking:
- I actually think the previous behavior you showed in Slack was what we were looking for, where clicking on the text of the link opens the page and hides the Sidenav (even if there are children)
- I think because we are not using the `setIsOpen(!isOpen)` anymore, if you click on an opened caret (on Desktop) it doesn't close anymore (for example, navigate to a page with children and then click on the sidenav item (text or twisty) for the current page you're on - in prod it closes back up). You will also see one of the tests failing because of this! If we can preserve the current behavior of the dropdown so that you can open and close the side nav item even when you're on the page that would be great!

biancalaube:
@mayaraman19 

Thanks for the comments! Both things should be fixed by the 2 latest commits. Please review again and let me know if you see more issues :)

src/components/Sidenav/TOCNode.js
```diff
@@ -82,6 +82,11 @@ const TOCNode = ({ activeSection, handleClick, level = BASE_NODE_LEVEL, node, pa
     setIsOpen(!isOpen);
   };
 
+  const clickSideNav = (handleClick) => {
+    setIsOpen(!isOpen);
```
mayaraman19, line 86: oh no, now the carets/"twisties" are being treated like text in mobile sidenav in that if you click them they open the page and hide the sidenav, instead of just opening the caret to see the children dropdown. I think this is probably interfering with [onCaretClick](https://github.com/mongodb/snooty/pull/1254/files#diff-2d7a96ba34af02c762677b1fb55a5211639cdc8864420ff53db587e367a306f8R80-R83). Lmk if you want to discuss this more!

mayaraman19 (commented):
Hey Bianca, thanks for addressing the feedback! I did leave one more comment below!

src/components/Sidenav/TOCNode.js
seungpark: agreed! i think we can leave the behaviors as is, and actually add a listener on location change that closes the mobile menu. ie something like [this pattern](https://stackoverflow.com/questions/46244279/how-to-close-navigation-menu-on-route-change)

seungpark (changes requested):
left a suggestion for closing the menu on location change!

src/components/Sidenav/TOCNode.js
biancalaube: @seungpark , Im a little confused on how the link you sent is suppose to fix my issue. Sould I not be using the handleclick() anymore that I passed in from sidenav.js ? And rewrite that function in the tocnode.js file.

seungpark: i think this is more of a `set state to close menu on location change` than on sidenav button click (which would trigger a location change). 

> we can leave the behaviors as is
as in leave the current click handlers, and add a new effect in the component to listen for location changes

biancalaube: @seungpark and @mayaraman19 . Thanks for the advice I made a change that I think changes this problem. Please let me know if you see anything else.

package-lock.json
seungpark, line 1: not sure why we have changes here! there shouldnt be any diff if we merge master into your branch

src/components/Sidenav/Sidenav.js
```diff
@@ -203,6 +205,11 @@ const Sidenav = ({ chapters, guides, page, pageTitle, repoBranches, slug, eol })
     setHideMobile(true);
   }, [setHideMobile]);
 
+  // close navigation panel on mobile screen, but leaves open if they click on a twsity
```
seungpark, line 208: minor typo in comment!

seungpark (approved):
this LGTM! minor comments below. please resolve package (no diffs!) before merge!

mayaraman19 (approved):
nice! glad you got the path prefix thing figured out as well. looks good after seung's comments have been addressed

seungpark:
discussed offline of the package diff. this was failing in main branch so this actually will resolve this !

PR #1259, opened 2024-10-01
DOP-5026 and DOP-5045

seungpark:
### Stories/Links:

DOP-5026 - Disable navigation in search menu when chatbot is hidden
DOP-5045 - Initialize conversation on focus in search input

### Current Behavior:

[on prod, conversation is not initialized until search menu is shown](https://www.mongodb.com/docs/)

No network calls to `https://knowledge.mongodb.com/api/v1/conversations` until Menu is shown:
image

### Staging Links:

The netlify link [below](https://github.com/mongodb/snooty/pull/1259#:~:text=https%3A//app.netlify.com/sites/mongodb%2Dsnooty/deploys/66fc6bbac08c5e0008fafd07)!

[docs qa](https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/DOP-5026/) - must be connected to VPN to connect to staging chatbot API. This will fail (gracefully) and not show the chatbot option, as well as disabling navigation via keyboard (DOP-5026). Key difference is that the request is being made when the searchbar is focused

Conversation being initialized on focus:
image

### Notes:
- Removed usage of [useImperativeHandle](https://react.dev/reference/react/useImperativeHandle) in favor of passing props from parent to lazy loaded child.
- Add a state to parent component of SearchInput to keep track of chatbot availability, to prevent keyboard navigation when chatbot conversation fails
- Moved chatbot conversation init function to onFocus - Focusing updates the isFocused state, which the lazy loaded SearchInput watches to init a conversation.
- In development mode, failed chatbot messages will throw an uncaught promise error ([EAI-555](https://jira.mongodb.org/browse/EAI-555))

### To test with a [local chatbot server](https://github.com/mongodb/chatbot):
- You will need .env files to run local chatbot - reach out for these configs
- From repo root `npm install` 
-  `npm run build`
- `npm run dev:server` and it should be running on port 3000

### README updates

### DIFF
--- a/src/components/ActionBar/SearchInput.js
+++ b/src/components/ActionBar/SearchInput.js
@@ -207,8 +207,6 @@ const SearchInput = ({ className, slug }) => {
   const CHATBOT_SERVER_BASE_URL = ['dotcomprd', 'production'].includes(metadata?.snootyEnv)
     ? 'https://knowledge.mongodb.com/api/v1'
     : 'https://knowledge.staging.corp.mongodb.com/api/v1';
-  // TODO: revert. for local testing
-  // const CHATBOT_SERVER_BASE_URL = 'http://localhost:3000/api/v1';
 
   return (
     <StyledInputContainer
--- a/src/components/ActionBar/SearchInput.js
+++ b/src/components/ActionBar/SearchInput.js
@@ -1,4 +1,4 @@
-import React, { lazy, useCallback, useEffect, useRef, useState } from 'react';
+import React, { lazy, useCallback, useEffect, useReducer, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import { useLocation } from '@gatsbyjs/reach-router';
 import { css, cx } from '@leafygreen-ui/emotion';
@@ -50,6 +50,11 @@ export const SEARCH_SUGGESTIONS = [
   },
 ];
 
+function chatbotAvailReducer(state, action) {
+  if (state) return state;
+  return action;
+}
+
 const SearchInput = ({ className, slug }) => {
   const [searchValue, setSearchValue] = useState('');
   const [isOpen, setIsOpen] = useState(false);
@@ -64,7 +69,7 @@ const SearchInput = ({ className, slug }) => {
   const [mobileSearchActive, setMobileSearchActive] = useState(false);
   const { search } = useLocation();
   const locale = getCurrLocale();
-  const [chatbotAvail, setChatbotAvail] = useState(false);
+  const [chatbotAvail, setChatbotAvail] = useReducer(chatbotAvailReducer, false);
   const isEnglish = locale === 'en-us';
 
   useBackdropClick(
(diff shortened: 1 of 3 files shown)


src/components/ActionBar/SearchInput.js
```diff
@@ -62,6 +69,7 @@ const SearchInput = ({ className, slug }) => {
   const [mobileSearchActive, setMobileSearchActive] = useState(false);
   const { search } = useLocation();
   const locale = getCurrLocale();
+  const [chatbotAvail, setChatbotAvail] = useReducer(chatbotAvailReducer, false);
```
branberry, line 72: What's the benefit of `useReducer` as opposed to `useState` here? Looks like the state is a boolean which would work with `useState`

branberry, line 72: NIT: `isChatbotAvailable` would be more clear

src/components/ActionBar/SearchInput.js
```diff
@@ -50,9 +50,16 @@ export const SEARCH_SUGGESTIONS = [
   },
 ];
 
+function chatbotAvailReducer(state, action) {
+  if (state) return state;
```
branberry, line 54: If I understand correctly, once we set this to `true`, we never update the state again essentially. It'll always return `true`. Is that correct? Is that behavior intended?

src/components/ActionBar/SearchMenu.js
```diff
@@ -37,52 +47,56 @@ const SearchMenu = forwardRef(function SearchMenu({ searchValue, searchBoxRef, i
...
+    setConversationInit(true);
+    conversation
+      .createConversation()
+      .then((e) => {
+        // NOTE: create conversation does not return anything
+      })
+      .catch((e) => {
+        // no errors here
+      })
+      .finally((e) => setConversationInit(false));
```
branberry, line 97: If we only need the `finally` callbacks, we can omit the `try` and `catch` callbacks

src/components/ActionBar/SearchInput.js
seungpark: that was the behavior intended but now that i think about it, we can actually make usage of [conversation properties](https://github.com/mongodb/chatbot/blob/mongodb-chatbot-ui-v0.8.1/packages/mongodb-chatbot-ui/src/useConversation.tsx#L409-L426) to properly set status. updated!

src/components/ActionBar/SearchInput.js
seungpark: was thinking of this as `once available, should stay available` but see reply to other comment. updated to be dependent on conversation properties

src/components/ActionBar/SearchMenu.js
seungpark: moved the comments and removed the try/catch blocks

src/components/ActionBar/SearchInput.js
```diff
@@ -203,6 +207,8 @@ const SearchInput = ({ className, slug }) => {
   const CHATBOT_SERVER_BASE_URL = ['dotcomprd', 'production'].includes(metadata?.snootyEnv)
     ? 'https://knowledge.mongodb.com/api/v1'
     : 'https://knowledge.staging.corp.mongodb.com/api/v1';
+  // TODO: revert. for local testing
```
seungpark, line 210: reminder to remove before merge

branberry (approved):
LGTM!

PR #1262, opened 2024-10-02
DOP-3424: "On this page" subnav incorrectly displays headings for all language tabs

mayaraman19:
### Stories/Links:

DOP-3424

This PR extends work done for [DOP-4967](https://jira.mongodb.org/browse/DOP-4967) (see [parser PR](https://github.com/mongodb/snooty-parser/pull/625) here). In the parser, we create a `selector_ids` object that reflects the AST in terms of the order of tabs/method selectors. This object looks something like this:

```
method-option: abc
children: {
     tab: def
     children: {
          tab: efg
     }
}
```

Within the ContentsContext as well as Tabs and Method Selector, we maintain an object that reflects the active selected method options and tabs/tab selectors. This object relies on the Method Selector being a top-level component, and tabs on the same page not having the same tab ID (unless they are sibling tabs that move together). These are continuously compared to ensure that the Contents reflect the selected Tab/Method Selector. This can be extended in the future to include other components as well.

We also store the value of the active tab in local storage on initial load. Previously, this was only reflected for tabs selectors.

To test, please go through the staging links and ensure that 1. the correct headings show up and 2. you can click on them and they direct to the right section of the page.

### Current Behavior:

[AWS Lambda Page](https://www.mongodb.com/docs/atlas/manage-connections-aws-lambda/)

### Staging Links:

[Tabs within Method Selector](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOP-4973-example/cloud-docs/maya/DOP-3424/test-method-selector/index.html)
[AWS Lambda (Tabs Selectors)](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOP-4973-example/cloud-docs/maya/DOP-3424/manage-connections-aws-lambda/index.html)
[Drivers Connection (should be unchanged)](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOP-4973-example/cloud-docs/maya/DOP-3424/driver-connection/index.html)
[Nested Tabs](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOP-4973-example/cloud-docs/maya/DOP-3424/test-tabs/index.html)

### README updates


rayangler (commented):
This is awesome to see working!

src/components/Tabs/index.js
```diff
@@ -111,11 +122,33 @@ const Tabs = ({ nodeData: { children, options = {} }, page, ...rest }) => {
     return activeTabIdx > -1 ? activeTabIdx : 0;
   });
 
+  if (!compare(Object.values(activeTabs), activeSelectorIds)) {
+    setActiveSelectorIds({
+      ...activeSelectorIds,
+      tab: Object.values(activeTabs),
+    });
+  }
```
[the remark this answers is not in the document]
mayaraman19: For both this and the method selector, when I try to move the statements into `useEffect` calls I run into an infinite loop 😭 reading through [documentation](https://react.dev/learn/you-might-not-need-an-effect) it seems as if it would be ideal to not put the setState call in useEffect, and this was the solution I came up with, although I missed this warning so thanks for pointing it out... however, I am at my wits' end for getting this logic to work without any issues </3

src/components/Contents/index.js
```diff
@@ -9,13 +9,25 @@ const formatTextOptions = {
   literalEnableInline: true,
 };
 
+// recursively go through selector ids defined by parser
+// everything in headingSelectorIds must be present in activeSelectorIds
+const checkNodes = (headingSelectorIds, activeSelectorIds) => {
+  if (!headingSelectorIds || JSON.stringify(headingSelectorIds) === '{}') return true;
```
mmeigs, line 15: nit: I'd prefer using the lodash method `isEmpty(headingSelectorIds)`. A little more readable. Possibly more efficient too.

src/components/Contents/index.js
```diff
@@ -9,13 +9,25 @@ const formatTextOptions = {
...
+  if (headingSelectorIds['method-option'] && headingSelectorIds['method-option'] !== activeSelectorIds.methodSelector) {
+    return false;
+  }
+  if (headingSelectorIds['tab'] && !activeSelectorIds.tab?.includes(headingSelectorIds['tab'])) {
+    return false;
```
mmeigs, line 20: nit: Could shorten both of these conditionals more with optional chaining. ie:
if (headingSelectorIds?.['method-option'] !== activeSelectorIds.methodSelector)

src/components/Contents/contents-context.js
```diff
@@ -6,23 +6,23 @@ const defaultContextValue = {
   activeHeadingId: null,
   headingNodes: [],
   showContentsComponent: true,
-  activeSelectorId: null,
-  setActiveSelectorId: () => {},
+  activeSelectorIds: {},
```
mmeigs, line 9: I'm wondering if, for clarity, we can declare in a comment the possible structure of activeSelectorIds? I understand it can have a `tab` and a `method-option` key with a array of string ids as values? These are the times I miss typescript lol

src/components/MethodSelector/MethodSelector.js
```diff
@@ -105,9 +105,14 @@ const hrStyle = css`
 const MethodSelector = ({ nodeData: { children } }) => {
   const optionCount = children.length;
   const [selectedMethod, setSelectedMethod] = useState(children[0]?.options?.id);
-  const { setActiveSelectorId } = useContext(ContentsContext);
+  const { activeSelectorIds, setActiveSelectorIds } = useContext(ContentsContext);
   const [selectedIdx, setSelectedIdx] = useState(0);
 
+  if (activeSelectorIds.methodSelector !== selectedMethod) {
+    setActiveSelectorIds({ ...activeSelectorIds, methodSelector: selectedMethod });
+  }
```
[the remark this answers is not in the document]
seungpark: something like 
```
useEffect(() => {
    if (!activeSelectorIds.methodSelector || !selectedMethod) {
      return;
    }
    if (activeSelectorIds.methodSelector !== selectedMethod) {
      setActiveSelectorIds({ ...activeSelectorIds, methodSelector: selectedMethod });
    }
  }, [activeSelectorIds, selectedMethod, setActiveSelectorIds]);
```

src/components/Tabs/index.js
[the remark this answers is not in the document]
seungpark: this can be done in the child context using the parent context, or here within the component, both with similar early returns as above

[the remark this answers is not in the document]
rayangler: Yup, I agree with that docs page that chaining setState calls via `useEffect` wouldn't be ideal. I think we'd be circumventing any chaining though if we ensure that it's called (1) only once at initial load/render whenever we grab local storage data (we can leverage the existing one, if possible), and (2) whenever a tab is selected.

src/components/MethodSelector/MethodSelector.js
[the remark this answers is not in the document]
rayangler: I was thinking of leveraging the existing `useEffect` for something like this (untested) to avoid the dependency on `selectedMethod` since it can be handled by setting the state during the onChange event like it currently does:

```js
  // Load method ID saved from last session, if applicable.
  useEffect(() => {
    const savedMethodId = getLocalValue(STORAGE_KEY);
    const validOptions = children.reduce((arr, child) => {
      arr.push(child?.options?.id);
      return arr;
    }, []);

    if (savedMethodId && validOptions.includes(savedMethodId)) {
      setSelectedMethod(savedMethodId);
      setSelectedIdx(validOptions.indexOf(savedMethodId));
      setActiveSelectorIds((activeSelectorIds) => ({ ...activeSelectorIds, methodSelector: savedMethodId }));
    } else {
      setActiveSelectorIds((activeSelectorIds) => ({ ...activeSelectorIds, methodSelector: children[0]?.options?.id }));
    }
  }, [children]);
```

src/components/Tabs/index.js
```diff
@@ -102,9 +105,13 @@ const productLandingTabContentStyling = css`
   }
 `;
 
+const activeMethodIncludesActiveTab = (activeTabValues, activeTabsSelector) =>
```
mayaraman19, line 108: this function name may be misleading since the active tab doesn't necessarily need to be within a method - could be tabs on their own, nested tabs, etc

seungpark:
Updated the branch including changes mentioned above, and a staging link is provided below to test the mix of method selectors and tabs:

https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOP-4973-example-2/cloud-docs/seung.park/DOP-3424/test-method-selector/index.html

Verify driver tabs and nested tabs and procedures here (switch to/from `Atlas UI` and procedure headings appear on the right)
https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOP-4973-example-2/cloud-docs/seung.park/DOP-3424/atlas-search/autocomplete/index.html

src/components/Tabs/index.js
seungpark: oh good catch. would it be better as `activeTabsAncestorIsActive`

src/components/Tabs/index.js
```diff
@@ -113,12 +120,46 @@ const Tabs = ({ nodeData: { children, options = {} }, page, ...rest }) => {
...
+    }
+  }, [activeSelectorIds, activeTabs]);
+
+  useEffect(() => {
+    if (!correctedSelectorIds) {
+      return;
+    }
+
+    setActiveSelectorIds(correctedSelectorIds);
+  }, [correctedSelectorIds, setActiveSelectorIds]);
```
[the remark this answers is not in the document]
seungpark: this was to clean up the dependency used in that useEffect, but can see the argument to leave it in

src/components/Tabs/index.js
mayaraman19: more something like `updateActiveTabs` since this is checking the tabs themselves, not necessarily their ancestor (might want to rename `activeTabsSelector` to `activeSelectors` or something like that)

seungpark: but this returns a boolean, does not do any updating

src/components/Tabs/index.js
```diff
@@ -113,12 +120,46 @@ const Tabs = ({ nodeData: { children, options = {} }, page, ...rest }) => {
     return activeTabIdx > -1 ? activeTabIdx : 0;
   });
 
+  const correctedSelectorIds = useMemo(() => {
```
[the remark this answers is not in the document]
seungpark: actually, moved this to the tab context, where we already have an action on `activeTabs` change. i enforced a correction each time (with the suggestion above for updating state), but prevent activeTabs from being updated if user is selecting the currently active tab to prevent the effect from running. 

PTAL at [this commit](https://github.com/mongodb/snooty/pull/1262/commits/9f25901e0f434aaa9332ebdc72ed5c4b207c03f6)

src/components/Tabs/index.js
seungpark: i've actually removed this check! we can enforce this correction in the tab context, and only if there are active tabs. and this should be enforced every time an active tab is changed. [PTAL at the latest commit ](https://github.com/mongodb/snooty/pull/1262/commits/9f25901e0f434aaa9332ebdc72ed5c4b207c03f6)

rayangler (approved):

PR #1272, opened 2024-10-07
DOP-5047: Font face of banners are 16px at load then change to 13px

biancalaube:
### Stories/Links:

DOP-5047:

### Current Behavior:

[current behaviuor](https://www.mongodb.com/docs/atlas/atlas-search/view-query-analytics/)

### Staging Links:

[staged behavior](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/atlas-app-services/bianca.laube/DOP-5047/index.html)

### Notes:

* This was a one line change but I believe the below code could be removed since it seems redundant. I left it in for now but I can remove it if need be.
```
 p,
  a {
    font-size: ${theme.fontSize.small};
  }
```
[Link to code ^](https://github.com/mongodb/snooty/blob/main/src/components/Banner/styles/bannerItemStyle.js#L26)

### README updates


src/components/Banner/styles/bannerItemStyle.js
```diff
@@ -19,7 +19,7 @@ export const baseBannerStyle = css`
   }
 
   /* Force all content to be 13px in banners */
-  font-size: ${theme.fontSize.small};
+  font-size: ${theme.fontSize.small} !important;
```
seungpark, line 22: instead of having an `!important` (this will be harder to overwrite later), can we address this in the banner component where we are using a mix of `styled` and `css`.

The preference is to use `import { css, cx } from @leafygreen-ui/emotion` since this is the same package they will be using. And for applying classes, we should be using the className prop with a `cx(css"display:block;")` kind of style. This kicks in during HTML build (versus during hydration currently) and can be verified with a build version and looking at the raw HTML file

seungpark (changes requested):

src/components/Banner/styles/bannerItemStyle.js
biancalaube: Hi Seung, I was having issues implementing your suggested fix so I did something a little different. I removed the `!important` and added the `baseFontSize` props to the Leafygreen Banner component. If this fix doesn't suffice please let me know and I can do more research into getting the "display:block" to work.

seungpark: sorry, that was a bad example. the display block was irrelevant, its the usage of leafygreen/emotion that is key ([versus current @emotion/styled](https://github.com/mongodb/snooty/blob/main/src/components/Banner/Banner.js#L4))

A good example can be [seen here](https://github.com/mongodb/snooty/blob/6f3b77bce59095cae9f9987d2ec153856675b27c/src/components/Card/index.js#L168) in Card components, where we are using the className property of the LG component, and adding a `cx(css"")` style string.

package-lock.json
seungpark, line 1: we shouldnt be getting any package-lock diffs here AFAIK

seungpark (commented):
changes LGTM. but the package diff should not be there. suggest merging main branch into yours, and running `npm ci --legacy-peer-deps`

PR #1289, opened 2024-10-21
DOP-5052: fixing link buffer's on mobile

biancalaube:
### Stories/Links:

DOP-5052

### Current Behavior:

1. on the [$merge](https://www.mongodb.com/docs/manual/reference/operator/aggregation/merge/) page, the links in the field columns in the [syntax](https://www.mongodb.com/docs/manual/reference/operator/aggregation/merge/#syntax) table
2. [replSetGetStatus](https://www.mongodb.com/docs/manual/reference/command/replSetGetStatus/#mongodb-data-replSetGetStatus) page, some lines in the [output](https://www.mongodb.com/docs/manual/reference/command/replSetGetStatus/#output) section have this behavior as well (such as electionCandidateMetrics )

### Staging Links:

[merge page](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/bianca.laube/DOP-5052/reference/operator/aggregation/merge/index.html)

[replSetGetStatus page](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/bianca.laube/DOP-5052/reference/command/replSetGetStatus/index.html)

[definition list item staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/bianca.laube/DOP-5052/reference/glossary/index.html#std-term-journal)

### Notes:
* I got rid of the buffer variable in Permalink since it was never used and always undefined when I was testing. No other file called Permalink and gave a buffer value.
* I tested locally with `banner-dev` and it seemed to work without changing any of the values, but feel free to check yourself
* Make sure the links work on mobile, tablet and full screen
* Check to make sure it didnt mess up other links

### README updates


src/components/Target.js
```diff
@@ -1,9 +1,15 @@
 import React, { useRef } from 'react';
 import PropTypes from 'prop-types';
+import { cx, css } from '@leafygreen-ui/emotion';
 import useHashAnchor from '../hooks/use-hash-anchor';
 import ComponentFactory from './ComponentFactory';
 import Permalink from './Permalink';
 
+const HeaderBuffer = css`
```
seungpark, line 8: [minor] we can name this styling with camelCase.

and similar comment as to above

src/components/Permalink.js
seungpark, line 1: can we make the HeaderBuffer an inline element? this will attempt to capture the element it is already on the line with. ie.

```
export const HeaderBuffer = styled.div`
  margin-top: ${({ bufferSpace }) => bufferSpace};
  position: absolute;
  // Add a bit of padding to help headings be more accurately set as "active" on FF and Safari
  padding-bottom: 2px;
  display: inline;
`;
```

a bonus if we can make the parent element (`headingStyle`) `position: relative;` and add a `left: 0; top: 0;` to the HeaderBuffer (so when it wraps to another line, it still calculates the margin-top from the top of the parent element.

src/theme/docsTheme.js
```diff
@@ -75,6 +75,7 @@ const header = {
   docsMobileMenuHeight: '52px',
   // used for scrolling elements into place, considering sticky header
   navbarScrollOffset: '85px',
+  navbarScrollMobileOffset: '110px',
```
seungpark, line 78: do we need another offset here? the actionbar is the same height on desktop and mobile

src/components/Target.js
```diff
@@ -1,9 +1,15 @@
...
+  scroll-margin-top: 85px;
```
seungpark, line 9: [minor] use the navbarScrollOffset value here

seungpark (changes requested):
see comments below!

src/components/Permalink.js
biancalaube: what exactly do you mean when you say make it an inline element, like should i change it from a styled div to a styled span ? or is there another way to make it an inline element

src/components/Target.js
biancalaube: thanks for that catch, fixed

src/theme/docsTheme.js
biancalaube: I think we do need an offset here, since it was using the 85px and it was still being hiding. I think its because the spacing is a little different in the mobile view on the paragraphs but basically it was using the 85px and it was still being hidden. I can try another way to fix the issue but Im unsure of what that would be.

seungpark: i believe the comment above will alleviate that, since it will go from the top of the element and count, instead of wrapping another line

seungpark (commented):
minor points: 1) can we update the docstheme so that navbar has more buffer space (dont need to add another, but we can increase the existing)

and 2) update the staging link pls! it is still on an old commit

biancalaube:
@seungpark , change the offset to be 95px but let me know if you want something different, staging links should work now!

seungpark (approved):
this LGTM!

PR #1296, opened 2024-10-28
DOP-5111: Disable gatsby-plugin-image on staging

mmeigs:
### Stories/Links:

DOP-5111

### Current Behavior:

[Current](https://www.mongodb.com/docs/manual/)

### Staging Links:

[Staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/matt.meigs/DOP-5111-disable-image/index.html)

### Notes:

Disallow gatsby-plugin-image on staging builds (used safe boolean to account for staging vs account for production) - unnecessarily adds time to build

### README updates


gatsby-config.js
```diff
@@ -6,13 +6,17 @@ const layoutComponentRelativePath = `./src/layouts/index.js`;
 
 console.log('PATH PREFIX', pathPrefix);
 
+// Disallow gatsby-plugin-image on staging builds - unnecessarily adds time to build
+const stagingEnvs = ['stg', 'prd', 'dev', 'staging'];
+const useImagePlugin = !stagingEnvs.includes(process.env.SNOOTY_ENV);
```
seungpark, line 11: minor. can use siteMetadata.snootyEnv here

seungpark (commented):
can we stage a site where it is not being optimized? seems like the staged site has been (check hero image source) and the deploy log also showing that env is not set https://app.netlify.com/sites/mongodb-snooty/deploys/671fe183bee54d000871c90e#L303

src/utils/setup/save-asset-files.js
```diff
@@ -1,8 +1,12 @@
 const fs = require('fs').promises;
 const path = require('path');
+const { siteMetadata } = require('../site-metadata');
 
 const GATSBY_IMAGE_EXTENSIONS = ['webp', 'png', 'avif'];
 
+const prodEnvs = ['dotcomprd', 'production'];
```
rayangler, line 7: Looks like `SNOOTY_ENV` in Netlify is set to `production`, most likely due to unfortunate legacy frontend shenanigans (prod vs dotcomprd). We should probably just stick to `dotcomprd`? Might be worth double-checking though

gatsby-config.js
```diff
@@ -10,9 +10,9 @@ console.log('PATH PREFIX', pathPrefix);
 // Keep our main plugin at top to include file saving before image plugins
 const plugins = [
   'gatsby-source-snooty-prod',
-  `gatsby-plugin-image`,
-  `gatsby-plugin-sharp`,
-  `gatsby-transformer-sharp`, // Needed for dynamic images
+  'gatsby-plugin-image',
+  'gatsby-plugin-sharp',
+  'gatsby-transformer-sharp', // Needed for dynamic images
```
rayangler, line 15: (Non-blocking question) Would the image nodes created by these plugins also eat up memory? Should we also be omitting them unless it's dotcomprd?

src/utils/setup/save-asset-files.js
mmeigs: Shoot. I did that locally a second ago. 
But I do get nervous about this one because we have code that seems to show both values... 
https://github.com/mongodb/snooty/blob/d928ec109f04c7edc05b40aa59705a09c12a1fbe/src/components/ChatbotUi.js#L138
and
https://github.com/mongodb/snooty/blob/d928ec109f04c7edc05b40aa59705a09c12a1fbe/plugins/gatsby-source-snooty-prod/gatsby-node.js#L266

gatsby-config.js
rayangler: Ah, makes sense. Forgot about the queries.

> Do these nodes take up a lot of memory?

Unsure tbh! Figured I'd just bring it up since the original ticket brought up memory usage of the plugins as a concern. Down to keep as-is for now and re-investigate at a later time

src/utils/setup/save-asset-files.js
rayangler: I see. I think these examples help support that we should just use `dotcomprd` if the intention is to only do this on production deploys.

Based on the AC of the ticket, do we actually want to limit this to `dotcomprd` and `dotcomstg` as well? For test deploys and prod deploys?

mmeigs: Good callout on dotcomstg. 

But I am still nervous about if snootyEnv is ever 'production'. Almost all of the checks on our frontend reference 'staging' and 'development' and 'production' rather than 'dotcomprd' and 'dotcomstg'

seungpark: +1 for `dotcomstg` and `dotcomprd`. `production` i would also like to leave in, but if netlify builds are included, we should write exclude it and write a comment to explain this

rayangler: I would be in favor of keeping the list to `dotcomprd` and `dotcomstg` given the current state of our envs, and then re-evaluating if we have any incorrect configurations for other parts of the codebase later. Either way, we know that we only want to limit it to prod deploys and test deploys.

Seems like the Autobuilder uses `dotcomprd` for prod deploys based on [this call](https://github.com/mongodb/docs-worker-pool/blob/main/.github/workflows/deploy-prd-ecs.yml#L69) in the GH action and [this definition](https://github.com/mongodb/docs-worker-pool/blob/f54eaa08cb7262e811a4e2519708b5e153cdfae0/infrastructure/ecs-main/ecs_service.yml#L35-L36), so it should be safe to use either way.

mmeigs: This was a very good discussion because it turns out Netlify builds might use ... 'production'

rayangler: Yup, I'm assuming it was to keep consistent with Gatsby Cloud builders, which were trying to keep consistent with what the Autobuilder used to set back in the day for certain [database fetching](https://github.com/mongodb/snooty/blob/d4f05a747cb20504888e386833444a05d97ea4cd/src/utils/site-metadata.js#L19) on the frontend. Definitely an opportunity for us to update everywhere in the future once we finally get around to renaming/consolidating our environments better.

Thanks for looking into it + handling the change!

rayangler (approved):

seungpark (approved):
LGTM 👍

PR #1300, opened 2024-10-31
DOP-4810: All Products dropdown docs-qa links

mmeigs:
### Stories/Links:

DOP-4810

### Current Behavior:

I just staged this branch... but clicking on any of the products would take you to something like `www.mongodb.com.website/atlas`

### Staging Links:

[Preprd landing](https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/)
Might be staged back by someone else to how it used to be (buggy)

Pictures of hovered urls
Screenshot 2024-10-31 at 4 21 18 PM
Screenshot 2024-10-31 at 4 21 27 PM

### Notes:

Went for the simplest approach. Modifying `baseUrl` util would have had wider ranging consequences.

### README updates

### DIFF
--- a/plugins/utils/products.js
+++ b/plugins/utils/products.js
@@ -1,13 +1,7 @@
-const { siteMetadata } = require('../../gatsby-config');
-const { baseUrl } = require('../../src/utils/base-url');
-
 const createProductNodes = async ({ db, createNode, createNodeId, createContentDigest }) => {
   // Get all MongoDB products for the sidenav
   const products = await db.fetchAllProducts();
   products.forEach((product) => {
-    let url = baseUrl(product.baseUrl + product.slug);
-    if (siteMetadata.snootyEnv === 'dotcomstg') url = product.baseUrl + product.slug;
-
     createNode({
       children: [],
       id: createNodeId(`Product-${product.title}`),
@@ -17,7 +11,7 @@ const createProductNodes = async ({ db, createNode, createNodeId, createContentD
       },
       parent: null,
       title: product.title,
-      url,
+      url: product.baseUrl + product.slug,
     });
   });
 };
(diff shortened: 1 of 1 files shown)


plugins/utils/products.js
```diff
@@ -1,10 +1,12 @@
+const { siteMetadata } = require('../../gatsby-config');
 const { baseUrl } = require('../../src/utils/base-url');
 
 const createProductNodes = async ({ db, createNode, createNodeId, createContentDigest }) => {
   // Get all MongoDB products for the sidenav
   const products = await db.fetchAllProducts();
   products.forEach((product) => {
-    const url = baseUrl(product.baseUrl + product.slug);
+    let url = baseUrl(product.baseUrl + product.slug);
+    if (siteMetadata.snootyEnv === 'dotcomstg') url = product.baseUrl + product.slug;
```
rayangler, line 9: Would it make sense to just make `url = product.baseUrl + product.slug` the default here instead of needing to worry about different behavior per environment? The `baseUrl()` function was useful here back when we had to worry about building sites on both `docs.mongodb.com` and `mongodb.com/docs`, but I think it makes sense to respect the full url as it is now.

(Perhaps an out of scope tangent, but I'm also wondering if we still need to separate the baseUrl and slug of a product into separate fields? Would it make sense to just consolidate into some targetUrl field for simplicity?)

mmeigs: Oh, I like this. I will go over the data manually just to make sure there are no errors that baseUrl was correcting unbeknownst to us. 

And I'll create a ticket to look into merging the fields if they're not used otherwise!

mmeigs: Created: https://jira.mongodb.org/browse/DOP-5164

rayangler (approved):
Love not needing `baseUrl`!

PR #1304, opened 2024-11-07
DOP-5105: add offline banner

seungpark:
### Stories/Links:

DOP-5105

### Current Behavior:

[Prod page with video](https://www.mongodb.com/docs/atlas/atlas-search/view-query-analytics/)
[Prod page with instruqt lab](https://www.mongodb.com/docs/atlas/atlas-search/tutorials/)

### Staging Links:

[atlas landing page](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/seung.park/DOP-5105/index.html)
["Staged offline docs" for the same video example as above](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/seung.park/DOP-5105/atlas-search/view-query-analytics/index.html) - the resulting HTML is also found [here](https://us-east-2.console.aws.amazon.com/s3/buckets/docs-mongodb-org-stg?region=us-east-2&bucketType=general&prefix=master/cloud-docs/seung.park/DOP-5105/atlas-search/view-query-analytics/&showversions=false) 
[Instruqt lab example](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/seung.park/DOP-5105/atlas-search/tutorials/index.html)
[Page with neither for regression check](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/seung.park/DOP-5105/setup-cluster-security/index.html)
[Regression check for existing banner](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/seung.park/DOP-5105/crud/index.html)
[checking for drivers index template as well](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/drivers/seung.park/DOP-5105/index.html)

### Notes:
- Posted [a question in the figma](https://www.figma.com/design/vmpUl9fzD4b7DHScodaVDu?node-id=398-7825#1015657709) to verify that this banner is only for pages with disabled components ie. Video and Instruqt labs
- UPDATE: above has been changed to `have this banner on every page`. Verifying with Design that the widths/centering are good on different templates.

### README updates

### DIFF
--- a/src/components/Banner/OfflineBanner.js
+++ b/src/components/Banner/OfflineBanner.js
@@ -8,7 +8,7 @@ import { baseBannerStyle, offlineBannerStyle, offlineBannerContainerStyle } from
 const OfflineBanner = ({ linkUrl = 'https://mongodb.com/docs', template = 'document' }) => {
   return (
     <div className={cx(offlineBannerContainerStyle({ template }))}>
-      <LeafyBanner className={cx(baseBannerStyle, offlineBannerStyle)} variant={'warning'}>
+      <LeafyBanner className={cx(baseBannerStyle, offlineBannerStyle({ template }))} variant={'warning'}>
         You are viewing an offline version of MongoDB documentation. Some page features might be unavailable. To view
         the latest version of the page or use interactive features, visit the&nbsp;
         <Link to={linkUrl}>live page.</Link>
--- a/src/components/Banner/OfflineBanner.js
+++ b/src/components/Banner/OfflineBanner.js
@@ -3,15 +3,16 @@ import PropTypes from 'prop-types';
 import LeafyBanner from '@leafygreen-ui/banner';
 import { cx } from '@leafygreen-ui/emotion';
 import Link from '../Link';
+import { assertTrailingSlash } from '../../utils/assert-trailing-slash';
 import { baseBannerStyle, offlineBannerStyle, offlineBannerContainerStyle } from './styles/bannerItemStyle';
 
-const OfflineBanner = ({ linkUrl = 'https://mongodb.com/docs', template = 'document' }) => {
+const OfflineBanner = ({ linkUrl = 'https://mongodb.com/docs/', template = 'document' }) => {
   return (
     <div className={cx(offlineBannerContainerStyle({ template }))}>
       <LeafyBanner className={cx(baseBannerStyle, offlineBannerStyle({ template }))} variant={'warning'}>
         You are viewing an offline version of MongoDB documentation. Some page features might be unavailable. To view
         the latest version of the page or use interactive features, visit the&nbsp;
-        <Link to={linkUrl}>live page.</Link>
+        <Link to={assertTrailingSlash(linkUrl)}>live page.</Link>
       </LeafyBanner>
     </div>
   );
--- a/src/components/Banner/OfflineBanner.js
+++ b/src/components/Banner/OfflineBanner.js
@@ -5,9 +5,9 @@ import { cx } from '@leafygreen-ui/emotion';
 import Link from '../Link';
 import { baseBannerStyle, offlineBannerStyle } from './styles/bannerItemStyle';
 
-const OfflineBanner = ({ linkUrl = 'https://mongodb.com/docs', template = 'document' }) => {
+const OfflineBanner = ({ linkUrl = 'https://mongodb.com/docs' }) => {
   return (
-    <LeafyBanner className={cx(baseBannerStyle, offlineBannerStyle({ template }))} variant={'warning'}>
+    <LeafyBanner className={cx(baseBannerStyle, offlineBannerStyle)} variant={'warning'}>
       You are viewing an offline version of MongoDB documentation. Some page features might be unavailable. To view the
       latest version of the page or use interactive features, visit the&nbsp;
       <Link to={linkUrl}>live page.</Link>
--- a/src/components/Banner/OfflineBanner.js
+++ b/src/components/Banner/OfflineBanner.js
@@ -3,15 +3,17 @@ import PropTypes from 'prop-types';
 import LeafyBanner from '@leafygreen-ui/banner';
 import { cx } from '@leafygreen-ui/emotion';
 import Link from '../Link';
-import { baseBannerStyle, offlineBannerStyle } from './styles/bannerItemStyle';
+import { baseBannerStyle, offlineBannerStyle, OfflineBannerContainer } from './styles/bannerItemStyle';
 
-const OfflineBanner = ({ linkUrl = 'https://mongodb.com/docs' }) => {
+const OfflineBanner = ({ linkUrl = 'https://mongodb.com/docs', template = 'document' }) => {
   return (
-    <LeafyBanner className={cx(baseBannerStyle, offlineBannerStyle)} variant={'warning'}>
-      You are viewing an offline version of MongoDB documentation. Some page features might be unavailable. To view the
-      latest version of the page or use interactive features, visit the&nbsp;
-      <Link to={linkUrl}>live page.</Link>
-    </LeafyBanner>
+    <OfflineBannerContainer template={template}>
+      <LeafyBanner className={cx(baseBannerStyle, offlineBannerStyle)} variant={'warning'}>
+        You are viewing an offline version of MongoDB documentation. Some page features might be unavailable. To view
+        the latest version of the page or use interactive features, visit the&nbsp;
+        <Link to={linkUrl}>live page.</Link>
+      </LeafyBanner>
+    </OfflineBannerContainer>
   );
 };
 
--- a/src/components/Banner/OfflineBanner.js
+++ b/src/components/Banner/OfflineBanner.js
@@ -3,17 +3,17 @@ import PropTypes from 'prop-types';
 import LeafyBanner from '@leafygreen-ui/banner';
 import { cx } from '@leafygreen-ui/emotion';
 import Link from '../Link';
-import { baseBannerStyle, offlineBannerStyle, OfflineBannerContainer } from './styles/bannerItemStyle';
+import { baseBannerStyle, offlineBannerStyle, offlineBannerContainerStyle } from './styles/bannerItemStyle';
 
 const OfflineBanner = ({ linkUrl = 'https://mongodb.com/docs', template = 'document' }) => {
   return (
-    <OfflineBannerContainer template={template}>
+    <div className={cx(offlineBannerContainerStyle({ template }))}>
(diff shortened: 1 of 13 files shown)


src/components/Banner/OfflineBanner.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import LeafyBanner from '@leafygreen-ui/banner';
+import Link from '../Link';
+import { baseBannerStyle } from './styles/bannerItemStyle';
+
+const OfflineBanner = () => (
+  <LeafyBanner className={baseBannerStyle} variant={'warning'}>
+    You are viewing an offline version of MongoDB documentation. Some page features might be unavailable. To view the
+    latest version of the page or use interactive features, visit the&nbsp;
+    <Link to={'https://mongodb.com/docs'}>live page.</Link>
```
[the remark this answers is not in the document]
seungpark: good catch! added getting prod url logic !

src/components/DocumentBody.js
```diff
@@ -134,6 +137,16 @@ const DocumentBody = (props) => {
     sessionStorage.setItem('pageInfo', JSON.stringify(pageInfo));
   }
 
+  const [OfflineBannerComponent] = useState(() => {
```
rayangler, line 140: Might be confusing to make this a state if we never plan to change it. Is it more accurate to just make this a regular variable or `useMemo` instead?

src/components/OfflineNotAvailable.js
```diff
@@ -72,6 +72,7 @@ const OfflineNotAvailable = ({ assetKey }) => {
   const assetLabel = assetLabelFromKey[assetKey];
   const altText = 'Unavailable offline';
   const imgPath = withPrefix('assets/offline-asset.png');
+  // TODO: update this URL. missing project in url
```
rayangler, line 75: Do we need the project in the URL? Not every docs repo has its project in its path prefix (for example, Atlas/`cloud-docs`)

src/components/Banner/OfflineBanner.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import LeafyBanner from '@leafygreen-ui/banner';
+import { cx } from '@leafygreen-ui/emotion';
+import Link from '../Link';
+import { baseBannerStyle, offlineBannerStyle, offlineBannerContainerStyle } from './styles/bannerItemStyle';
+
+const OfflineBanner = ({ linkUrl = 'https://mongodb.com/docs', template = 'document' }) => {
```
rayangler, line 8: (Nit) Can we include a trailing slash for the default `linkUrl`? Since our docs site redirects to the same URL with a trailing slash, it'd be good to add it here for consistency, and to avoid an extra redirect.

Alternatively, we can maybe assert trailing slash on line 14, assuming `linkUrl` will always be a docs URL?

src/components/DocumentBody.js
```diff
@@ -134,6 +137,16 @@ const DocumentBody = (props) => {
...
+    if (!isOfflineDocsBuild) return <></>;
+    const currentBranch = repoBranches.branches.find((repoBranch) => repoBranch.gitBranchName === branch);
+    const currentBranchPrefix =
+      currentBranch?.urlSlug || currentBranch?.urlAliases?.[0] || currentBranch?.urlSlug || '';
+    return (
+      <OfflineBanner linkUrl={getCompleteUrl(getUrl(currentBranchPrefix, project, 'docs', slug))} template={template} />
```
rayangler, line 146: Will offline builds still be built with a path prefix? If so, since we're just getting the branch that the site is built with, would it be possible to just do `linkUrl={getCompleteUrl(withPrefix(slug))}`?

src/templates/document.js
```diff
@@ -13,7 +13,7 @@ import AssociatedVersionSelector from '../components/AssociatedVersionSelector';
 import { theme } from '../theme/docsTheme';
 import { usePageContext } from '../context/page-context';
 
-const MAX_CONTENT_WIDTH = '775px';
+export const MAX_CONTENT_WIDTH = '775px';
```
rayangler, line 16: Are we importing this somewhere?

src/components/DocumentBody.js
seungpark: fair point. i was going for the `this never changes state` but we can utilize usememo here (was going back and forth between the two)

src/components/DocumentBody.js
seungpark: i wasn't sure of the reliability of this, especially with offline docs, so i went with constructing the path manually. offline docs will be building with `--no-prefix` flag to prevent links from prepending links (since it will be purely filepaths relative to page paths)

src/components/OfflineNotAvailable.js
seungpark: ah. thats with the prefix i see. i was under the impression this wasn't working due to my staging branch / local builds not linking correctly. i think this can be verified with offline builds

src/templates/document.js
seungpark: good catch! removed. remnant of banner styling

src/components/Banner/OfflineBanner.js
seungpark: good catch! done

src/components/DocumentBody.js
rayangler: Ah, I didn't realize we were building without the prefix, thanks. Feel free to ignore this then!

src/components/DocumentBody.js
```diff
@@ -134,6 +137,16 @@ const DocumentBody = (props) => {
     sessionStorage.setItem('pageInfo', JSON.stringify(pageInfo));
   }
 
+  const OfflineBannerComponent = useMemo(() => {
+    if (!isOfflineDocsBuild) return <></>;
+    const currentBranch = repoBranches.branches.find((repoBranch) => repoBranch.gitBranchName === branch);
+    const currentBranchPrefix =
+      currentBranch?.urlSlug || currentBranch?.urlAliases?.[0] || currentBranch?.urlSlug || '';
```
rayangler, line 144: We're checking for `urlSlug` twice. Is that intentional? 

Also, should we consider when `urlSlug` is an empty string, such as for unversioned sites like Atlas? Should we use `??` instead of `||`?

src/components/DocumentBody.js
```diff
@@ -134,6 +137,16 @@ const DocumentBody = (props) => {
...
+    return (
+      <OfflineBanner linkUrl={getCompleteUrl(getUrl(currentBranchPrefix, project, 'docs', slug))} template={template} />
```
rayangler, line 146: Passing in `'docs'` as the site base prefix to `getUrl()` (typically used for versioning) might be problematic for non-server docs like Atlas, which has a base prefix of `docs/atlas`, and a branch URL slug of `''`. Should we use [reposBranches.siteBasePrefix](https://github.com/mongodb/snooty/blob/3a2b959ae9133108fcd0625a887afb7eba9d94bf/src/components/DocumentBody.js#L85) instead? Thinking about it now, would having similar logic to add a new `fullPathPrefix` field to the `reposBranches` static query be useful for reuse in both `OfflineBanner` and `OfflineNotAvailable`?

src/components/OfflineNotAvailable.js
seungpark: discussed offline. will do as part of DOP-5172 without prefix !

src/components/DocumentBody.js
```diff
@@ -142,7 +158,7 @@ const DocumentBody = (props) => {
             <FootnoteContext.Provider value={{ footnotes }}>
               <PageContext.Provider value={{ page, template, slug, options: page?.options, tabsMainColumn }}>
                 <div id={TEMPLATE_CONTAINER_ID}>
-                  <Template {...props} useChatbot={useChatbot}>
+                  <Template {...props} useChatbot={useChatbot} offlineBanner={OfflineBannerComponent}>
```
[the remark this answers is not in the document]
seungpark: i've actually passed it as pre-pended to the children previously, but the design showed that we want it above the Breadcrumbs, and templates dictate the Breadcrumbs before the children props. open to exploring other options here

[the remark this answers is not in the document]
rayangler: Ah, I see. Thanks for explaining. I should've taken a better look at the designs.

src/templates/document.js
```diff
@@ -44,6 +44,7 @@ const Document = ({ children, data: { page }, pageContext: { slug, isAssociatedP
         <div className="body">
           <Breadcrumbs siteTitle={title} slug={slug} />
           {activeTutorial && <StepNumber slug={slug} activeTutorial={activeTutorial} />}
+          {offlineBanner}
```
rayangler, line 47: Should `offlineBanner` be above `Breadcrumbs` as described in [this thread](https://github.com/mongodb/snooty/pull/1304/files#r1838219363)?

seungpark: done

tests/unit/__snapshots__/CTABanner.test.js.snap
```diff
@@ -13,8 +13,6 @@ exports[`renders a CTABanner correctly when invalid icon is specified 1`] = `
     href="https://university.mongodb.com/"
   />
   .emotion-0 {
-  margin: 16px 0;
-  font-size: 13px;
```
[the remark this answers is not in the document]
seungpark: done

src/components/Banner/OfflineBanner.js
```diff
@@ -0,0 +1,25 @@
...
+import { assertTrailingSlash } from '../../utils/assert-trailing-slash';
+import { baseBannerStyle, offlineBannerStyle, offlineBannerContainerStyle } from './styles/bannerItemStyle';
+
+const OfflineBanner = ({ linkUrl = 'https://mongodb.com/docs/', template = 'document' }) => {
+  return (
+    <div className={cx(offlineBannerContainerStyle({ template }))}>
+      <LeafyBanner className={cx(baseBannerStyle, offlineBannerStyle({ template }))} variant={'warning'}>
+        You are viewing an offline version of MongoDB documentation. Some page features might be unavailable. To view
+        the latest version of the page or use interactive features, visit the&nbsp;
+        <Link to={assertTrailingSlash(linkUrl)}>live page.</Link>
```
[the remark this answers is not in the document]
seungpark: verified that this works with Design! 

>  let's just defer to whatever lg default is now. it looks like the color changed between when i made the designs and what it currently is so let's just go with whatever the lg color is now

rayangler (approved):

PR #1306, opened 2024-11-11
DOP-5130: Collapsible default expanded

mmeigs:
### Stories/Links:

DOP-5130

### Current Behavior:

No current uses that I know of.

### Staging Links:

Staging with three dummy collapsibles. 
Here's rst (first true, second false, third unspecified)

### Notes:

Content wants collapsibles to be expanded by default. We now will use the expanded option in rst as a boolean. 
Parser PR here.

### README updates


rayangler (approved):
Makes sense. Thanks for the examples!

seungpark (approved):

PR #1311, opened 2024-11-22
DOP-5203: change Chatbot access from search input to button

seungpark:
### Stories/Links:

DOP-5203

### Current Behavior:

[current manual docs with dropdown search input](https://www.mongodb.com/docs/manual/)

### Staging Links:

[docs manual with chatbot action button](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/seung.park/DOP-5203/index.html)
[404 page verification](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs-404/seung.park/DOP-5203/index.html)
Landing preview should be below on the deploy preview from netlify
[staged a mocked translated site here to show chatbot button being hidden](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/zh-cn/master/docs/seung.park/DOP-5203/index.html)

### Notes:
- There were some places where the alignment was not correct, so addressed these in this PR. ie. https://www.mongodb.com/docs/ at 1840px to 2000px

### README updates


src/components/ActionBar/ActionBar.js
```diff
@@ -60,6 +70,12 @@ const ActionBar = ({ template, slug, sidenav, ...props }) => {
       {!isOfflineDocsBuild && (
         <ActionsBox>
           {template !== 'openapi' && <DarkModeDropdown />}
+          <SuspenseHelper>
```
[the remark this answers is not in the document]
rayangler: I was just gonna leave a comment on this! Will that be done as part of this PR?

rayangler, line 73: Are we expected to lazy load this button? This causes a noticeable shift in the dark mode button placement. If not, can we statically load the button but dynamically load the actual chatbot?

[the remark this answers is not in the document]
seungpark: yes! i was just shifting gears after opening so i didn't have the time. will address this

seungpark: good point. i kept the lazy load of this Chatbot to conform with previous lighthouse testing to keep the initial bundle size small, but if it is prominent in the UI on load then perhaps we should not be lazy loading the chatbot at all. WDYT?

if we were to lazy load the Chatbot but use its context in a preloaded button, how would that work? wrap them in a context that lazy loads the Chatbot provider and register the button's function after lazy load(??) this also introduces a race condition if user were to use the button before the lazy load completes

src/components/ActionBar/ActionBar.js
```diff
@@ -60,6 +70,12 @@ const ActionBar = ({ template, slug, sidenav, ...props }) => {
...
+            <Chatbot serverBaseUrl={CHATBOT_SERVER_BASE_URL}>
+              <ChatbotButton></ChatbotButton>
```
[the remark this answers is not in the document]
seungpark: done!

src/components/ActionBar/SearchInput.js
```diff
@@ -1,43 +1,29 @@
...
+import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { navigate } from 'gatsby';
 import PropTypes from 'prop-types';
 import { useLocation } from '@gatsbyjs/reach-router';
 import { css, cx } from '@leafygreen-ui/emotion';
 import IconButton from '@leafygreen-ui/icon-button';
 import Icon from '@leafygreen-ui/icon';
-import { useBackdropClick } from '@leafygreen-ui/hooks';
-import { useDarkMode } from '@leafygreen-ui/leafygreen-provider';
+// import { useDarkMode } from '@leafygreen-ui/leafygreen-provider';
```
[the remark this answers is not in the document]
seungpark: good catches! address comments of commented code

src/components/ActionBar/ActionBar.js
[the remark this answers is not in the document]
seungpark: done!

src/components/ActionBar/SearchInput.js
```diff
@@ -141,99 +85,48 @@ const SearchInput = ({ className, slug }) => {
...
   return (
-    <StyledInputContainer
-      className={cx(className)}
-      mobileSearchActive={mobileSearchActive}
-      onKeyDown={handleSearchBoxKeyDown}
-    >
+    <StyledInputContainer className={cx(className)} mobileSearchActive={mobileSearchActive}>
       <StyledSearchBoxRef ref={searchBoxRef}>
         <LGSearchInput
           aria-label="Search MongoDB Docs"
```
mmeigs, line 119: old nit: Can use PLACEHOLDER_TEXT

src/components/ActionBar/SearchInput.js
```diff
@@ -1,43 +1,25 @@
...
-  Delete: 'Delete',
-  Enter: 'Enter',
-  Escape: 'Escape',
-  Space: ' ',
-  Tab: 'Tab',
-};
+export const PLACEHOLDER_TEXT = `Search MongoDB Docs`;
+const PLACEHOLDER_TEXT_MOBILE = 'Search';
 
 export const SEARCH_SUGGESTIONS = [
```
mmeigs, line 24: This should be deleted.

src/components/ActionBar/styles.js
```diff
@@ -254,6 +250,7 @@ export const overlineStyling = css`
 export const searchIconStyling = css`
   ${displayNone.onLargerThanMedium};
   float: right;
+  justify-content: right;
 `;
 
 // using content before/after to prevent event bubbling up from lg/search-input/search-result
```
mmeigs, line 256: `suggestionStyling ` below here should be deleted.

mmeigs (commented):
Such a nit but I'm wondering if we want to switch over to the mobile action bar sooner OR just allow the search bar to scrunch more. There are about 20 pixels worth of screensize where the styling is rough.

Screenshot 2024-11-25 at 9 31 57 AM

src/components/ActionBar/styles.js
seungpark: done!

src/components/ActionBar/SearchInput.js
seungpark: good catch! done

src/components/ActionBar/ActionBar.js
seungpark: done! kept the Button and button click status as a state in the parent action bar, which is passed as a prop to a lazy loaded chatbot wrapper. PTAL!

src/components/ActionBar/ActionBar.js
```diff
@@ -60,6 +82,25 @@ const ActionBar = ({ template, slug, sidenav, ...props }) => {
       {!isOfflineDocsBuild && (
         <ActionsBox>
           {template !== 'openapi' && <DarkModeDropdown />}
+          {locale === 'en-us' && (
+            <Button
+              className={cx(buttonStyling)}
```
rayangler, line 87: I just remembered that `locale` is grabbed client-side while the button is now static. Is there a chance we can hide the button through CSS to ensure it's removed on initial render? Something like:

```js
const buttonStyling = css`
  // Other CSS here

  // Hide on all languages, by default
  display: none;

  // Only show in English; might need to wrap this in screen size query
  // to make sure this doesn't overwrite the conditional display for mobile/tablet.
  &:lang(en) {
    display: inline-flex;
  }
`;
```

src/components/ActionBar/LazyChatbot.js
```diff
@@ -0,0 +1,34 @@
+import React, { useEffect } from 'react';
+import { useChatbotContext, ModalView, MongoDbLegalDisclosure, PoweredByAtlasVectorSearch } from 'mongodb-chatbot-ui';
+import { css } from '@leafygreen-ui/emotion';
+
+const LazyChatbot = ({ chatbotClicked, setChatbotClicked }) => {
```
rayangler, line 5: (Nit) This component itself isn't inherently lazy since that'd depend on how it's rendered. Would it be more accurate to call it something like `ChatbotModal`?

src/components/ActionBar/styles.js
```diff
@@ -254,34 +271,16 @@ export const overlineStyling = css`
...
+  justify-content: right;
 `;
 
 export const offlineStyling = css`
   @media ${theme.screenSize.largeAndUp} {
     display: none;
   }
 `;
+
+export const buttonStyling = css`
```
rayangler, line 283: Will there be a follow-up PR for the mobile button for this?

src/components/ActionBar/ActionBar.js
rayangler: Makes sense. Thanks!

package-lock.json
```diff
@@ -4727,10 +4727,9 @@
       }
     },
     "node_modules/@leafygreen-ui/icon": {
-      "version": "12.7.0",
-      "resolved": "https://artifactory.corp.mongodb.com/artifactory/api/npm/npm/@leafygreen-ui/icon/-/icon-12.7.0.tgz",
-      "integrity": "sha512-RM+ohvFLF9JCj1WumiqiKUSF7bvYNV4uNWgHCWeXFpydJC+j6wqGXpclIXMHYeP2/bl54aHOgmaY/1FKq7QvRA==",
-      "license": "Apache-2.0",
+      "version": "12.8.0",
```
rayangler, line 4730: (Non-blocking) Any idea why there is a `package-lock` update without a `package` update? 🤔

src/components/ActionBar/ActionBar.js
rayangler: Actually, I think we can simplify the CSS logic to be something like:

```js
const buttonStyling = css`
  // Other CSS here

  &:not(:lang(en)) {
    display: none;
  }
`;
```

seungpark: thanks for the suggestion! done. added a staging link to test translated sites above

src/components/ActionBar/styles.js
```diff
@@ -254,34 +274,31 @@ export const overlineStyling = css`
...
   }
 `;
 
-export const offlineStyling = css`
-  @media ${theme.screenSize.largeAndUp} {
-    display: none;
-  }
+export const mobileButtonStyling = css`
+  cursor: pointer;
+  ${displayNone.onLargerThanTablet};
```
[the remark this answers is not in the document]
seungpark: included the language check in both components

src/components/ActionBar/styles.js
```diff
@@ -221,12 +240,13 @@ export const ActionsBox = styled('div')`
   grid-column: -2/-1;
 
   @media ${theme.screenSize.upToLarge} {
+    column-gap: 1px;
```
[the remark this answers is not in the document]
seungpark: good catch! updated the mobile gaps and updated the staging links

rayangler (approved):
LGTM!

PR #1332, opened 2024-12-31
DOP-5242 🐛  fixes the issue of the tab selector now rendering on mobile view

caesarbell:
### Stories/Links:

DOP-5242

### Current Behavior:

[Docs Altas Prerequisites](https://www.mongodb.com/docs/atlas/driver-connection/#prerequisites)

### Staging Links:

### Notes:

This PR fixes the issue of the tab selector not rendering in the mobile view.

### README updates

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -7,6 +7,6 @@ command = ". ./build.sh $REPO_NAME $PARSER_VERSION"
 
 [build.environment]
 ORG_NAME = "mongodb"
-REPO_NAME = "docs-landing"
+REPO_NAME = "cloud-docs"
 BRANCH_NAME = "master"
 PARSER_VERSION = "0.18.9"
--- a/netlify.toml
+++ b/netlify.toml
@@ -6,7 +6,7 @@ publish = "public"
 command = ". ./build.sh $REPO_NAME $PARSER_VERSION"
 
 [build.environment]
-ORG_NAME = "mongodb"
+ORG_NAME = "10gen"
 REPO_NAME = "cloud-docs"
 BRANCH_NAME = "master"
 PARSER_VERSION = "0.18.9"
--- a/netlify.toml
+++ b/netlify.toml
@@ -6,7 +6,7 @@ publish = "public"
 command = ". ./build.sh $REPO_NAME $PARSER_VERSION"
 
 [build.environment]
-ORG_NAME = "10gen"
+ORG_NAME = "mongodb"
 REPO_NAME = "cloud-docs"
 BRANCH_NAME = "master"
 PARSER_VERSION = "0.18.9"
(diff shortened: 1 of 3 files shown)


netlify.toml
```diff
@@ -6,7 +6,7 @@ publish = "public"
 command = ". ./build.sh $REPO_NAME $PARSER_VERSION"
 
 [build.environment]
-ORG_NAME = "mongodb"
-REPO_NAME = "docs-landing"
+ORG_NAME = "10gen"
```
seungpark, line 9: do we still need this change? seems like it was used for staging

seungpark (approved):
Change lgtm! i think we forgot to add this property when [allowing tabs to be directly in content](https://github.com/mongodb/snooty/pull/1228/files)

Would verify with design team for mobile/tablet views:

image

image


image

mmeigs (approved):
Same exact comments as Seung! The netlify.toml and checking about spacing with design team for that medium screen.

caesarbell:
@seungpark @mmeigs good catch. 
image
image

src/components/Heading.js
```diff
@@ -58,7 +58,7 @@ const Heading = ({ sectionDepth, nodeData, className, as, ...rest }) => {
   const asHeadingNumber = as ?? sectionDepth;
   const asHeading = asHeadingNumber >= 1 && asHeadingNumber <= 6 ? `h${asHeadingNumber}` : 'h6';
   const isPageTitle = sectionDepth === 1;
-  const { isMobile, isTabletOrMobile } = useScreenSize();
+  const { isTabletOrMobile } = useScreenSize();
```
seungpark, line 61: [nit] would prefer if we used css styling with @media queries instead of props using media query states - the latter is dependent on state being updated to mobile (with default false)

seungpark (commented):
looks good stacked! left a minor comment below

src/components/Tabs/TabSelectors.js
```diff
@@ -17,8 +17,7 @@ const selectStyle = css`
   }
 
   @media ${theme.screenSize.smallAndUp} {
-    /* Min width of right panel */
-    max-width: 180px;
+    max-width: 300px;
```
rayangler, line 20: Seems like this is causing horizontal scrolling in the righthand column on desktop sizes due to an existing `margin-right` rule that was probably done for previous spacing concerns (I don't remember what it was, but maybe something to do with how lengthier instances of the "On This Page" component aligned with the selector? 🤷). We should either remove the [right margin](https://github.com/mongodb/snooty/blob/815a53723564d0a336d91fe824be6cc4590ee236/src/components/RightColumn.js#L30) or make the max width smaller to accommodate

Screenshot 2025-01-06 at 5 35 33 PM

rayangler (approved):

src/components/Heading.js
seungpark: change LGTM !

PR #1336, opened 2025-01-09
DOP-5257: implement offline docs download modal

seungpark:
### Stories/Links:

DOP-5257
This PR creates the download modal in the sidenav and allows users to download offline zips into their system.

### Staging Links:

[netlify staging link!](https://6786d873b7306c0008946881--docs-frontend-dotcomstg.netlify.app/)

### Notes:
- This PR caters to the front end of the epic, and the data is currently not available so it has been mocked [here](https://github.com/mongodb/snooty/pull/1336/files#r1915629051). The Atlas Docs selection should lead to a download failure (there should be a visible toast notification), while all others should lead to saved file (with visible success toast notifications). However, this is fake data so once these saved files are unzipped, they are all the same bi-connector zipped file (used this since address is already live on prod)

### README updates


netlify.toml
```diff
@@ -7,6 +7,8 @@ command = ". ./build.sh $REPO_NAME $PARSER_VERSION"
 
 [build.environment]
 ORG_NAME = "mongodb"
-REPO_NAME = "docs-landing"
+REPO_NAME = "cloud-docs"
 BRANCH_NAME = "master"
 PARSER_VERSION = "0.18.9"
+GATSBY_OFFLINE_DOWNLOAD_UI = "true"
+OFFLINE_SNOOTY_ENABLED = "true"
```
seungpark, line 14: reminder: revert before merge

src/components/OfflineDownloadModal/DownloadContext.tsx
```diff
@@ -0,0 +1,136 @@
...
+        if (!branch.active) {
+          continue;
+        }
+        if (branch.offlineUrl) {
+          offlineRepo.versions.push({
+            displayName: branch.label,
+            url: branch.offlineUrl,
+          });
+        }
+        // TODO: REMOVE TESTING
```
seungpark, line 78: TODO: remove before merge

src/components/OfflineDownloadModal/DownloadModal.tsx
```diff
@@ -0,0 +1,253 @@
...
+        Navigate the table to find the product and version you wish to download. Looking for another product? Visit
+        our&nbsp;
+        <Link hideExternalIcon={false} href={'https://mongodb.com/docs/legacy/'}>
+          legacy docs site
+        </Link>
+      </Body>
+
+      <TextInput
+        className={cx(searchInputStyle)}
+        aria-labelledby={'Search for offline documentation'}
```
rayangler, line 203: (Nit) My understanding is that `aria-labelledby` should have a value of an `id`. Would it be more accurate to use `aria-label` here? This comment applies to all instances of `aria-labelledby`

src/utils/download-file.ts
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    // Convert the response to a Blob
+    const blob = await response.blob();
+
+    // Create a temporary object URL
+    const objectURL = URL.createObjectURL(blob);
+    const a = Object.assign(document.createElement('a'), {
+      href: objectURL,
+      download: filename || 'downloaded-file',
```
rayangler, line 18: (Nit) Do we need a default file name if we can ensure that `filename` is always called with a string?

If we do want to keep a default name, can we make it a little more user friendly so they can more easily identify what the file is? Maybe something like `mongodb-docs`

src/components/Sidenav/Sidenav.js
```diff
@@ -300,12 +302,15 @@ const Sidenav = ({ chapters, guides, page, pageTitle, repoBranches, slug, eol })
               <>
                 {isGuidesTemplate && <StyledChapterNumberLabel number={guidesChapterNumber} />}
                 <SideNavItem
-                  className={cx(titleStyle, sideNavItemBasePadding)}
-                  as={Link}
-                  to={isGuidesTemplate ? slug : activeToc.url || activeToc.slug || '/'}
+                  as={Box}
+                  className={cx('not-ia', titleStyle, sideNavItemBasePadding)}
```
rayangler, line 306: What's `not-ia` being used for?

src/utils/assert-leading-brand.ts
```diff
@@ -0,0 +1,18 @@
+type options = {
+  titleCase: boolean;
+};
+
+const assertLeadingBrand = (title: string, options?: options): string => {
+  const casingFn = options?.titleCase
+    ? (e: string) =>
+        e
+          .replace(/(\w\S*)/g, (text) => text.charAt(0).toUpperCase() + text.substring(1))
+          .replace(/(-\w*)/g, (text) => text.substring(0, 1) + text.charAt(1).toUpperCase() + text.substring(2))
```
rayangler, line 10: Do we have example cases of what these are targeting that we can leave as comments? Regex with a combination of substrings with indexes makes it a bit difficult for me to understand what exactly is going on here. (Alternatively, unit tests with different cases could work in place of comments)

rayangler (commented):
This looks really cool!

src/components/Sidenav/Sidenav.js
seungpark: removed! debugging purposes

src/utils/assert-leading-brand.ts
seungpark: left comments and created some test cases for this! lmk if this is sufficient or you'd like to see [more cases included ](https://github.com/mongodb/snooty/pull/1336/files#diff-6bb40ae521e7562dd723e15cf43db9d5ffed13a2144fab192112a7a1849eef1b)

src/components/OfflineDownloadModal/DownloadModal.tsx
seungpark: done! thanks for this catch

src/utils/assert-leading-brand.ts
rayangler: Nice, I think this is useful for understanding and future maintainability! Thanks!

src/components/OfflineDownloadModal/VersionSelector.tsx
```diff
@@ -0,0 +1,75 @@
...
+    <PortalContainer ref={portalRef}>
+      <Select
+        onChange={(e) => {
+          setSelected(parseInt(e, 10));
+        }}
+        portalContainer={portalRef.current}
+        scrollContainer={portalRef.current}
+        className={cx(selectStyling)}
+        allowDeselect={false}
+        aria-labelledby={'Select Offline Version'}
```
rayangler, line 59: (Nit) Should we make this `aria-label` as well?

rayangler (approved):

src/components/OfflineDownloadModal/VersionSelector.tsx
seungpark: looks like a bug on LG Select side [with the current version](https://github.com/mongodb/leafygreen-ui/blob/%40leafygreen-ui/select%407.0.1/packages/select/src/Select.tsx#L105). The updated version seems to take in `label` [by itself](https://github.com/mongodb/leafygreen-ui/blob/main/packages/select/src/Select/Select.tsx#L111). 

also seems like `aria-labelledby` and `label for=""` is preferred when the label is actually visible, so in this case aria-label would be sufficient, but not in our current version of select. i will bypass this by adding a non-existent id for `aria-labelledby` with an `aria-label` that also gets translated into the DOM

seungpark: can see the diff in the snapshot update [here](https://github.com/mongodb/snooty/pull/1336/commits/57857091f177a6b8a14649f5a644c48930320f9c#diff-77f8aaf0731fda467eff8e2434ddb505914f3e00729c850d6e8d5329468c2644)!

rayangler: Interesting, thanks for looking into it. LGTM as long as it doesn't cause anything funky

biancalaube (changes requested):
Hi Seung, this work looks really cool. However, for me in dark mode the modal is looking a little weird and I think it might need some css updates in the `DownloadModal.tsx`
Screenshot 2025-01-15 at 2 09 45 PM

PR #1346, opened 2025-02-10
DOP-5342: ListTable consolidates cell content inside of divs

rayangler:
### Stories/Links:

DOP-5342
[Related Osiris PR](https://github.com/10gen/osiris/pull/32)

### Current Behavior:

[Test tables staging (main branch)](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/sandbox/cloud-docs/raymund.rodriguez/main/test-tables/index.html)

### Staging Links:

[Test tables staging page](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/sandbox/cloud-docs/raymund.rodriguez/DOP-5342-lone-table/test-tables/index.html) - should ideally look the same as the one on the main branch.

### Notes:

* Wraps cell content inside of a div so that all of its content are together when laid out. This allows us greater flexibility with layout styling of cell contents without needing to worry about exact structure.
* The main reason for this change was to help prevent awkward styling with elements passed in from Osiris. Seems like if there were header cells with styling in them, the layout might be awkward (can see a slight difference in spacing in the above [staging link](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/sandbox/cloud-docs/raymund.rodriguez/DOP-5342-lone-table/test-tables/index.html#inline-code-in-header-column)), so this PR helps avoid that going forward.

### README updates

### DIFF
--- a/src/components/ListTable.js
+++ b/src/components/ListTable.js
@@ -78,7 +78,9 @@ const bodyCellStyle = css`
     flex-direction: column;
     align-items: flex-start;
   }
+`;
 
+const bodyCellContentStyle = css`
   *,
   p,
   a {
@@ -86,14 +88,14 @@ const bodyCellStyle = css`
   }
 
   // Target any nested components (paragraphs, admonitions, tables) and any paragraphs within those nested components
-  & > div > *,
-  & > div > div p {
+  & > *,
+  & > div p {
     margin: 0 0 12px;
   }
 
   // Prevent extra margin below last element (such as when multiple paragraphs are present)
-  & > div > div *:last-child,
-  & > div > *:last-child {
+  & > div *:last-child,
+  & > *:last-child {
     margin-bottom: 0;
   }
 `;
@@ -308,7 +310,7 @@ const ListTable = ({ nodeData: { children, options }, ...rest }) => {
                 {headerGroup.headers.map((header) => {
                   return (
                     <HeaderCell className={cx(baseCellStyle, headerCellStyle)} key={header.id} header={header}>
-                      {flexRender(header.column.columnDef.header, header.getContext())}
+                      <div>{flexRender(header.column.columnDef.header, header.getContext())}</div>
                     </HeaderCell>
                   );
                 })}
@@ -323,7 +325,7 @@ const ListTable = ({ nodeData: { children, options }, ...rest }) => {
                 const role = isStub ? 'rowheader' : null;
                 return (
                   <Cell key={cell.id} className={cx(baseCellStyle, bodyCellStyle, isStub && stubCellStyle)} role={role}>
-                    {cell.renderValue()}
+                    <div className={cx(bodyCellContentStyle)}>{cell.renderValue()}</div>
                   </Cell>
                 );
               })}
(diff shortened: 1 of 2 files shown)


src/components/ListTable.js
```diff
@@ -78,23 +77,25 @@ const bodyCellStyle = css`
...
     line-height: 20px;
   }
 
   // Target any nested components (paragraphs, admonitions, tables) and any paragraphs within those nested components
-  & > div > *,
-  & > div > div p {
-    margin: 0 0 12px;
+  & > *,
+  & > div p {
+    margin: 0 0 12px !important;
```
caesarbell, line 92: @rayangler is there any way we can use another approach other than`!important`? if not, then it is what it is. `!important` is normally considered to be bad practice, and can make maintaining difficult down the road.

caesarbell (approved):
Overall, LGTM.  The layout looks real nice. 

I left a small suggestion/comment, but not a blocker.

src/components/ListTable.js
rayangler: Yeah, I'm not a fan of `!important` either. Honestly, I don't really understand it but when I was trying to work on it on local development, it was working fine without `!important`, but when statically built, the order of CSS was different. I couldn't think of or find a more straightforward way to handle it.

I can timebox 30-60 minutes at some point to look into this more to see if there are other options that don't require writing some roundabout CSS/JS logic. Would that work?

caesarbell: @rayangler that would be perfect, but please don't spend anything more than what you mentioned above. I don't think it is worth eating up too much of your time.

src/components/ListTable.js
```diff
@@ -308,7 +309,7 @@ const ListTable = ({ nodeData: { children, options }, ...rest }) => {
                 {headerGroup.headers.map((header) => {
                   return (
                     <HeaderCell className={cx(baseCellStyle, headerCellStyle)} key={header.id} header={header}>
-                      {flexRender(header.column.columnDef.header, header.getContext())}
+                      <div>{flexRender(header.column.columnDef.header, header.getContext())}</div>
```
mmeigs, line 312: My only request: Maybe we leave a comment here and below to explain why the div is there so no one removes it one day.

mmeigs (approved):

src/components/ListTable.js
rayangler: I spent some time looking into it but couldn't find a full fix for it that kept changes minimal. I've tried upgrading a couple of dependencies `@leafygreen-ui/emotion` and `gatsby-plugin-emotion`, and modifying the selectors to be a little bit more specific (this worked for some elements but not for others, and it might require an exhaustive list of HTML since tables can have any elements in them).

Some things I didn't try that might be interesting in the future would be:
* Pass some `table-cell-content` classname down to all cell child components and then use those classnames as the selectors? (Unsure how well this would work with deeply nested components; we might have to make sure every component is passing that exact classname every time)
* We have a context that can keep track if a component is inside of a table or not. Maybe we can leverage that and write some conditional css at the component level to handle the styling accordingly (this may be annoying to have to add to every possible nested component though...)

src/components/ListTable.js
rayangler: Good callout. This sounds like a div I'd accidentally remove

src/components/ListTable.js
caesarbell: @rayangler thank you for looking into this a bit more. Those are some interesting approaches, and like you said, something we can look into in the possible future.

PR #1350, opened 2025-02-18
DOP-5356: Adding versioning functionality to the Unified TOC using versions.toml

biancalaube:
### Stories/Links:

DOP-5356

### Current Behavior:

N/A

### Staging Links:

[Staging link](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/v2.23/csharp/bianca.laube/DOP-5356/quick-start/index.html)

### Notes:
* Staging link should work to switch between versions
* This link is loading using the `docs-csharp` repo so only the first static toc item will load data
* In `UnifiedSidenav.js` I made edits to the url links on the toctree since the toctree will not be using gatsby's `withprefix()` function since in theory we want to switch between sites. I am just printing this new toctree to the console for now. But, I want to test using the modified tree in preprd once this PR is approved to make sure it will load the site properly.
* I edited the `version-context.js` that adds functionality to deal with the `versions.toml` file instead of using `repoBranches`. But I am open to hear feedback of a better way to edit this file to work both ways.

### README updates


src/context/version-context.js
```diff
@@ -31,6 +32,20 @@ const getInitVersions = (branchListByProduct) => {
   return initState;
 };
 
+// Set the inital active version using versions.toml
+const getInitVersionsToml = (bigData) => {
+  const { project } = bigData;
+  let initState = {};
+  const localStorage = getLocalValue(STORAGE_KEY);
+  if (localStorage) {
```
branberry, line 40: This could be simplified to:

```js
  if (localStorage)  return localStorage[project];

  return getDefaultActiveVersionsToml(bigData);

```

src/context/version-context.js
```diff
@@ -199,7 +229,9 @@ const VersionContextProvider = ({ repoBranches, slug, children }) => {
         if (!mountRef.current) {
           return;
         }
-        setActiveVersions(getInitVersions(versions));
+        isUnifiedToc
+          ? setActiveVersions(getInitVersionsToml(project, versionsData))
```
branberry, line 233: It looks like `getInitVersionsToml` only takes in a single argument, but we're providing two here:

```js
const getInitVersionsToml = (bigData) => {

```
Did you mean to use `getDefaultActiveVersionsToml` instead?

branberry (changes requested):
Hey Bianca! This is looking great! I have some feedback around some implementation details, but I think this is definitely in the right direction.

src/components/UnifiedSidenav/UnifiedSidenav.js
```diff
@@ -179,8 +181,57 @@ const isActiveTocNode = (currentUrl, slug, children) => {
...
+};
+
+// Function that adds a prefix to all the urls
+const updateURLs = (tree, prefix, activeVersions, versions, project, snootyEnv) => {
+  tree?.map((item) => {
+    // Getting the path prefix and editing it based on the environment so links work correctly
+    if (item.prefix) {
+      if (snootyEnv === 'dotcomprd') item.prefix = `/docs${item.prefix}`;
+      if (snootyEnv === 'dotcomstg') item.prefix = `/docs-qa${item.prefix}`;
+      const result = replaceVersion(item.prefix, activeVersions[project], versions, project);
```
branberry, line 212: Instead of mutating `'item.prefix`, you can just set the first argument of `replaceVersion` with `${ snootyEnv === 'dotcomprd' ? `/docs${item.prefix}` : snootyEnv === 'dotcomstg' ?  item.prefix = `/docs-qa${item.prefix}` : item.prefix}`

branberry (approved):
LGTM!

PR #1352, opened 2025-02-21
DOP-5411: hide style on code outputs

seungpark:
### Stories/Links:

DOP-5411
This PR fixes the problem where we see the styling as plain text on offline docs. This can be seen on the offline docs, as well as pre-hydration on live production sites. This was due to a small `display: inline !important` styling done in code components that also mistakenly captured style elements

### Current Behavior:

[This page shows a flash of styling text on top of the code output](https://www.mongodb.com/docs/languages/python/pymongo-driver/current/data-formats/custom-types/) - original page from issue description

[Another page where flash of styling text on page from atlas docs](https://www.mongodb.com/docs/atlas/atlas-search/aggregation-stages/search/#troubleshooting)

The problem can be seen on the offline doc [here](http://mongodb.com/docs/offline/pymongo-master.tar.gz) (navigate to the `data-formats/custom-types` page) - this one is persistent because there is no hydration.

### Staging Links:

[netlify staging link](https://67b88d93d9292100087909df--docs-frontend-stg.netlify.app/data-formats/custom-types/) (no styling text pre hydration)

[S3 HTML output](https://us-east-2.console.aws.amazon.com/s3/object/docs-mongodb-org-stg?region=us-east-2&bucketType=general&prefix=HEAD/pymongo/seung.park/DOP-5411/data-formats/custom-types/index.html)

### README updates

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -7,6 +7,6 @@ command = ". ./build.sh $REPO_NAME $PARSER_VERSION"
 
 [build.environment]
 ORG_NAME = "mongodb"
-REPO_NAME = "docs-landing"
+REPO_NAME = "docs-pymongo"
 BRANCH_NAME = "master"
 PARSER_VERSION = "0.18.11"
--- a/netlify.toml
+++ b/netlify.toml
@@ -7,6 +7,6 @@ command = ". ./build.sh $REPO_NAME $PARSER_VERSION"
 
 [build.environment]
 ORG_NAME = "mongodb"
-REPO_NAME = "docs-pymongo"
+REPO_NAME = "docs-landing"
 BRANCH_NAME = "master"
 PARSER_VERSION = "0.18.11"
--- a/src/components/Code/Output.js
+++ b/src/components/Code/Output.js
@@ -12,6 +12,9 @@ const OutputContainer = styled.div`
   > div > * {
     display: inline !important;
   }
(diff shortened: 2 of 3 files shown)


netlify.toml
```diff
@@ -7,6 +7,6 @@ command = ". ./build.sh $REPO_NAME $PARSER_VERSION"
 
 [build.environment]
 ORG_NAME = "mongodb"
-REPO_NAME = "docs-landing"
+REPO_NAME = "docs-pymongo"
```
seungpark, line 10: Revert before merge

src/components/Code/Output.js
```diff
@@ -12,6 +12,9 @@ const OutputContainer = styled.div`
   > div > * {
     display: inline !important;
   }
+  style {
+    display: none !important;
+  }
```
rayangler, line 17: Do we happen to know why this `style` tag is appearing and rendering as HTML on initial load in the first place? Might be preferable to prevent the root of the problem rather than treating the side effect if possible, but if it's not possible, then I'm okay with this

rayangler: Nvm, I just finally understood that the `display` rule you mentioned was causing it... Weird that that applies to style tags as well... 🤔

mmeigs (approved):

PR #1353, opened 2025-02-24
DOP-5291: make code components interactive offline

seungpark:
### Stories/Links:

DOP-5291

### Current Behavior:

Current offline docs do not have any interactivity within code components
[Example of pymongo offline docs](https://www.mongodb.com/docs/offline/pymongo-master.tar.gz)

### Staging Links:

[Sample offline HTML uploaded to S3 bucket](https://us-east-2.console.aws.amazon.com/s3/object/docs-mongodb-org-stg?region=us-east-2&bucketType=general&prefix=HEAD/pymongo/seung.park/DOP-5291/data-formats/custom-types/index.html) - download the file and you can show/hide the code output, as well as copying the code.
**Note:** there is no tooltip when copying the code, but the code is copied

### README updates


src/components/Code/Code.js
```diff
@@ -105,6 +118,7 @@ const Code = ({
         />
       )}
       <div
+        className={isOfflineDocsBuild ? OFFLINE_CONTAINER_CLASSNAME : ''}
```
rayangler, line 121: (Nit) For online builds, this seems to cause the css classname to have unnecessary leading whitespace (if there wasn't any CSS-in-JS, there would just be an empty class attribute on the element). 

This shouldn't negatively impact anything, but would it be possible to return `null` or `undefined` as the falsy value to make it explicit that we don't expect a class in the HTML?

src/components/Code/CodeIO.js
```diff
@@ -69,15 +91,16 @@ const CodeIO = ({ nodeData: { children }, ...rest }) => {
...
               role="button"
-              className={cx(outputButtonStyling)}
+              className={cx(outputButtonStyling, isOfflineDocsBuild ? OFFLINE_BUTTON_CLASSNAME : '')}
               disabled={false}
               onClick={handleClick}
               leftGlyph={<Icon glyph={arrow} fill="#FF0000" />}
             >
-              {buttonText}
+              <span>{buttonText}</span>
+              {isOfflineDocsBuild && <span>SHOW OUTPUT</span>}
```
rayangler, line 100: Is the text intentionally different? "Show output" vs. "View output" (for online builds)?

mmeigs (commented):
Very nice

src/components/Code/Code.js
seungpark: thanks! was wondering why an empty space was appended 🙌

src/components/Code/CodeIO.js
seungpark: this is for [showing/hiding via CSS](https://github.com/mongodb/snooty/pull/1353/files#diff-94d1798d39fd8efcdfd3e5f0f78c00b959c77a669985af17484dc36798d7d695R27-R35) instead of using JS. the two spans are either shown or hidden via css with an appended classname. old school JS CSS.

seungpark: 🤦 updated! made it a getter fn to keep it constant

mmeigs (approved):

rayangler (approved):
Should be good. Small reminder to update the PR title to be descriptive

PR #1357, opened 2025-03-03
DOP-5452: fixes anchor tag wrapping for longer text headings.

caesarbell:
### Stories/Links:

DOP-5452

### Current Behavior:

[Getting Started with the Atlas Architecture Center](https://www.mongodb.com/docs/atlas/architecture/current/getting-started/)

### Staging Links:

### Notes:

~~One approach is to reduce the padding from `0 10px` to `0 5px`, which seems like the simplest solution with the design team's approval, of course. We can also consider using flex styling on the header and inline-block or inline-flex on the anchor tag.~~

~~There is also a `max-content` method we can explore, but this will disrupt the current responsiveness, requiring us to introduce some media query breakpoints.~~

~~I am completely open to other options and approaches.~~

Setting `position: absolute` around the Permalink so it doesn't need to take up any extra whitespace.

### README updates

### DIFF
--- a/src/components/Permalink.js
+++ b/src/components/Permalink.js
@@ -32,7 +32,7 @@ const HeaderBuffer = styled.div`
 
 const headingStyle = (copied) => css`
   ${!!copied && 'visibility: visible !important;'}
-  position:relative;
+  position:absolute;
   align-self: center;
   padding: 0 5px;
   visibility: hidden;
--- a/src/components/Permalink.js
+++ b/src/components/Permalink.js
@@ -34,7 +34,7 @@ const headingStyle = (copied) => css`
   ${!!copied && 'visibility: visible !important;'}
   position:relative;
   align-self: center;
-  padding: 0 10px;
+  padding: 0 5px;
   visibility: hidden;
 `;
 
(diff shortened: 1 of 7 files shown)


caesarbell:
I got the 👍 from design on the spacing between the heading and permalink.

caesarbell:
> this is still an issue at certain breakpoints with certain length headers as the header text fills up the line block.
> 
> image
> 
> i think the approach should be making the permalink absolute positioned so that it does not take up any extra space

Do you mind sharing at what breakpoint you saw this happening?

seungpark:
> > this is still an issue at certain breakpoints with certain length headers as the header text fills up the line block.
> > image
> > i think the approach should be making the permalink absolute positioned so that it does not take up any extra space
> 
> Do you mind sharing at what breakpoint you saw this happening?

https://deploy-preview-1357--docs-frontend-stg.netlify.app/getting-started/

on this page at 845 - 860 px

src/components/Permalink.js
```diff
@@ -32,7 +32,7 @@ const HeaderBuffer = styled.div`
 
 const headingStyle = (copied) => css`
   ${!!copied && 'visibility: visible !important;'}
-  position:relative;
+  position:absolute;
```
seungpark, line 35: [minor nit]

```suggestion
  position: absolute;
```

seungpark (approved):
LGTM. From what i can see, we use Permalinks in Heading and Target components, and [this page](https://www.mongodb.com/docs/manual/reference/error-codes/#mongodb-error-6) using Targets LGTM using `position: absolute` as well.

PR #1360, opened 2025-03-04
DOP-5370: Adding Accordion View to Unified TOC

biancalaube:
### Stories/Links:

DOP-5370

### Current Behavior:

N/A

### Staging Links:

[staging link](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/bianca.laube/DOP-5370/getting-started/index.html)

### Notes:
* This staging link was loaded with cloud-docs so the `csharp` tab on the unified nav wont load, the `Get started` and `application development` tab should work however to click through
* The expected behavior is that on large screen the unified nav should be double panned and then on tablet and mobile it should switch to accordion view
* Since this ticket was done before design the css is minimal and it will be completed in a later ticket, please review if the code style makes sense and if the expected behavior works. 

### README updates


branberry (approved):
Hey Bianca! Looks great, thank you!

PR #1361, opened 2025-03-04
DOP-5412: fix CTA banners in offline

seungpark:
### Stories/Links:

DOP-5412

### Current Behavior:

Current behavior highlighted in the issue can be seen in tar bundle of an offline doc ([link](https://us-east-2.console.aws.amazon.com/s3/object/docs-mongodb-org-dotcomstg?region=us-east-2&bucketType=general&prefix=docs-qa/offline/docs-v6.0.tar.gz)). Navigate to CRUD operations page in the bundle, and the CTA banner:

image

### Staging Links:

Uploaded an offline tarball by updating the contents of a site to use this frontend branch.

[Staging link using netlify deploy](https://67c8b99ac13e52b53570e9e4--seung-cloud-docs.netlify.app/getting-started/)
Above link is generated using [this doc change](https://github.com/seungpark/cloud-docs/pull/3/files#diff-4d2a8eefdf2a9783512a35da4dc7676a66404b6f3826a8af9aad038722da6823R22) to generate a preview and an offline doc download.
[Deploy log here](https://app.netlify.com/sites/seung-cloud-docs/deploys/67c8b99ac13e52b53570e9e4#L780) **NOTE** there is an error updating the db since the doc content belongs to non designated branch.

The generated tarball is located [here in the prd S3 bucket](https://us-east-2.console.aws.amazon.com/s3/object/docs-mongodb-org-prd?region=us-east-2&bucketType=general&prefix=docs/offline/cloud-docs-.tar.gz) and the CTA banner on the `getting-started` page is fixed

### README updates


mmeigs (changes requested):
This works stylistically, but it is changing the functionality. The whole CTA Banner has been an anchor tag. I think we need to keep that functionality? Unless something larger has changed?

seungpark:
> This works stylistically, but it is changing the functionality. The whole CTA Banner has been an anchor tag. I think we need to keep that functionality? Unless something larger has changed?

It's not good practice to have nested anchor tags in general. But you're right, we should preserve the UI, so I added an onClick on the div with a hover effect to keep it the same UX (minus seeing the href beforehand)

branberry (approved):
Looks good to me!

PR #1378, opened 2025-03-24
DOP-5510: handle nested tabs

seungpark:
### Stories/Links:

DOP-5510

### Current Behavior:

Can test by [downloading this Offline Doc](http://mongodb.com/docs/offline/cloud-docs-master.tar.gz), and navigating to the configure-alerts page.
Can also find this in this S3 link: https://us-east-2.console.aws.amazon.com/s3/object/docs-mongodb-org-dotcomprd?region=us-east-2&bucketType=general&prefix=docs/offline/cloud-docs-master.tar.gz

### Staging Links:
Offline pages of a snooty build have been uploaded to [this S3 path](https://us-east-2.console.aws.amazon.com/s3/buckets/docs-mongodb-org-stg?prefix=master%2Fcloud-docs%2Fseung.park%2FDOP-5510%2F&region=us-east-2&bucketType=general&tab=objects)
To replicate above behavior, see [this file](https://us-east-2.console.aws.amazon.com/s3/object/docs-mongodb-org-stg?region=us-east-2&bucketType=general&prefix=master/cloud-docs/seung.park/DOP-5510/configure-alerts/index.html) to see a fix. This is to demonstrate a fix for a nested tab ([RST here](https://github.com/10gen/cloud-docs/blob/master/source/configure-alerts.txt#L218-L227))

See below deploy preview for regression checks. This behavior is only changed with ENV flag `OFFLINE_DOCS = true`

### Notes:
- The behavior was that the script was collecting all Tab Buttons and Tab Panels within a snooty tab component, even when nested. Updated the queries so that they are returned based on 1) relative level and 2) part of the tab data set.

### README updates

### DIFF
--- a/src/utils/head-scripts/offline-ui/tabs.js
+++ b/src/utils/head-scripts/offline-ui/tabs.js
@@ -5,12 +5,45 @@
  */
 
 function bindTabUI() {
+  /**
+   * Get all the tab panels (div [role=tabpanel]) within given snooty tab component
+   * Filters by the tab ids that are tied to the snooty tab (for handling nested tabs)
+   * @param   {Element}   snootyTab
+   * @returns {Element[]}
+   */
   function getTabPanels(snootyTab) {
-    return snootyTab.querySelectorAll(`[class*="lg-ui-tab-panels"] > div > [role=tabpanel]`);
+    // get the tab ids required from the snooty tab
+    const tabIds = (snootyTab.dataset.tabids ?? '').split('/');
+    const tabPanelsContainer = snootyTab.querySelector(`[class*="lg-ui-tab-panels"]`);
+
+    const res = [];
+    // get the tab content divs by tab ids
+    // return list of content divs' parent elms
+    for (const tabId of tabIds) {
+      const tabContentElm = tabPanelsContainer?.querySelector(`[data-tabid=${CSS.escape(tabId)}]`);
+      if (tabContentElm?.parentElement?.getAttribute('role') === 'tabpanel') {
+        res.push(tabContentElm.parentElement);
+      }
+    }
+    return res;
   }
 
+  /**
+   * Returns all the buttons associated with this snooty tab.
+   * Uses the tab ids associated with this tab
+   *
+   * @param   {Element}   snootyTab
+   * @returns {Element[]}
+   */
   function getTabButtons(snootyTab) {
-    return snootyTab.querySelectorAll(`[class*="lg-ui-tab-list"] > [role=tab]`);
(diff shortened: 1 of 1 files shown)


src/utils/head-scripts/offline-ui/tabs.js
```diff
@@ -5,20 +5,52 @@
...
+    // get the tab ids required from the snooty tab
+    const tabIds = (snootyTab.dataset.tabids ?? '').split('/');
+    const tabPanelsContainer = snootyTab.querySelector(`[class*="lg-ui-tab-panels"]`);
+
+    const res = [];
+    // get the tab content divs by tab ids
+    // return list of content divs' parent elms
+    for (const tabId of tabIds) {
+      const tabContentElm = tabPanelsContainer?.querySelector(`[data-tabid=${CSS.escape(tabId)}]`);
+      if (tabContentElm?.parentElement?.getAttribute('role') === 'tabpanel') {
```
rayangler, line 24: For my own understanding, is this just a safety check that we're getting the expected element, or are there known times when this may be false?

src/utils/head-scripts/offline-ui/tabs.js
```diff
@@ -5,20 +5,52 @@
...
+   * @returns {Element[]}
+   */
   function getTabButtons(snootyTab) {
-    return snootyTab.querySelectorAll(`[class*="lg-ui-tab-list"] > [role=tab]`);
+    // get first tab buttons list (vs nested tab buttons)
+    const tabList = snootyTab.querySelector(`[class*="lg-ui-tab-list"]`);
+    const tabIds = (snootyTab.dataset.tabids ?? '').split('/');
+    return tabList
+      ? Array.from(tabList.children).filter(
+          (elm) => elm.getAttribute('role') === 'tab' && tabIds.includes(elm?.dataset?.tabid)
```
rayangler, line 44: (Nit) We shouldn't need to optionally chain `elm` since `elm` will always be defined in the context of the filter function

rayangler (approved):

src/utils/head-scripts/offline-ui/tabs.js
seungpark: ha you're so smart but this is just a safety check. since we already have the tabIds, they should only be tied to tab inner divs, but in the (nonexistant) case where the tab inner divs have tab ids but without a role=tab parent. again, non existent but safety check

src/utils/head-scripts/offline-ui/tabs.js
seungpark: good catch. dataset also always exists, removing

PR #1379, opened 2025-03-27
DOP-5468 🐛  Fixes Tabs in procedures are not fully visible in mobile view issue.

caesarbell:
### Stories/Links:

DOP-5468

### Current Behavior:

[Production](https://www.mongodb.com/docs/manual/tutorial/install-mongodb-on-windows/)

### Staging Links:

### Notes:

The tabs are now visible and scrollable horizontally, which also prevents the content from being cutoff. 

The fix solution: 

Allow the content to shrink properly in a flex container. 

Adding the following

```css
flex: 1; 
min-width: 0;
```

does two things. First, it allows the child to grow and shrink to fill the available space in the flex container, and second tells the browser it's okay for this flex item to shrink below its content's intrinsic width.

### README updates

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -6,7 +6,7 @@ publish = "public"
 command = ". ./build.sh $REPO_NAME $PARSER_VERSION"
 
 [build.environment]
-ORG_NAME = "10gen"
-REPO_NAME = "cloud-docs"
-BRANCH_NAME = "master"
+ORG_NAME = "mongodb"
+REPO_NAME = "docs"
+BRANCH_NAME = "v8.0"
 PARSER_VERSION = "0.18.11"
--- a/netlify.toml
+++ b/netlify.toml
@@ -6,7 +6,7 @@ publish = "public"
 command = ". ./build.sh $REPO_NAME $PARSER_VERSION"
 
 [build.environment]
-ORG_NAME = "mongodb"
-REPO_NAME = "docs"
-BRANCH_NAME = "v8.0"
+ORG_NAME = "10gen"
+REPO_NAME = "cloud-docs"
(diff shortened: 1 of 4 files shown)


mmeigs (changes requested):
This definitely helps! But there are still some pieces that are cut off. Sadly, content is not helping you because of how nested things get.

But I think the solution has to be some kind of actual horizontal scroll, which your PR description says that it can, but I am finding it to be unable to scroll. It just shrinks to screen width rather than allowing overflow and scrolling.

caesarbell:
> This definitely helps! But there are still some pieces that are cut off. Sadly, content is not helping you because of how nested things get.
> 
> But I think the solution has to be some kind of actual horizontal scroll, which your PR description says that it can, but I am finding it to be unable to scroll. It just shrinks to screen width rather than allowing overflow and scrolling.

@mmeigs, sorry for the lack of clarity, the horizontal scroll was referring to the tab issue from the ticket i.e.,`Tabs in procedures are not fully visible in mobile view`. 

Can you tell me where the content is still being cut off?

caesarbell:
Feedback from the design team is the content team should follow best practices and avoid deeply nested content. The tabs should be horizontally scrollable.

mmeigs:
Just for clarity, the design team says and the PR executes the idea that the Tab buttons scroll, but not the tabs (ie not the tab content)

mmeigs (approved):

caesarbell:
> Just for clarity, the design team says and the PR executes the idea that the Tab buttons scroll, but not the tabs (ie not the tab content)

That is correct. The tab content should be addressed at the content creation level.

PR #1380, opened 2025-03-28
DOP-5382: Update breadcrumbs to use data from toc.toml

biancalaube:
### Stories/Links:

DOP-5382

### Current Behavior:

N/A

### Staging Links:

[staging link](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/csharp/bianca.laube/DOP-5382/fundamentals/databases-collections/run-command/index.html)

### Notes:
* updated the `parents` portion of the breadcrumbs
* Goes through the toc.toml and adds the parent breadcrumbs to each entry. Then it searches for the current slug and gets the parents for the current page. Callesd these functions in `Breadcrumb/index.js` and also in `structured-data.js`
### README updates


src/components/Breadcrumbs/UnifiedTocBreadCrumbs.js
```diff
@@ -0,0 +1,38 @@
...
+    newBreadCrumb.path = assertLeadingSlash(removeTrailingSlash(item.url));
+    newBreadCrumb.title = item.label;
+    newCrumbs.push(newBreadCrumb);
+  }
+
+  if (item.items) {
+    for (const nestItem of item.items) {
+      createParentFromToc(nestItem, newCrumbs);
+    }
+  }
```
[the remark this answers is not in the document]
mmeigs: Also, looking again - I wonder if this is why there is a duplicate C#/.NET breadcrumb? It might be pushing this `newCrumb` twice? Just a hypothesis.

mmeigs (changes requested):
The staging link seems to have duplicate breadcrumbs for C#/.NET. Something to look into!

This could be a great time to migrate these files to Typescript! I would totally be down for pairing if you are nervous about doing that. And also, if this is something you're uncomfortable with, maybe just make remove-trailing-slash Typescript for now as a fun easy one!

src/components/Breadcrumbs/UnifiedTocBreadCrumbs.js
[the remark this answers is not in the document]
biancalaube: Hi @mmeigs, thanks for catching that, Im going to look into only do it once since what you are saying makes sense. As per the duplicate `C#/.NET`, that is actually not something that I edited and is in the main branch of snooty (which you can see in the main branch staging link [here](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/csharp/bianca.laube/main/fundamentals/databases-collections/run-command/)). I noticed that they actually link to two different sites. And I looked into the [prod link ](https://www.mongodb.com/docs/drivers/csharp/current/fundamentals/databases-collections/run-command/) and its also duplicated but the second title is edited to `C#/.NET Drivers`. So i wondering if the data was edited somewhere for prod but not staging? Basically Im saying I'm not sure if its a bug or not, but if it is it might be for another ticket that we should create then.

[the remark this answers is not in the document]
biancalaube: also when you say migrate these files to typescript do you mean all of them? Or the ones i created?

src/components/Breadcrumbs/UnifiedTocBreadCrumbs.js
```diff
@@ -0,0 +1,38 @@
+import { assertLeadingSlash } from '../../utils/assert-leading-slash';
+import { removeTrailingSlash } from '../../utils/remove-trailing-slash';
+
+// Goes through toc.toml and builds parent breadcrumb data for each entry
+export const createParentFromToc = (tree, breadcrumbs) => {
+  return tree?.map((item) => {
+    const newCrumbs = [...breadcrumbs];
+    if (item.url) {
+      item.breadcrumbs = [...breadcrumbs];
```
branberry, line 9: `newCrumbs` is equivalent to `[...breadcrumbs];` here

branberry: We don't need to mutate `item` here

src/components/Breadcrumbs/UnifiedTocBreadCrumbs.js
```diff
@@ -0,0 +1,38 @@
...
+      const newBreadCrumb = {};
+      newBreadCrumb.path = assertLeadingSlash(removeTrailingSlash(item.url));
+      newBreadCrumb.title = item.label;
+      newCrumbs.push(newBreadCrumb);
+    }
+
+    const items = createParentFromToc(item.items, newCrumbs);
+
+    return {
+      ...item,
```
branberry, line 19: You can assign `breadcrumbs: newCrumbs,` here

branberry (changes requested):

src/components/Breadcrumbs/UnifiedTocBreadCrumbs.js
biancalaube: I think i made all the changes, please let me know what you think

zdw001:
We'll want to rebase this off the current feature branch that uses `toc.ts` instead of `toc.toml`. I'm also wondering if this is a good opportunity to remove the reliance on `gatsby-node.js` too for the breadcrumbs

src/components/UnifiedSidenav/UnifiedSidenav.js
```diff
@@ -178,7 +178,7 @@ function UnifiedTocNavItem({
               <UnifiedTocNavItem
                 {...tocItem}
                 level={level}
-                key={slug}
+                key={url}
```
rayangler, line 181: Seems like `url` could be shared across different items and I'm seeing console errors about shared keys. Can we change this to be more unique? This comment may apply to other places where we use `.map()` but the same `url` as the key

src/components/StructuredData/BreadcrumbSchema.js
```diff
@@ -1,22 +1,41 @@
...
   const parentPathsSlug = parentPaths[slug];
 
   const queriedCrumbs = useBreadcrumbs();
 
+  const unifiedTocParents = useMemo(() => {
+    if (!isUnifiedToc) return null;
+    const tree = createParentFromToc(tocTree, []);
+
+    return findParentBreadCrumb(slug, tree);
+  }, [slug, tocTree, isUnifiedToc]);
```
rayangler, line 26: (Nit) Seems like this is the same hook logic as the one in `Breadcrumbs/index.js`. Would it be worth making this a reusable hook if we plan on keeping them the same?

Otherwise, I wonder if it'd make sense to refactor the component in a way such that we pass in the data this component needs so that we don't need to construct breadcrumbs in two different places, but maybe this change is out of scope.

rayangler (commented):
Do we need to create tests for these changes, or is this something that will be handled separately?

src/components/Breadcrumbs/UnifiedTocBreadCrumbs.js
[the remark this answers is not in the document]
biancalaube: converted!

src/components/Breadcrumbs/UnifiedTocBreadCrumbs.js
```diff
@@ -0,0 +1,38 @@
...
+
+// Finds breadcrumb through slug
+export const findParentBreadCrumb = (slug, tocTree) => {
+  for (const item of tocTree) {
+    if (assertLeadingSlash(removeTrailingSlash(item.url)) === assertLeadingSlash(removeTrailingSlash(slug))) {
+      return item.breadcrumbs;
+    }
+
+    if (item.items) {
+      return findParentBreadCrumb(slug, item.items);
```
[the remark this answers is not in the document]
biancalaube: I added this change!

src/components/UnifiedSidenav/UnifiedSidenav.js
biancalaube: added the key to be more unique by including the label

rayangler (approved):
LGTM with the assumption that we'll fix any pathing issues in dev/prod accordingly in the future!

PR #1401, opened 2025-04-18
DOP-5595: Work around temp query error by clearing cache and retrying build

cbush:
### TICKET
https://jira.mongodb.org/browse/DOP-5595.

There does not appear to be a fix for the apparent race condition in Gatsby that causes the random 'temp query error'. So as a workaround, we can retry after clearing the LMDB cache if we hit that error.

## Testing

- [x] Successful build
- [x] Build fail but not for temp query: triggered by putting garbage code into gatsby-node.js and forcing an error
- [ ] Retry on temp query error: ran the build 100 times in a loop, eventually hit the error, verified that the script successfully cleared cache and retried and completed the build


cbush:
Could not repro locally (>900 builds!) while retry-script activated. Could the script itself prevent the apparent race condition?

PR #1403, opened 2025-04-21
Rating component bug fixes 

biancalaube:
### Stories/Links:

DOP-5544 / bug fix 

### Current Behavior:

[current](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/cloud-docs/bianca.laube/DOP-5544/index.html)

### Staging Links:

[staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/docs/bianca.laube/rating-fix/reference/connection-string/index.html)

### Bugs that were fixed:
* language selector won't be merging with the rating pop up
* screenshot should work now
* no screenshot for tablet since it should match mobile view now

### Notes:
I was having trouble with the feedback widget for the screenshot view. Since now its not attached to the action bar and attached to the page we are trying to screenshot. I tried moving the pop up to the the body of the page but this caused the submission to not work since the feedback form wasn't attached to the feedback provider anymore. So then I kept the pop up attached to the column and changed the position of the column to be unset for the screenshot, then change it back to sticky once submitted. Please let me know please if this is not proper coding architecture and needs to be done a different way. 

### README updates


src/components/Contents/index.js
seungpark, line 1: a lot of undefined imports here. is this intentional?

src/components/Widgets/FeedbackWidget/components/ScreenshotButton.js
```diff
@@ -194,6 +197,7 @@ const ScreenshotButton = ({ size = 'default', ...props }) => {
   const resetProperties = () => {
     currElem.current = null;
     currElemProperties.current = initialElemProperties;
+    document.getElementById(STYLED_RIGHT_COLUMN).style.position = 'sticky';
```
seungpark, line 200: instead of setting this property of an external component from this button, can we have a state within the right column that checks status via feedback context, and sets it to be sticky / unset there?

seungpark (changes requested):

biancalaube:
@seungpark , sorry for the delay. I realized when i merged with main last time it added back a bunch of files i didnt wan it to. So i fixed that now. I also changed so when we are in screenshot view i change the the form to use a portal and i think that fixes the issues you've pointed out. Please let me know what you think of the new changes.

zdw001 (approved):

PR #1408, opened 2025-04-30
DOP-5375: Adding different nav for drivers 

biancalaube:
### Stories/Links:

DOP-5375

### Current Behavior:

[figma link](https://www.figma.com/proto/KEtaPAkucUDw2Fdr1k6U7v/TOC-WORK?node-id=3629-16846&t=3nvotTc41a2HASxA-0&scaling=min-zoom&content-scaling=fixed&page-id=3615%3A21126&starting-point-node-id=3629%3A16846)

### Staging Links:

[staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/java/bianca.laube/DOP-5375/builders/index.html)

### Notes:
* adding different functionality based on driver figma link
* didn't do all the css since design isn't finalized yet, just focused on functionality 

### README updates


src/hooks/use-unified-toc.js
```diff
@@ -1,5 +1,6 @@
 import { tocData } from '../../toc_data/toc';
 
 export function useUnifiedToc() {
+  // TODO: use memo for preprocessing, if doesnt work, try build time solution
   return tocData();
```
rayangler, line 5: Even if we don't do any preprocessing now, we should probably use `useMemo` to wrap the return value since it's an object and it's used in some dependency arrays (like for `tree` in `UnifiedNav.js`). Dependency arrays can sometimes assume that un-memoized objects changed even if they didn't.

src/components/UnifiedSidenav/UnifiedSidenav.js
```diff
@@ -342,6 +377,43 @@ const updateURLs = ({ tree, prefix, activeVersions, versionsData, project, snoot
   });
 };
 
+const findPageParent = (tree, targetUrl) => {
```
rayangler, line 380: (Non-blocking) I think this might be a good function to write tests for. Happy to have this done at a later time though

src/components/UnifiedSidenav/UnifiedSidenav.js
```diff
@@ -342,6 +377,43 @@ const updateURLs = ({ tree, prefix, activeVersions, versionsData, project, snoot
...
+    return null;
+  };
+
+  for (const item of tree) {
+    const result = dfs(item);
+    if (result) return result;
+  }
+
+  // Should not get here
+  return [false, null];
```
rayangler, line 414: Wouldn't this function get here on the initial homepage when no L1 is selected (like on the docs homepage, based on [designs](https://www.figma.com/proto/KEtaPAkucUDw2Fdr1k6U7v/TOC-WORK?page-id=3565%3A23307&node-id=4310-10409&viewport=-10679%2C769%2C0.85&t=H6PVLKU2bSLKB3SS-1&scaling=scale-down-width&content-scaling=fixed&starting-point-node-id=4323%3A42391&show-proto-sidebar=1))?

src/components/UnifiedSidenav/UnifiedSidenav.js
```diff
@@ -201,6 +208,8 @@ function UnifiedTocNavItem({
             slug={slug}
             isStatic={false}
             isTabletOrMobile={isTabletOrMobile}
+            setCurrentL2s={setCurrentL2s}
+            setShowDriverBackBtn={setShowDriverBackBtn}
```
rayangler, line 212: What's the reason for prop drilling these? Doesn't seem like they're used anywhere other than in within the containing `UnifiedSidenav` component

src/components/UnifiedSidenav/UnifiedSidenav.js
```diff
@@ -354,26 +426,42 @@ export function UnifiedSidenav({ slug, versionsData }) {
...
 
-  const [activeTabUrl, setActiveTabUrl] = useState(() => {
-    const activeToc = tree.find((staticTocItem) => {
+  const [isDriver, currentL2List] = findPageParent(tree, slug);
+  const [showDriverBackBtn, setShowDriverBackBtn] = useState(isDriver);
+  const changeRef = useRef(showDriverBackBtn);
+
+  useEffect(() => {
+    changeRef.current = showDriverBackBtn;
+  }, [showDriverBackBtn]);
```
rayangler, line 449: If `changeRef.current` will always be tied to the value of `showDriverBackBtn`, why do we need `changeRef`? Would we be able to use `showDriverBackBtn` directly for any logic that needs the value?

src/hooks/use-unified-toc.js
biancalaube: Hey raymund, thank you for the comment. I added the usememo(), but i didnt add any dependencies within the useMemo itself. I just want to make sure thats what you want.

src/components/UnifiedSidenav/UnifiedSidenav.js
biancalaube: I've add a write test ticket [DOP-5619](https://jira.mongodb.org/browse/DOP-5619), so i will be doing this in the future!

src/components/UnifiedSidenav/UnifiedSidenav.js
biancalaube: yes, thanks for this catch, updated the comment

src/components/UnifiedSidenav/UnifiedSidenav.js
biancalaube: I add these functions since I use them in `UnifiedTocNavItem()` in a onClick function. Is there a better way to pass these functions?

src/components/UnifiedSidenav/UnifiedSidenav.js
biancalaube: I added this reference since in the useEffect() function on line 462 I update the currentL2s, but ONLY if `showDriverBackBtn` is false. But I didnt want this useffect() function to run on if this variable changed. So I created a reference so i could use the variable in the useffect() without it having a dependency on `showDriverBackBtn`.

src/hooks/use-unified-toc.js
rayangler: Yup, the empty dependency array is what I had in mind since the data is static! Thanks!

src/components/UnifiedSidenav/UnifiedSidenav.js
rayangler: Ah, thanks for this! I misread the change and thought the handle click function was inside the nav container and not in the nav item. 

I think this is fine for now! The only other way I can think of is through a React context, but idk if that's necessarily better atm

src/components/UnifiedSidenav/UnifiedSidenav.js
rayangler: Gotcha, thanks for the explanation. I'm not sure if the extra dependency in the second `useEffect` is necessarily a bad thing given the if statement will still prevent the `setCurrentL2` from running. `changeRef.current` and `showDriverBackBtn` will always have the same value in that case regardless of which variable used, right?

biancalaube: thanks raymund for this explanation. I got rid of the `useRef()` and it should still work as expected. Please let me know if there is anything else

src/components/UnifiedSidenav/UnifiedSidenav.js
```diff
@@ -498,6 +502,7 @@ export function UnifiedSidenav({ slug, versionsData }) {
                       currentL2s={currentL2s.url}
                       isTabletOrMobile={isTabletOrMobile}
                       setCurrentL2s={setCurrentL2s}
+                      setShowDriverBackBtn={setShowDriverBackBtn}
```
rayangler, line 505: (Nit) Since we've renamed the `drivers` field to `showSubNav`, it might make sense to rename this button too to match

rayangler (approved):
LGTM! Left one non-blocking comment but I think it should be fine either way

PR #1410, opened 2025-05-06
DOP-5643: Fix canonical meta option removing other meta options

rayangler:
### Stories/Links:

DOP-5643

### Current Behavior:

### Staging Links:

[Landing staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/raymund.rodriguez/DOP-5643-meta-description/index.html)

### Notes:

The frontend already has logic to check for meta directives towards the top 2 layers of the AST and creates a `meta` array. However, there's additional logic that looks for meta directives that have a `canonical` option. The function responsible for that splices the `meta` array to prevent a "canonical" meta element from being made along with a link element. This causes information of other options within the same meta directive to be lost by the time `meta` is used to render options.

This solution prevents the splicing, and instead filters the `canonical` option out when trying to construct meta elements.

**What to look out for:**

* Inspect the HTML or view page source and notice that the staging link now has "description" as a meta element.
* Other pages on the staging link like `/launch-manage` should still have its expected meta elements.
* Meta component tests and the homepage HTML should still show canonicals as a link element as expected.

**Bonus changes:**
* Converts `Meta.js` to `Meta.tsx`. I didn't find this component converted in the `next-hack` branch, so I figured it was okay to convert here.
* Fixes pre-commit hook to include more than just `.js` files.

### README updates


.lintstagedrc.js
```diff
@@ -8,7 +8,7 @@ const { CLIEngine } = require('eslint');
 // https://github.com/okonet/lint-staged/blob/fa15d686deb90b7ffddfbcf644d56ed05fcd8a38/README.md#how-can-i-ignore-files-from-eslintignore
 
 module.exports = {
-  '*.js': (files) => {
+  '*.{js,jsx,ts,tsx}': (files) => {
```
mmeigs, line 11: thank you!

mmeigs (approved):
Very nice catch

PR #1414, opened 2025-05-09
DOP-5598: Updated pathing of toc.ts

biancalaube:
### Stories/Links:

DOP-5598

### Current Behavior:

### Staging Links:

### Notes:

**For this to work locally now you will need to set the** `PATH_PREFIX` **in your** `.env.development` **and** `.env.production` **files**
* for example: `PATH_PREFIX='docs/atlas'` if you want to load `cloud-docs`
* How it works: 
    * updated the `Links.js` file so now it compares the `prefix` variable in the toc.ts (which should be changed to `site` in DOP-5618) to the `project` variable we get in snootyMetaData(). If its the same it will use a gatsby link and get rid of the the path prefix in the `url` variable. ie: `/docs/atlas/connect-to-database-deployment --> /connect-to-database-deployment` so the gatsby link works correctly. Then if its linking to another netlify docs site it will just use the full `url` ie: `/docs/atlas/connect-to-database-deployment`(this variable named with be changed in DOP-5618 as well). 
* After these changes are implemented we wont before to use s3 to create staging links
* Right now it doesnt work on netlify since the front-end stage site is broken, but it should in theory work
 * The only thing im unsure if it will work is how to get the `docs-frontend-stg` site to build with the feature flag `GATSBY_USE_UNIFIED_TOC` since it hasnt build the unified toc yet
 * However this does work locally to build a site
    * `npm run develop` : should work but you wont see the path prefix
    * `npm run clean`, `npm run build`, `npm run serve`: should work on localhost:9000 with the path prefix you set in the `.env.production` file 
 * TODO: i havent tested these changes in a versioned site, needs to be done in the future
 * NOTE: i branched this pr off of DOP-5371 which is why it has some of the css changes since that is not merged yet in the `feat/unified-nav`

### README updates


zdw001 (approved):
Looks good to go into feature branch!

PR #1416, opened 2025-05-13
DOP-5434: Prevent chatbot from being rendered on non-English locale

rayangler:
### Stories/Links:

DOP-5434

### Current Behavior:

* [zh-cn landing](https://www.mongodb.com/zh-cn/docs/)
* [English landing](https://www.mongodb.com/docs/)

### Staging Links:

* [zh-cn landing staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/zh-cn/master/landing/raymund.rodriguez/DOP-5434-ai-locale/index.html) (Chatbot should not be there on load and when re-translating back to English)
* [English landing staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/landing/raymund.rodriguez/DOP-5434-ai-locale/index.html) (Regression check. Chatbot should still be there)

### Notes:

**The problem:**

The main problem is that client-side translations through Google overwrite the `html` element's `lang` property. This property was used as a CSS selector to ensure the button was displayed as none on non-English locales and prevented jumpy rendering behavior on initial load.

**The change:**

The main change is that the chatbot button is only rendered when the site URL is the English path (based on existing client-side pathname logic). This works together with the existing CSS logic.

**To test:**

1. Click the zh-cn staging link. Confirm that there is no chatbot button.
2. Right-click the page and use Google to translate it to back to English. The chatbot button should still not appear (in contrast to the production staging link). Alternatively, use the browser's inspect tool to modify the `html` element's `lang` property to be "en", if unable to use Google translate.
3. Check the English staging link to confirm that the chatbot still appears.

**Bonus changes:**

* I updated the README to formally go over how to mock languages locally. This is nothing new, but writing down the instructions would probably be good.
* I deleted the `ChatbotFAB` widget. I didn't see it being used anywhere, especially since we're using the button in the action bar. If this is not desired, please let me know and I can change it back!

### README updates


seungpark (approved):
LGTM

minor note the description:
`The main change is that the chatbot button is only rendered when the site isn't English (based on existing client-side pathname logic). This works together with the existing CSS logic.` think you meant `is in English`.

rayangler:
> minor note the description: `The main change is that the chatbot button is only rendered when the site isn't English (based on existing client-side pathname logic). This works together with the existing CSS logic.` think you meant `is in English`.

Updated! Thanks for catching

PR #1422, opened 2025-05-20
DOP-5692: Prevent nav UX bugs in FF and Safari

rayangler:
### Stories/Links:

DOP-5692

### Current Behavior:

Landing Prod

### Staging Links:

Landing Staging

### Notes:

**The problem:**

Interacting with the nav dropdowns of the desktop size UnifiedNav were rendering differently across browsers (works fine on Chrome, but not on Firefox and Safari).

**Changes:**

* Updates the `@mdb/consistent-nav` package to the latest version. This actually made the problem worse because the dropdowns would remain out of place but also now had a width of 0. I'm keeping this in for now as it seems like there's interest in bumping to the latest version anyways.
* Overwrote some CSS that exists in the dropdowns of the nav. This CSS change fixes both the original positional problem as well as the width issue with the latest version. There is a tiny visual shift in Chrome (it moves like a couple of pixels compared to before the change) to accommodate, but I think it still looks fine for now and I think it's worth it to fix the problems.

**Misc.**

* I'm still unsure on the **root cause** of the CSS issues. Doesn't seem like the main dotcom site nor the University site had this issue on their respective versions of the nav, but bumping to `v2.3.5` (the version I'm assuming they have) didn't seem to fix the issue automatically either.
* I also tried updating specific dependencies matching the `consistent-nav` package's [peer dependencies](https://github.com/10gen/consistent-nav/blob/e4189404c10c04982256bc75c5a5f0577c14f2d4/package.json#L77), but none seemed to work.

### README updates

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,7 +47,7 @@
         "@leafygreen-ui/tooltip": "^11.0.4",
         "@leafygreen-ui/typography": "^18.3.0",
         "@loadable/component": "^5.14.1",
-        "@mdb/consistent-nav": "^2.2.0",
+        "@mdb/consistent-nav": "^2.3.6",
         "@mdb/flora": "1.14.2",
         "@mdx-js/react": "^2.2.1",
         "abort-controller": "^3.0.0",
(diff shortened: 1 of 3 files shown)


mmeigs (approved):
How annoying! But this change absolutely patches up the issue.

PR #1432, opened 2025-06-02
bugfix. Dismissible Skills Card allowed without Contents directive

mmeigs:
### Staging Links:

[Staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/main/docs/matt.meigs/fix-dismiss/data-modeling/design-patterns/index.html)

Regression testing staged pages 

[Staging page with Contents list and Card](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/main/docs/matt.meigs/fix-dismiss/data-modeling/design-antipatterns/index.html)

[Staging page that also was not showing a Card in last iteration that now is](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/main/docs/matt.meigs/fix-dismiss/tutorial/authenticate-a-user/index.html)

[Staging page with no Card](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/main/docs/matt.meigs/fix-dismiss/data-modeling/design-patterns/group-data/bucket-pattern/index.html)

### Notes:

Dismissible Skills Cards would not render if there wasn't a Contents directive. Also, fixes line height and removes box-shadow after talking to design.

### README updates

### DIFF
--- a/src/components/Contents/index.tsx
+++ b/src/components/Contents/index.tsx
@@ -9,7 +9,6 @@ import { HeadingNodeSelectorIds } from '../../types/ast';
 import { theme } from '../../theme/docsTheme';
 import useScreenSize from '../../hooks/useScreenSize';
 import useSnootyMetadata from '../../utils/use-snooty-metadata';
-import DismissibleSkillsCard from '../DismissibleSkillsCard';
 import ContentsList from './ContentsList';
 import ContentsListItem from './ContentsListItem';
 import { type ActiveSelectorIds, ContentsContext } from './contents-context';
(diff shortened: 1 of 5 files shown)


src/components/RightColumn.tsx
```diff
@@ -34,7 +34,7 @@ const RightColumn = ({
 
           & > *:not(.${DISMISSIBLE_SKILLS_CARD_CLASSNAME}) {
             margin-bottom: 30px;
-            margin-right: 24px;
+            ${!hasDismissibleSkillsCard ? 'margin-right: 24px;' : ''}
           }
         `)}
```
mmeigs, line 39: This change is because there's a different structure when there isn't a Contents directive, so the margin-right gets applied anyway. If there's a dismissible skills card, the right column is automatically 250px, so it doesn't need the extra inner margin.

src/components/Contents/index.tsx
```diff
@@ -111,6 +111,9 @@ const Contents = ({
   if (filteredNodes.length === 0 || !showContentsComponent) {
     return (
       <div className={className}>
+        {!!dismissibleSkillsCard && (
+          <DismissibleSkillsCard skill={dismissibleSkillsCard.skill} url={dismissibleSkillsCard.url} slug={slug} />
+        )}
```
rayangler, line 116: (Question) Sorry if this question was answered previously in the initial implementation! 

If we want to ensure the `DismissibleSkillsCard` component is rendered regardless of the `Contents` component, is there a reason why it's rendered inside of this component instead of as a direct child within the [`RightColumn` component](https://github.com/mongodb/snooty/blob/5fefc7c7674871dd66d02164d378f25ca3d46e58/src/templates/document.js#L73-L77)?

PR #1440, opened 2025-06-09
Dop 5747

kamilkhan-mongo:
### Stories/Links:

DOP-NNNN

### Current Behavior:

### Staging Links:

### README updates


.gitignore
```diff
@@ -17,4 +17,5 @@ bundle.zip
 # Local Netlify folder
 .netlify
 # pyenv
-.python-version
\ No newline at end of file
+.python-version
+.npmrc
```
mmeigs, line 21: We actually want to keep the `.npmrc` file in our git history! So, remove this change.

src/components/Footnote/FootnoteReference.tsx
```diff
@@ -18,10 +17,26 @@ const refStyles = css`
...
+
+type FootnotesContextType = {
+  footnotes: Record<string, any>;
+};
+
+const FootnoteReference = ({ nodeData }: FootnoteReferenceProps) => {
+  const { footnotes } = useContext(FootnoteContext) as FootnotesContextType;
   const { darkMode } = useDarkMode();
 
+  const { id, refname } = nodeData;
```
mmeigs, line 38: Just wondering why you moved the destructuring down here? I think it's cleaner up where it was!

src/components/Footnote/FootnoteReference.tsx
```diff
@@ -18,10 +17,26 @@ const refStyles = css`
  * Component used to show a clickable reference to footnote on page
  * scrolls to referenced element by id property
  */
-const FootnoteReference = ({ nodeData: { id, refname } }) => {
-  const { footnotes } = useContext(FootnoteContext);
+
+type NodeData = {
+  id: string;
+  refname: string;
+};
```
mmeigs, line 24: So, what you'll need to do is create the `FootnoteReferenceNode` type in `ast.ts` and then just use it below on line 27.

src/components/Footnote/footnote-context.ts
```diff
@@ -0,0 +1,11 @@
+import { createContext } from 'react';
+
+type FootnoteContextType = {
+  footnotes: Record<string, unknown>;
+};
```
mmeigs, line 5: Ah, we don't want to use `any` or `unknown` anywhere really. Especially when it is a known value. I can help you look into exactly what type this is later!

src/types/ast.ts
```diff
@@ -173,7 +173,37 @@ interface Root extends ParentNode {
...
+export interface FootnoteReferenceNode extends Node {
+  id: string;
+  refname?: string;
+}
+
+// For React props:
+export interface FootnoteProps {
+  nodeData: FootnoteNode;
+  [key: string]: any;
+}
```
mmeigs, line 196: All Props types should just live in the component's file! And remove the `[key: string]: any;` line!

src/types/ast.ts
```diff
@@ -173,7 +173,37 @@ interface Root extends ParentNode {
...
+
+export interface FootnoteContextType {
+  footnotes: Record<
+    string,
+    {
+      references: string[];
+      label: string;
+    }
+  >;
+}
```
mmeigs, line 206: Similarly, we will want to have the `FootnoteContextType` live in the `footnote-context` file!

src/components/icons/Book.tsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 
-const IconBook = ({ ...styles }) => (
+type IconBookProps = React.SVGProps<SVGSVGElement>;
+
+const IconBook: React.FC<IconBookProps> = (styles) => (
   <svg width="10" height="11" viewBox="0 0 10 11" fill="currentColor" {...styles} xmlns="http://www.w3.org/2000/svg">
```
mmeigs, line 6: For all of these, it's cleanest to just use these types next to the parameters!

`const IconBook = (styles: IconBookProps) => (`

src/components/Header/index.tsx
```diff
@@ -50,28 +60,28 @@ const Header = ({ eol, template }) => {
...
-                property={{ name: unifiedNavProperty }}
+                property={{ name: unifiedNavProperty, searchParams: [] }}
                 showLanguageSelector={true}
                 onSelectLocale={onSelectLocale}
-                locale={getCurrLocale()}
+                data-locale={getCurrLocale()}
                 enabledLocales={enabledLocales}
                 darkMode={true}
-                className={cx('nav-dark', isOfflineDocsBuild ? offlineClass : '')}
+                data-className={cx('nav-dark', isOfflineDocsBuild ? offlineClass : '')}
```
mmeigs, line 84: Hm, I don't believe we need to add `data-` to these. As we talked about offline, our use of `Locale` is a bit different from the `UnifiedNav`'s repo, so this is the RARE case where I'm okay using a `// @ts-ignore` statement above the locale props. 

But `className` should be fine as it is!

src/components/Footnote/index.tsx
```diff
@@ -41,25 +41,25 @@ const tdStyling = css`
   padding: 11px 5px 12px;
 `;
 
-const Footnote = ({ nodeData: { children, id, name }, ...rest }) => {
-  const { footnotes } = useContext(FootnoteContext);
+const Footnote = ({ nodeData: { children, id, name }, ...rest }: FootnoteProps) => {
+  const { footnotes } = useContext(FootnoteContext) as FootnoteContextType;
```
mmeigs, line 45: This should already have it's typing from FootnotContext, so no need to typecast

mmeigs (commented):
Left some comments below - some are similar to the other PR you have open, so I'm sorry if you already realized :).

Also, make sure to rename the PR something descriptive (but keep the ticket in the title) and also add a description.

src/components/Footnote/footnote-context.ts
kamilkhan-mongo: Yeah good catch! I changed it to this: 

```typescript
export interface FootnoteContextType {
  footnotes: Record<
    string,
    {
      references: string[];
      label: string;
    }
  >;
}
```
How does this look?

ASteinheiser (commented):
left a few small comments/questions, but looks pretty good!

src/components/Header/index.tsx
```diff
@@ -50,24 +59,26 @@ const Header = ({ eol, template }) => {
...
-                fullWidth="true"
-                hideSearch="true"
+                fullWidth={true}
+                hideSearch={true}
                 position="relative"
-                property={{ name: unifiedNavProperty }}
+                property={{ name: unifiedNavProperty, searchParams: [] }}
                 showLanguageSelector={true}
                 onSelectLocale={onSelectLocale}
+                // @ts-ignore
```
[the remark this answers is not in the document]
mmeigs: Yep, so I told Kamil this is ok! The UnifiedNav is a component handled by another team in Mongo and it is actually not being actively supported anymore, and we use Locale a tiny bit differently than they have defined. I thought for now we go with this. I wish we could at least type it as their Locale, but it looks like they don't even export their type... which is just so cool.

mmeigs:
Can you merge in the latest `main`?

kamilkhan-mongo:
Yes, just pushed it!

src/components/Footnote/index.tsx
```diff
@@ -41,26 +41,26 @@ const tdStyling = css`
   padding: 11px 5px 12px;
 `;
 
-const Footnote = ({ nodeData: { children, id, name }, ...rest }) => {
+const Footnote = ({ nodeData: { children, id, name }, ...rest }: FootnoteReferenceProps) => {
```
mmeigs, line 44: Hmm, I could be wrong - but I believe a `FootnoteNode` and a `FootnoteReferenceNode` will be different. So, these components do not share the same type of props.

src/components/Header/index.tsx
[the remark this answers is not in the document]
kamilkhan-mongo: Good to know!

src/components/Footnote/index.tsx
kamilkhan-mongo: Good point! I combined them, but what you’re saying makes sense. 😊

src/types/ast.ts
```diff
@@ -639,6 +643,7 @@ export type {
   HeadingNodeSelectorIds,
   HighlightNode,
   HighlightRoleNames,
+  FootnoteNode,
   FootnoteReferenceNode,
```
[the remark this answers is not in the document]
kamilkhan-mongo: Yes! sorry about this one :smile:

mmeigs (approved):

PR #1441, opened 2025-06-09
dop-5745 typescriptify Card, Collapsible, AssociatedVersionSelector

kamilkhan-mongo:
### Stories/Links:

DOP-5745

### Current Behavior:

- convert Card, Collapsible, AssociatedVersionSelector to ts

### Staging Links:

--

### Notes:

--

### README updates


src/components/Card/CardGroup.tsx
```diff
@@ -1,11 +1,44 @@
...
+      [key: string]: any;
+    };
+    [key: string]: any;
+  };
+  [key: string]: any;
+}
+
+interface CarouselStylingProps {
+  children?: React.ReactNode;
+  theme: Record<string, any>;
```
ASteinheiser, line 23: theme shouldn't be needed here, it comes from the `css` thing we use I believe.

src/components/Card/CardGroup.tsx
```diff
@@ -1,11 +1,44 @@
...
+}
+
+interface StyledGridProps {
+  isForDrivers?: any;
+  isLanding?: any;
+  columns?: any;
```
ASteinheiser, line 29: Can you cleanup these `any`s? Needs to be a more specific type

src/components/Card/index.tsx
```diff
@@ -14,6 +13,22 @@ import Link from '../Link';
...
+  isExtraCompact?: boolean;
+  isCenterContentStyle?: boolean;
+  isLargeIconStyle?: boolean;
+  nodeData: CardNode;
+  page?: {
+    options?: {
+      template?: string;
+      [key: string]: any;
+    };
+    [key: string]: any;
```
ASteinheiser, line 29: remove `[key: string]: any;`

src/components/Card/index.tsx
```diff
@@ -126,21 +149,21 @@ const bodyStyling = css`
...
 `;
 
-const onCardClick = (url) => {
+const onCardClick = (url?: string) => {
+  if (!url) return;
   isRelativeUrl(url) ? navigate(url) : (window.location.href = url);
 };
 
-const Card = ({
+const Card: React.FC<CardProps> = ({
```
ASteinheiser, line 157: nit: apply props to prop object not function

src/components/AssociatedVersionSelector/index.tsx
```diff
@@ -27,12 +55,14 @@ const StyledSelect = styled(Select)`
   }
 `;
 
-const AssociatedVersionSelector = () => {
-  const { project } = useSnootyMetadata();
-  const { activeVersions, availableVersions, hasEmbeddedVersionDropdown, onVersionSelect } = useContext(VersionContext);
+const AssociatedVersionSelector: React.FC = () => {
```
ASteinheiser, line 58: remove React.FC please, as this type is already inferred

src/components/AssociatedVersionSelector/index.tsx
```diff
@@ -27,12 +55,14 @@ const StyledSelect = styled(Select)`
...
+  const { project } = useSnootyMetadata() as { project: string };
+  const { activeVersions, availableVersions, hasEmbeddedVersionDropdown, onVersionSelect } = useContext(
+    VersionContext
+  ) as VersionContextType;
```
ASteinheiser, line 62: in cases where you're type casting (using the `as` keyword), you should ask yourself "why?". I think you might need to go into those functions and make sure their types are proper instead of defining and casting them in this file.

src/components/AssociatedVersionSelector/index.tsx
```diff
@@ -4,8 +4,36 @@ import { VersionContext } from '../../context/version-context';
...
+interface SelectChangeEvent {
+  value: string;
+}
+
+interface VersionContextType {
+  activeVersions: Record<string, string>;
+  availableVersions: Record<string, Branch[]>;
+  hasEmbeddedVersionDropdown: boolean;
+  onVersionSelect: (project: string, version: string) => void;
+}
```
ASteinheiser, line 34: this probably needs to be applied somewhere else (context?)

src/components/AssociatedVersionSelector/index.tsx
```diff
@@ -4,8 +4,36 @@ import { VersionContext } from '../../context/version-context';
...
+};
+
+type Choice = {
+  value: string;
+  text: string;
+};
+
+interface SelectChangeEvent {
+  value: string;
+}
```
ASteinheiser, line 27: nit: can you just define this inline since it's not props and it really small/not shared

src/components/AssociatedVersionSelector/index.tsx
```diff
@@ -4,8 +4,36 @@ import { VersionContext } from '../../context/version-context';
...
+type ThemeType = typeof theme;
+
+declare module '@emotion/react' {
+  export interface Theme extends ThemeType {}
+}
+
+type Branch = {
+  gitBranchName: string;
+  [key: string]: any;
+};
```
ASteinheiser, line 18: I believe this type is already defined elsewhere in the project

src/components/AssociatedVersionSelector/index.tsx
```diff
@@ -4,8 +4,36 @@ import { VersionContext } from '../../context/version-context';
...
+
+type Choice = {
+  value: string;
+  text: string;
+};
```
ASteinheiser, line 23: you can probably let typescript infer this type, can remove I believe

src/components/AssociatedVersionSelector/index.tsx
```diff
@@ -4,8 +4,36 @@ import { VersionContext } from '../../context/version-context';
...
 import { getUILabel } from '../VersionDropdown';
 import useSnootyMetadata from '../../utils/use-snooty-metadata';
+import { theme } from '../../theme/docsTheme';
 
-const buildChoices = (branches) => {
+type ThemeType = typeof theme;
+
+declare module '@emotion/react' {
+  export interface Theme extends ThemeType {}
+}
```
ASteinheiser, line 13: this should be removed, not sure why you're re-declaring the theme? I believe the `css` helpers should have the types for theming set already

ASteinheiser (changes requested):

src/types/ast.ts
```diff
@@ -271,6 +271,28 @@ interface TextNode extends Node {
   type: 'text';
   value: string;
 }
+type CardGroupOptions = {
+  columns: number;
+  layout: string;
+  style: string;
+  type?: string;
+};
+export interface CardGroupNode extends Directive<CardGroupOptions> {
```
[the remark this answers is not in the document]
kamilkhan-mongo: Good call!

src/components/AssociatedVersionSelector/index.tsx
ASteinheiser: search for `interface Branch`

mmeigs:
Andrew's comments are all great up there! Also, could you rename the PR to be descriptive of what the change is doing? And make sure not to delete anything from the PR description template, it's there for a reason! Add a description (this one at least is easy)

src/types/ast.ts
```diff
@@ -271,6 +271,28 @@ interface TextNode extends Node {
...
+  options: CardGroupOptions;
```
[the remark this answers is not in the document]
mmeigs: Actually! Andrew, you're normally right! But here, options is an optional property on the base `Directive`, however in these nodes, options would be mandatory. So, I would recommend actually still declaring them! That way you don't have to optionally chain later on. The parser makes sure the mandatory options are there.

src/components/Collapsible/index.tsx
```diff
@@ -14,11 +13,19 @@ import { isOfflineDocsBuild } from '../../utils/is-offline-docs-build';
...
+import { CollapsibleNode } from '../../types/ast';
 import { collapsibleStyle, headerContainerStyle, headerStyle, iconStyle, innerContentStyle } from './styles';
 import './styles.css';
 
-const Collapsible = ({ nodeData: { children, options }, sectionDepth, ...rest }) => {
-  const { id, heading, expanded, sub_heading: subHeading } = options;
+interface CollapsibleProps {
+  nodeData: CollapsibleNode;
+  sectionDepth: number;
+  [key: string]: any;
```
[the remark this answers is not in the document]
kamilkhan-mongo: Gotcha, I can do that!

kamilkhan-mongo:
Thanks for the feedback, @ASteinheiser @mmeigs! I’ve pushed the latest changes and would appreciate another review when you get a chance :)

src/components/Card/index.tsx
```diff
@@ -126,21 +147,14 @@ const bodyStyling = css`
...
-  page,
-  nodeData: {
-    children,
-    options: { cta, headline, icon, 'icon-dark': iconDark, 'icon-alt': iconAlt, tag, url },
-  },
-}) => {
+const Card = ({ isCompact, isExtraCompact, isCenterContentStyle, isLargeIconStyle, page, nodeData }: CardProps) => {
+  const { children, options } = nodeData;
+  const { cta, headline, icon, 'icon-dark': iconDark, 'icon-alt': iconAlt, tag, url } = options ?? {};
   const template = page?.options?.template;
```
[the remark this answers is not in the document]
kamilkhan-mongo: Thanks for pointing that out! I’ll keep that in mind for future changes.

src/components/AssociatedVersionSelector/index.tsx
kamilkhan-mongo: yup sorry for that!

src/components/AssociatedVersionSelector/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { BranchData } from '../../types/data';
+
+const StyledSelect = css`
+  width: 100%;
+
+  button[aria-expanded='true'] {
+    svg {
+      transform: rotate(180deg);
+    }
+  }
```
mmeigs, line 17: Can you rename this variable now that it's not a styled component? Normally a component is capitalized, but this is now just a style, so just name it `selectStyling` or `selectStyles`

src/components/AssociatedVersionSelector/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  const currentVersions: BranchData[] = availableVersions[project] ?? [];
+
+  return (
+    <>
+      {hasEmbeddedVersionDropdown && !!activeVersions[project] && currentVersions.length > 0 && (
+        <>
+          <Label>Specify your version</Label>
+          <Select
+            className={StyledSelect}
```
mmeigs, line 44: Ah, this needs to use the `cx` function from `@leafygreen-ui/emotion` package now! Like this:
`className={cx(selectStyling)}`

mmeigs (commented):
Two more!

src/components/AssociatedVersionSelector/index.tsx
kamilkhan-mongo: Aha, good catch!

src/components/AssociatedVersionSelector/index.tsx
kamilkhan-mongo: why do we need to use the cx function? from what i understand, its used to help combine multiple class names, but here we only use 1 (selectStyling)

mmeigs: Talked offline: I was wrong - this is not needed!

mmeigs (approved):

PR #1443, opened 2025-06-10
Dop 5758 typescriptify

ASteinheiser:
### Stories/Links:

DOP-5758

### Current Behavior:

- convert standalone, strong, superscript, subscript to ts

### Staging Links:

--

### Notes:

--

### README updates


src/types/ast.ts
```diff
@@ -480,6 +491,17 @@ interface CTABannerNode extends Directive<CTABannerOptions> {
...
 }
 
+type StandaloneHeaderOptions = {
+  columns: number;
+  cta: string;
+  url: string;
+};
+
+interface StandaloneHeaderNode extends Directive<StandaloneHeaderOptions> {
+  argument: Array<Node>;
```
mmeigs, line 501: You shouldn't have to declare line 501 because `Node[]` is already defined in `Directive`.

mmeigs (commented):
Could you change the PR name to obviously signify what's changed? Easiest for us during releases and look-backs to understand quickly. ie `DOP-5758: TS -  Standalone, Strong, Superscript, Subscript`.

Also, look into the test failures! I believe you just haven't added all the correct strings to the ast types.

src/types/ast.ts
ASteinheiser: much cleaner now, thanks!

src/types/ast.ts
```diff
@@ -163,6 +165,7 @@ type DirectiveOptions = {
 
 interface Node {
   type: NodeType;
+  value?: string;
 }
```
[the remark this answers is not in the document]
ASteinheiser: I much prefer that approach! Good call!

src/types/ast.ts
```diff
@@ -516,10 +519,7 @@ type StandaloneHeaderOptions = {
   url: string;
 };
 
-interface StandaloneHeaderNode extends Directive<StandaloneHeaderOptions> {
-  argument: Array<Node>;
-  options: StandaloneHeaderOptions;
-}
+type StandaloneHeaderNode = Directive<StandaloneHeaderOptions>;
```
mmeigs, line 523: Another note: this node has a mandatory option of `StandaloneHeaderOptions`, so I would change it back. I left [this comment](https://github.com/mongodb/snooty/pull/1441#discussion_r2140693595) that explains more. Once you do, you can remove the optional chaining you've added in the `StandaloneHeader` component!

ASteinheiser: I think the tests are fine now? It's still stripping a style out, but I'm curious if that's just the post css whatever cleaning optimizing stuff?

mmeigs (approved):

PR #1445, opened 2025-06-10
Dop 5759 typescriptify

ASteinheiser:
### Stories/Links:

DOP-5759

### Current Behavior:

- convert suspenseHelper, tag, target, and subReference to ts

### Staging Links:

--

### Notes:
--
### README updates


src/components/Target.tsx
```diff
@@ -12,40 +12,44 @@ const headerBuffer = css`
...
-const partition = (array, isValid) => {
+const partition = (array: Node[], isValid: (elem: Node) => boolean) => {
   return array.reduce(
     ([pass, fail], elem) => {
       return isValid(elem) ? [[...pass, elem], fail] : [pass, [...fail, elem]];
     },
-    [[], []]
+    [[] as Node[], [] as Node[]]
   );
 };
```
ASteinheiser, line 22: Not really sure what this function does, wanted to rip it out, but not trying to change too much

mmeigs (commented):
Ok, rename the PR and fix the failing tests. 

But also, with the css prop. I've been removing them as I go and replacing them with CSS-in-JS that is more valid with our pipeline. But we just have to be careful that the styles still work. Luckily, it's usually only an issue with LeafyGreen components and the Tag is all us. 

Usually in these situations I use `cx` and `css` from `'@leafygreen-ui/emotion'` package. That was you can probably keep the `css` usage the same, but just use the cx to transform an array of them to be stuck right in the className prop.

The `GuidesLandingTree` component uses it similarly.

ASteinheiser:
I feel like the snapshot is failing properly, but I'm not sure what I did to change the css. I simply swapped out the old emotion css util for leafy

src/components/Tag.tsx
```diff
@@ -84,7 +84,7 @@ const Tag = ({ children, variant, ...rest }: TagProps) => {
   const variantStyle = VARIANTS[siteTheme][variant] || VARIANTS[siteTheme]['green'];
 
   return (
-    <span css={[baseStyle, variantStyle]} {...rest}>
+    <span className={cx(baseStyle, variantStyle)} {...rest}>
       {children}
     </span>
   );
```
mmeigs, line 90: Aha, I believe I know what's causing the wide difference in styles applied in the test. If you go and find where this component is used, you'll find it in [SearchResult here](https://github.com/mongodb/snooty/blob/7aff20fcb685cbe72d8e63209f920b8a5094c9a4/src/components/SearchResults/SearchResult.tsx#L252).
As you can see, we are passing in a className there directly. With your changes to the base Tag component here, you are basically already explicitly writing in a className and then spreading the rest. My guess is only one of them is winning the battle. So, maybe in this component we can also add className as an optional prop and pass it in to the className cx list.

ASteinheiser: Oh man that makes so much sense! Yup, I'm hijacking the className prop which was getting spread before. Adding that className should fix it I believe!

mmeigs: The best thing to do in this case is to check production versus staging if all the necessary styles are still there! Luckily this component seems to be only used in Search so pretty easy to check!

ASteinheiser: tested with seung and resolved!

src/components/SearchResults/SearchResult.tsx
```diff
@@ -249,28 +245,32 @@ const SearchResult = React.memo(
...
-                </StyledTag>
+                </Tag>
               )}
               {url.includes('/api/') && (
-                <StyledTag variant="purple" className={cx(purpleTagStyles)}>
+                <Tag variant="purple" className={cx(purpleTagStyles, searchTagStyle)}>
                   {'API'}
-                </StyledTag>
+                </Tag>
               )}
```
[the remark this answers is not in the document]
ASteinheiser: Actually the `purpleTagStyles` and `greenTagStyles` are already leafy css despite both emotion and leafy being present in that file! I think it's correct as is?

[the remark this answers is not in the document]
mmeigs: Yeah, those two styles are leafy css. That's why they use cx! But searchTagStyle is @emotion/react.

seungpark (approved):
LGTM !

src/components/SearchResults/SearchResult.tsx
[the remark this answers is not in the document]
ASteinheiser: Yeah, I think I'm missing those other spots you pointed out. Trying that now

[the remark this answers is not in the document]
ASteinheiser: Appears to have worked!

mmeigs (approved):

PR #1448, opened 2025-06-11
DOP-5639 ⚡ Graciously handles L1 and L2s for 404 custom pages

caesarbell:
### Stories/Links:

DOP-5639

### Current Behavior:

### Staging Links:

### Notes:

* Provides a 404 custom page for handling non-existent routes
* Checks if the Unified TOC is in dev mode to enforces client-side routing to properly show the custom 404 page for preview deploys
* Refactors to use the already preexisting isActiveTocNode utility
* Introduces a hijacking method to allow is active styling for the TOC nodes
* Some refactoring with setting the initial state for `showDriverBackBtn`, `currentL1` and `currentL2s`. We are no longer initially setting the state from the computed results, but rather doing that from within the `useEffect` allowing us to properly wait on the completion of the tree processing.
* Passes the `pathPrefix` to the `isActiveTocNode` method that is called within the unified TOC 

### README updates


biancalaube (changes requested):
Hi Caeser, great work! I tested locally and it works great :) my one concern that on the netlify previews if you click on 'database tools' it's not opening the second panel / L2s which i think we want it to do that for the A/C of this ticket ([link here for example](https://deploy-preview-1448--docs-frontend-stg.netlify.app/docs/database-tools/)). Im not exactly sure what the solution is but i think the `currentL2s` variable in `UnifiedSidenav.js` would need to be changed a bit.

src/utils/is-active-toc-node.js
```diff
@@ -1,14 +1,19 @@
 import { isCurrentPage } from './is-current-page';
+import { isUnifiedTocActive } from './is-unified-toc-active';
+
 /*
   Provided a current page slug, a slug, and a list of node children, returns
   true if either the given slug matches the current page slug or one of its children
   has a slug matching the given page slug, and false otherwise.
 */
-export const isActiveTocNode = (currentUrl, slug, children) => {
+export const isActiveTocNode = (currentUrl, slug, children, pathPrefix) => {
```
biancalaube, line 9: NiT: Should this change be behind the unified toc feature flag? Since I think this is a function we use in the main branch currently. However not sure if we need it but wanted to point that out.

biancalaube (approved):
LGTM! Tested locally and in netlify previews

I left a small NiT but these are non-blocking.

One more comment that I have, should we add a comment or some documentation somewhere that in your .env files locally to add the variable `GATSBY_UNIFIED_TOC_DEV_MODE=true` . And once we are moving for writers to test it we will need to add this variable somewhere in netlify for deploy previews but not for preprd and prd builds. Again not apart of this ticket per say but either create a new ticket or just a comment/documentation to remind us to do this for the future.

src/utils/is-active-toc-node.js
caesarbell: @biancalaube good call out, we could, and it would just give us an extra layer of protection. With that said it is still is behind a flag which is used inside the `isUnifiedTocActive` i.e., `isUnifiedTOCInDevMode`. 

But I will push up the change to put this behind the unified feature flag.

PR #1454, opened 2025-06-13
Fix version dropdown navigate

rayangler:
### Stories/Links:

N/A

### Current Behavior:

[Prod](https://www.mongodb.com/docs/cluster-to-cluster-sync/current/) - Using the version dropdown appends the target URL to the current pathname.

### Staging Links:

[Staging](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/node/raymund.rodriguez/fix-version-dropdown-navigate/index.html) - version dropdown should **replace** the pathname instead of add it onto the end.

### Notes:

* Seems like the behavior of `navigate` in `gatsby` and `@gatsby/reach-router` are different. The former seems to append to the current pathname while the latter replaces it as we currently use it. We can probably handle this without `navigate` (i.e. with `location` directly), but just reverting this change in the meantime.

### README updates

### DIFF
--- a/tests/unit/VersionDropdown.test.js
+++ b/tests/unit/VersionDropdown.test.js
@@ -1,5 +1,5 @@
 import { render, screen, within } from '@testing-library/react';
-import { navigate } from 'gatsby';
+import { navigate } from '@gatsbyjs/reach-router';
 import userEvent from '@testing-library/user-event';
 import * as realm from '../../src/utils/realm';
 import VersionDropdown from '../../src/components/VersionDropdown';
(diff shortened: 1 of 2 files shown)


tests/unit/VersionDropdown.test.js
```diff
@@ -1,5 +1,5 @@
 import { render, screen, within } from '@testing-library/react';
-import { navigate } from 'gatsby';
+import { navigate } from '@gatsbyjs/reach-router';
 import userEvent from '@testing-library/user-event';
```
mmeigs, line 3: IF the build fails, we might need a ts-ignore statement here too, but hopefully this will be good.

rayangler: This is a JS file so I'm not sure if a ts-ignore is needed?

PR #1456, opened 2025-06-16
DOP-5618: Update toc.ts structure

biancalaube:
### Stories/Links:

DOP-5618

### Notes:

Please take a close look at the naming for everything to do with the unified toc structure, can also look at the `unified toc tree` structure that logs to the console in the staging link. The writers will be editing the toc.ts so i want to make sure it's clear and makes sense. 

** also, is `contentSite` something we want children to inherit or something that needs to be declared for each side nav item

### README updates


src/components/UnifiedSidenav/UnifiedConstants.ts
```diff
@@ -19,3 +19,10 @@ export interface TocItem {
 export interface ActiveVersions {
   [project: string]: string;
 }
+
+// These has to match how we store the repo name in local storage
```
zdw001, line 23: Nit: I'm wondering if it makes sense to move this to the TOC.ts just assuming that the writers might be the ones adding these enums here

zdw001 (approved):

src/components/UnifiedSidenav/UnifiedConstants.ts
biancalaube: Thanks for noting this zach, i think it would make sense to move this somewhere. Im unsure of moving it into the toc.ts since it will be multiple files, I added a comment to move this to a more finalized location when we have a more finalized toc.ts in DOP-5379

PR #1464, opened 2025-06-18
DOP-5769 DOP-5804: make composable selection smarter

seungpark:
### Stories/Links:

DOP-5769 - cannot select Drivers from this [staging link](https://deploy-preview-12113--mongodb-manual.netlify.app/docs/manual/aggregation/examples/mask-sensitive-information/?interface=mongosh)
DOP-5804 - scroll position to be preserved when selecting a composable option

### Current Behavior:

See staging link above

### Staging Links:

[Locally staged Jeff's PR](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/DOCSP-46671-monorepo/docs/seung.park/DOP-5769/aggregation/examples/mask-sensitive-information/index.html)

### Notes:
- The first bug was due to the fact that when user makes a partial selection (in this case `Interface=Drivers`), we did not flesh the remaining selections before calling navigate, resulting in the search param readers in the useEffect to determine that it was not a full selection.
- The second feature of scroll position was fairly simple - add a state `{ state: { preserveScroll: true } }` to trigger the [gatsby browser config](https://github.com/mongodb/snooty/blob/main/gatsby-browser.js#L10)

### README updates


mmeigs (approved):
Love how simple this is!

PR #1471, opened 2025-06-27
DOP-5752: re add scrolly to style

seungpark:
### Stories/Links:

DOP-5752 follow up. Seems like we introduced a bug in the height of sidenav.

### Current Behavior:

Current behavior in [preprd on cloud manager docs](https://mongodb-cloud-manager-preprd.netlify.app/docs/cloud-manager/)

### Staging Links:

[Staging link on pymongo](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/master/pymongo/seung.park/DOP-5752-follow-up/index.html) (same behavior with sidenav)

### README updates

### DIFF
--- a/src/components/Sidenav/Sidenav.tsx
+++ b/src/components/Sidenav/Sidenav.tsx
@@ -86,8 +86,8 @@ const disableScroll = (shouldDisableScroll: boolean) => css`
 
 const getTopAndHeight = (topValue: string, scrollY: string) => {
   return css`
-    top: max(min(calc(${topValue} - var(${scrollY}))), ${theme.header.actionBarMobileHeight});
-    height: calc(100vh - max(min(calc(${topValue} - var(${scrollY}))), ${theme.header.actionBarMobileHeight}));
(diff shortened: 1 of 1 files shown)


src/components/Sidenav/Sidenav.tsx
```diff
@@ -100,7 +100,7 @@ const SidenavContainer = styled.div(
     top: 0px;
     height: calc(
       100vh + ${theme.header.actionBarMobileHeight} - ${topLarge} +
-        min(calc(${topLarge} - ${theme.header.actionBarMobileHeight}), var(${scrollY}))
```
rayangler, line 103: Hm, would it work the same if we just remove `var(${scrollY})` and instead have `${scrollY}`, to avoid needing both variables + the `ts-expect-error`? Seems like the root cause could be a typo?

seungpark: good catch! done. updating staging link

rayangler (approved):

PR #1473, opened 2025-06-30
DOP-5819: Load version dropdown for other sites

biancalaube:
### Stories/Links:

DOP-5819

### Staging Links:

[preview](https://deploy-preview-1473--docs-frontend-stg.netlify.app/)

### Notes:

* before this PR the version dropdown on the unified ToC would only load for the current site that you're on
* Now, with this PR we are going to make use of the assosiated-repos/versionSelector functionality we already have to have the version dropdown on cloud-docs for atlas-cli and k8s-operator. 
* The only caveat for this way to go about the version dropdown is that we will need to add the `associated_repo` info to `snooty_dotcomprd.metadata`, which the writers already need to do in happily so it matches current functionality.
* However, the look of the two different version dropdown should look the same

### README updates


src/components/Sidenav/VersionSelector.js
```diff
@@ -58,12 +55,21 @@ const wrapperStyle = css`
...
 
 const VersionSelector = ({ versionedProject = '', tocVersionNames = [] }) => {
+  const { isUnifiedToc } = getFeatureFlags();
   const { activeVersions, availableVersions, onVersionSelect } = useContext(VersionContext);
-  const [options, setOptions] = useState(buildChoices(availableVersions[versionedProject], tocVersionNames));
+  const [options, setOptions] = useState(
+    isUnifiedToc
+      ? (availableVersions[versionedProject] || []).map(buildChoice)
+      : buildChoices(availableVersions[versionedProject], tocVersionNames)
+  );
```
zdw001, line 64: Nit: This logic is duplicated below. Also, this is a pretty complex for inline logic. Consider breaking this up into a few different variables

src/components/Sidenav/VersionSelector.js
```diff
@@ -58,12 +55,19 @@ const wrapperStyle = css`
 `;
 
 const VersionSelector = ({ versionedProject = '', tocVersionNames = [] }) => {
+  const { isUnifiedToc } = getFeatureFlags();
   const { activeVersions, availableVersions, onVersionSelect } = useContext(VersionContext);
-  const [options, setOptions] = useState(buildChoices(availableVersions[versionedProject], tocVersionNames));
+  const computeOptions = useCallback(() => {
+    return isUnifiedToc
```
zdw001, line 61: Nit: can we break this logic out into multiple lines to not have overly complex in-line logic

PR #1488, opened 2025-07-21
Dop 5916

biancalaube:
### Stories/Links:

Change the "Back Button" in show subnav to be dynamic based on the chosen L1 
Screenshot 2025-07-21 at 12 47 59 AM

DOP-5916

### Current Behavior:

### Staging Links:

Put here 

### Notes:

* found small bug if it not scrolling in accordion view, so I added that fix in this ticket 
* Also realized I still had the variable `snootyEnv` so I got rid of that in the `UnifiedSidenav.js` file.

### README updates

### DIFF
--- a/src/components/UnifiedSidenav/AccordionNav.js
+++ b/src/components/UnifiedSidenav/AccordionNav.js
@@ -96,9 +96,9 @@ export const AccordionNavPanel = ({
             <BackLink
               className={cx(backLinkStyling)}
               onClick={() => setShowDriverBackBtn(false)}
-              href={currentL1.newUrl}
+              href={currentL1?.newUrl}
             >
-              Back to {currentL1.label}
+              Back to {currentL1?.label}
             </BackLink>
           )}
           {displayedItems?.map((navItems) => (
--- a/src/components/UnifiedSidenav/AccordionNav.js
+++ b/src/components/UnifiedSidenav/AccordionNav.js
@@ -19,6 +19,7 @@ export const leftPane = () => LeafyCSS`
 
 const panelStyling = LeafyCSS`
     position: fixed;
+    overflow-y: auto;
     top: 50px;
     height: calc(100% - 120px);
     padding-top: 10px;
@@ -71,6 +72,7 @@ export const AccordionNavPanel = ({
   setShowDriverBackBtn,
   displayedItems,
   slug,
+  currentL1,
   currentL2s,
   setCurrentL1,
   setCurrentL2s,
@@ -94,9 +96,9 @@ export const AccordionNavPanel = ({
             <BackLink
               className={cx(backLinkStyling)}
               onClick={() => setShowDriverBackBtn(false)}
-              href="/docs/kafka-connector/current/whats-new/"
+              href={currentL1.newUrl}
             >
-              Back to Client Libraries
+              Back to {currentL1.label}
             </BackLink>
           )}
           {displayedItems.map((navItems) => (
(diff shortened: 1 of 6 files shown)


caesarbell (commented):
Looking at the Figma, the "Back to Client Libraries" font-weight looks lessy heavier then what is in staging.

src/components/UnifiedSidenav/AccordionNav.js
```diff
@@ -94,12 +96,12 @@ export const AccordionNavPanel = ({
             <BackLink
               className={cx(backLinkStyling)}
               onClick={() => setShowDriverBackBtn(false)}
-              href="/docs/kafka-connector/current/whats-new/"
+              href={currentL1?.newUrl}
```
caesarbell, line 99: (nit) we may want to pass in an empty string if `currentL1.newUrl` is undefined or null.

src/components/UnifiedSidenav/AccordionNav.js
```diff
@@ -94,12 +96,12 @@ export const AccordionNavPanel = ({
...
             >
-              Back to Client Libraries
+              Back to {currentL1?.label}
```
caesarbell, line 101: Do want to have a default string here, or looking at the two changes, maybe we want to render the `Back to <something>` when there is a validate `currentL1`? maybe having the whole back component rendered conditionally based on if currentL1 is not null or undefined.

src/components/UnifiedSidenav/AccordionNav.js
```diff
@@ -19,6 +19,7 @@ export const leftPane = () => LeafyCSS`
 
 const panelStyling = LeafyCSS`
     position: fixed;
+    overflow-y: auto;
```
caesarbell, line 22: Asking for learning purposes: Why are we using overflow-y?

src/components/UnifiedSidenav/DoublePannedNav.js
```diff
@@ -106,9 +106,9 @@ export const DoublePannedNav = ({
               <BackLink
                 className={cx(backLinkStyling)}
                 onClick={() => setShowDriverBackBtn(false)}
-                href="/docs/kafka-connector/current/whats-new/"
+                href={currentL1?.newUrl}
```
caesarbell, line 109: Same as my comments from AccordionNav.

src/components/UnifiedSidenav/UnifiedSidenav.js
```diff
@@ -167,9 +166,8 @@ export function UnifiedSidenav({ slug }) {
       activeVersions,
       versionsData: availableVersions,
       project,
-      snootyEnv,
     });
-  }, [unifiedTocTree, activeVersions, availableVersions, project, snootyEnv]);
+  }, [unifiedTocTree, activeVersions, availableVersions, project]);
```
caesarbell, line 170: These changes are not related to the "Back to Client Libraries" correct? Just a nice win to sneak in with this ticket?

biancalaube: Ya I realized It was left over from previous code when I used the snootyEnv to set the version but I dont actually use it anymore to set the version and I just kept passing it around in this function so I cleaned it up

src/components/UnifiedSidenav/AccordionNav.js
biancalaube: I put it in the notes section above, but I found a small bug in accordion view where its not scrolling in overflow, so added this here

PR #1501, opened 2025-08-01
DOP-5923: removing assossiated repo's in Unified toc 

biancalaube:
### Stories/Links:

DOP-5923: removing assossiated repo's in Unified toc and resolving version selector issues in UAT

Solved issues:
* no more associated repos, so no more using version selector ([bug report line 13](https://docs.google.com/spreadsheets/d/1Yw46uX0owcq8Velbv_bihO5curwlVT3PwRgs9_xFJsA/edit?gid=0#gid=0))
* There was issue if it not loading the default version on inside of the version selector ([bug report line 21](https://docs.google.com/spreadsheets/d/1Yw46uX0owcq8Velbv_bihO5curwlVT3PwRgs9_xFJsA/edit?gid=0#gid=0))
*  ([bug report line 25](https://docs.google.com/spreadsheets/d/1Yw46uX0owcq8Velbv_bihO5curwlVT3PwRgs9_xFJsA/edit?gid=0#gid=0))

### Current Behavior:

N/A

### Staging Links:

https://mongodbcom-cdn.staging.corp.mongodb.com/docs/manual/core/document/

### Notes:

* Do not need associated Repo information anymore for the unified toc to work and load different sites version selectors. 
    * We will just be passing in which `contentSite` dropdown to load
    * In `version-context.tsx` we are already fetching all the contentsite repo_branches data and we will be using that to load `availableVersionData`
    * Needed to add the `groups` array to that query call since that is needed in the VersionDropdown
* When working on this ticket I noticed that the "version" being site by the parser is what the version folder is called by the monorepo. However, that wasn't always matching the "gitBranchName" value in the `branches` array in which we query versioned data. For example, the the version `manual` is being set as `manual` in the netlify-extension but in the repo_branches data its looking for `v8.0` as the gitBranchName. So to fix this I have it looking for `urlSlug` instead since that is the naming of all the version'd folders in the monorepo. Sorry I know this is confusing to explain with the switch of branches and versions and them being intertwined so please reach out to me if you're still confused
### README updates

### DIFF
--- a/src/hooks/use-is-valid-version.ts
+++ b/src/hooks/use-is-valid-version.ts
@@ -1,7 +1,7 @@
 import { useContext, useMemo } from 'react';
 import { VersionContext } from '../context/version-context';
 import useSnootyMetadata from '../utils/use-snooty-metadata';
-import { Versions } from '../types/versions';
+import { BranchData } from '../types/data';
 import { useCurrentUrlSlug } from './use-current-url-slug';
 import { useSiteMetadata } from './use-site-metadata';
 
@@ -9,7 +9,7 @@ export const useIsValidVersion = () => {
   const { availableVersions } = useContext(VersionContext);
   const { parserBranch } = useSiteMetadata();
   const { project } = useSnootyMetadata();
-  const branches: Versions = useMemo(() => {
+  const branches: BranchData[] = useMemo(() => {
     return availableVersions[project as keyof typeof availableVersions];
   }, [project, availableVersions]);
   const currentUrlSlug = useCurrentUrlSlug(parserBranch, branches);
--- a/src/layouts/index.js
+++ b/src/layouts/index.js
@@ -126,22 +126,24 @@ const DefaultLayout = ({ children, data, pageContext: { slug, repoBranches, temp
       >
         <GlobalGrid isInPresentationMode={isInPresentationMode}>
           {!isInPresentationMode ? <Header eol={eol} template={template} /> : <div />}
-          {isUnifiedToc ? (
-            <UnifiedSidenav slug={slug} />
-          ) : sidenav && !isInPresentationMode ? (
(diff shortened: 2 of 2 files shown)


plugins/gatsby-source-snooty-prod/gatsby-node.js
```diff
@@ -481,10 +481,18 @@ exports.createSchemaCustomization = ({ actions }) => {
       versionSelectorLabel: String
     }
 
+    type Group implements Node @dontInfer {
```
caesarbell, line 484: for learning purposes why are we using `@dontInfer`?

src/components/UnifiedSidenav/UnifiedSidenav.js
```diff
@@ -89,7 +89,8 @@ const updateURLs = ({ tree, contentSite, activeVersions, versionsData, project }
     // Replace version variable with the true current version
     if (item?.url?.includes(':version')) {
       const version = (versionsData[currentProject] || []).find(
-        (version) => version.gitBranchName === activeVersions[currentProject]
+        (version) =>
+          version.gitBranchName === activeVersions[currentProject] || version.urlSlug === activeVersions[currentProject]
```
caesarbell, line 93: Asking for clarity: Can we not use alias from the repro_branches for example `urlAliases`? Maybe use an includes to see if the currentProject assuming its something like `manual` is included in the urlAliases.

caesarbell (commented):
I left some small comments.

plugins/gatsby-source-snooty-prod/gatsby-node.js
biancalaube: `@dontInfer` prevents Gatsby from automatically inferring the shape of the data. As to why we use it, I dont know I just saw that all the other types in the `gatsby-node.js` file also have it

PR #1507, opened 2025-08-11
DOP-6042: L1 css flash of wrong color in dark mode

biancalaube:
### Stories/Links:

DOP-6042

* this is hard to test locally since it doesnt happen locally but released changes on the three staging sites: `landing, cloud-docs and drivers/ecosystem`

### Current Behavior:

Bugs found in bug report
* Dark mode hover is too light on click then flashes to correct color ([bug report line 8](https://docs.google.com/spreadsheets/d/1Yw46uX0owcq8Velbv_bihO5curwlVT3PwRgs9_xFJsA/edit?usp=sharing))
     * Main thing this PR fixes is that in dark mode the L1 selected item would flash light mode every time you change the selected l2 item. 
     * Test original behavior by going from : `Development > Documents` (withing docs manual)  
     * https://mongodbcom-cdn.staging.corp.mongodb.com/docs/development/
* L1 sometimes has blue indicator on hover ([Bug report line 6](https://docs.google.com/spreadsheets/d/1Yw46uX0owcq8Velbv_bihO5curwlVT3PwRgs9_xFJsA/edit?usp=sharing))
* L1 label text sometimes resizes briefly on click ([Bug report line 7](https://docs.google.com/spreadsheets/d/1Yw46uX0owcq8Velbv_bihO5curwlVT3PwRgs9_xFJsA/edit?usp=sharing))
     * Got rid of the gatsby linking css in hopes that [these lines](https://github.com/mongodb/snooty/blob/main/src/components/Link.tsx#L42C1-L45C3) were what were causing this issue, but unsure if that is to be true.
* L1s are sometimes highlighted and sometimes not at smaller screen sizes ([bug report line 28](https://docs.google.com/spreadsheets/d/1Yw46uX0owcq8Velbv_bihO5curwlVT3PwRgs9_xFJsA/edit?usp=sharing)), should be fixed! 

### Staging Links
* Should work in cloud-docs and landing so you can test by going from `Managment > Atlas users`
https://mongodbcom-cdn.staging.corp.mongodb.com/docs/management/

### Notes:
* I accidentally created this branch off of [DOP-6087 ](https://github.com/mongodb/snooty/pull/1506)so it has some of those changes
### README updates


zdw001 (approved):

src/components/Link.tsx
```diff
@@ -176,7 +176,7 @@ const Link = ({
 
       return (
         <GatsbyLink
-          className={cx(gatsbyLinkStyling(THEME_STYLES[siteTheme]), className)}
+          className={cx(className)}
           activeClassName={activeClassName}
```
mmeigs, line 180: Remind me why you're removing the `gatsbyLinkStyling` from here and below?

src/components/UnifiedSidenav/styles/SideNavItem.tsx
```diff
@@ -29,27 +30,41 @@ export const l1ItemStyling = ({ isActive, isAccordion }: { isActive: boolean; is
...
           }
         `
       : css`
-          color: var(--sidenav-active-before-color) !important;
-          background-color: var(--sidenav-active-bg-color);
+          color: ${palette.green.dark1} !important;
+          background-color: ${palette.green.light3};
 
-          ::before {
-            color: var(--sidenav-active-before-color) !important;
```
mmeigs, line 49: Do we need this `before` color specified still? As a TOTAL nitpick, the only "flash" I see now is the `before` highlight on an L1 in dark mode is slightly different after hydration. (GREAT job, btw, I know these are annoying and only having one style left is incredible)

src/components/UnifiedSidenav/styles/SideNavItem.tsx
```diff
@@ -29,27 +30,41 @@ export const l1ItemStyling = ({ isActive, isAccordion }: { isActive: boolean; is
...
+          background-color: unset !important;
+          color: unset;
 
           ::before {
             display: none;
           }
         `
       : css`
-          color: var(--sidenav-active-before-color) !important;
-          background-color: var(--sidenav-active-bg-color);
```
mmeigs, line 46: All these css vars - if you're not going to use them and if they're not used anywhere else in Snooty, can you remove the variables themselves too? Just clean up

src/components/Link.tsx
biancalaube: I removed it cause i think these lines are what was causing the wrong color tab and change of font size https://github.com/mongodb/snooty/blob/main/src/components/Link.tsx#L42C1-L45C3

src/components/UnifiedSidenav/styles/SideNavItem.tsx
biancalaube: I added it since this is what I though was override the the blue tab sometimes, but do you not think this is effecting it ?

src/components/Link.tsx
mmeigs: 1. These styles are necessary for other Link uses. So, if you want it to not apply for UnifiedTOC links, then you'll need to find a way to make this conditional or have the CSS itself be more specific to not apply to UnifiedTOC links
2. If it was taken out, we'd want to take out the variable as well.

src/components/UnifiedSidenav/styles/SideNavItem.tsx
mmeigs: Sounds good, as long as we don't forget!!

src/components/UnifiedSidenav/styles/SideNavItem.tsx
mmeigs: I'm not sure what you mean. I think you can take this out, but just make sure to add an overwrite of the ::before color in dark mode to match the correct one!

src/components/Link.tsx
biancalaube: I think I already have it conditional for just the unified toc links i have an `if (contentSite)` for all the unified toc stuff and then I leave the Link file as it was before

src/components/UnifiedSidenav/styles/SideNavItem.tsx
mmeigs: It looks like it should be... but I can't figure out quickly why there's still a slight flash of the darker color. This is NOT a blocker. It is such a faint change.

PR #1513, opened 2025-08-12
DOP-6105: Adding icon for internal toc node symlinks

biancalaube:
### Stories/Links:

DOP-6105: Adding the icon for internal nav links, ie symlinks

** Also found small bug of clicking the `Docs Home` button at the top of the side nav opens in new tab instead of same tab so made that fix 

[slack thread ](https://mongodb.slack.com/archives/C088QVAUS2Z/p1755008760361089)

[figma link](https://www.figma.com/proto/KEtaPAkucUDw2Fdr1k6U7v/TOC-WORK?page-id=3565%3A23307&node-id=4316-36888&viewport=-10679%2C769%2C0.85&t=H6PVLKU2bSLKB3SS-1&scaling=scale-down-width&content-scaling=fixed&starting-point-node-id=4072%3A5949&show-proto-sidebar=1)

### Current Behavior:

Screenshot 2025-08-12 at 1 09 23 PM

### Staging Links:

Screenshot 2025-08-12 at 1 09 38 PM

* Staging link N/A until [DOP-6105](https://jira.mongodb.org/browse/DOP-6105) is complete
* Also made sure regular external icon svg looks as it should
    * Screenshot 2025-08-12 at 1 13 19 PM

### Notes:
Adding the icon 
### README updates

### DIFF
--- a/src/components/Link.tsx
+++ b/src/components/Link.tsx
@@ -6,6 +6,7 @@ import { useDarkMode } from '@leafygreen-ui/leafygreen-provider';
 import { palette } from '@leafygreen-ui/palette';
 // @ts-ignore
 import ArrowRightIcon from '@leafygreen-ui/icon/dist/ArrowRight';
+import Icon from '@leafygreen-ui/icon';
 import { isRelativeUrl } from '../utils/is-relative-url';
 import { joinClassNames } from '../utils/join-class-names';
 import { validateHTMAttributes } from '../utils/validate-element-attributes';
@@ -44,6 +45,15 @@ const THEME_STYLES: LinkThemeStyles = {
 export const sharedDarkModeOverwriteStyles = `
   color: var(--link-color-primary);
   font-weight: var(--link-font-weight);
+`;
+
+const symLinkStyling = css`
+  display: inline;
+  svg {
+    transform: rotate(-45deg);
+    margin-left: 8px;
+    margin-bottom: -5px;
+  }
 `;
 
 /**
@@ -85,7 +95,7 @@ const lgLinkStyling = css`
   ${sharedDarkModeOverwriteStyles}
   svg {
     margin-left: 8px;
-    margin-bottom: -6px;
+    margin-bottom: -10px;
     color: ${palette.gray.base};
   }
 
@@ -96,7 +106,7 @@ const lgLinkStyling = css`
 `;
 
 export type LinkProps = {
-  children: ReactNode;
+  children?: ReactNode;
   to?: string;
   activeClassName?: string;
   className?: string;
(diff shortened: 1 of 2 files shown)


src/components/Link.tsx
```diff
@@ -223,14 +249,14 @@ const Link = ({
   const strippedUrl = to?.replace(/(^https:\/\/)|(www\.)/g, '');
   const isMDBLink = strippedUrl.includes('mongodb.com');
   const showExtIcon = showExternalIcon ?? (!anchor && !isMDBLink && !hideExternalIconProp);
-  const target = !showExtIcon ? '_self' : undefined;
+  const target = !showExtIcon || !openInNewTab ? '_self' : 'blank';
```
mmeigs, line 252: I'm worried with this conditional - because any Link that has `openInNewTab` as `true` but has `showExtIcon` as `false`, this means that it will NOT open in a new tab. But it needs to.

src/components/Link.tsx
```diff
@@ -85,7 +95,7 @@ const lgLinkStyling = css`
   ${sharedDarkModeOverwriteStyles}
   svg {
     margin-left: 8px;
-    margin-bottom: -6px;
+    margin-bottom: -10px;
```
mmeigs, line 98: This new margin seems to apply to every LG Link in docs - not just the UnifiedTOC. Is that the desired effect?

mmeigs (commented):
Couple comments below! 

Also, this might not be the place to put this question, but I wonder if we should actually make two different components. Link for most docs content and UnifiedTOCLink. The logic is pretty widely split. I don't know if that's the right decision, but interested in floating it rather than keep allowing this component to diverge so much. Idk, up to you! Not a blocker.

mmeigs:
Oh, random: my figma link isn't allowing me to inspect: but that symlink icon in your screenshot feeels like it's a little too low and maybe even a little too big vs the text?

src/components/Link.tsx
biancalaube: good catch, made it its own css styling

biancalaube:
I think its a good point that the link should be broken up into two components, i made this ticket [DOP-6108](https://jira.mongodb.org/browse/DOP-6108) to do that since i want the changes to go live for the showcase this week but that is def something i want to work on. I also updated the css so it should be matching the figma link better. the updated version looks like 
Screenshot 2025-08-12 at 4 06 19 PM

src/components/Link.tsx
```diff
@@ -249,7 +255,7 @@ const Link = ({
   const strippedUrl = to?.replace(/(^https:\/\/)|(www\.)/g, '');
   const isMDBLink = strippedUrl.includes('mongodb.com');
   const showExtIcon = showExternalIcon ?? (!anchor && !isMDBLink && !hideExternalIconProp);
-  const target = !showExtIcon || !openInNewTab ? '_self' : 'blank';
+  const target = !showExtIcon ? '_self' : undefined;
```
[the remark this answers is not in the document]
biancalaube: Thanks matt for the info! went with option 1

PR #1516, opened 2025-08-12
DOP-6046: fix styling for HL

seungpark:
### Stories/Links:

DOP-6046

### Current Behavior:

[See this page for horizontal list item](https://www.mongodb.com/docs/atlas/unsupported-commands/)

### Staging Links:

[Staged above individual page via S3](https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/preprd/cloud-docs/seung.park/DOP-6046/unsupported-commands/index.html)

### Notes:
- Verifying with Design team that this is the direction (vs scrollable table)
- Above has been verified, and we will tackle text wrapping with monospace as a separate mobile issue

### README updates

### DIFF
--- a/src/components/HorizontalList.tsx
+++ b/src/components/HorizontalList.tsx
@@ -55,7 +55,7 @@ const HorizontalList = ({
       <tbody>
         <tr>
           {columnArray.map((col, colIndex) => (
-            <td className={cx(tableDataStyling, 'hlist-cell')} key={colIndex}>
+            <td className={cx(tableDataStyling)} key={colIndex}>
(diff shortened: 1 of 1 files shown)


mmeigs (approved):
Looks great mainly, and I see your comment about dealing with mobile later... but I'm wondering if we can just throw in an easy mobile breakpoint so that this isn't a regression on mobile. Not a blocker if design has ok-ed. But maybe at like 500-600px we make the table `display: block, overflow-x: scroll`? Just so its readable. But up to design!

PR #1533, opened 2025-08-27
Language-fix for unified toc

biancalaube:
### Stories/Links:

Today when testing the release for Unified ToC a Bug came up that I havent previously accounted for. For unified toc we use the url and the toc.ts data to find out which page is active to show on the unified ToC. However, on translated pages the following four prefixes could be added to the url `'zh-cn', 'ja-jp', 'ko-kr', 'pt-br'`. This cause for the Unified ToC to never persist showing an active item. Changed the comparison function so it will remove the language prefix for comparison. Made sure in the link function still links the language 

### Current Behavior:

https://mongodbcom-cdn.staging.corp.mongodb.com/ko-kr/docs/atlas/

### Staging Links:

Deployed the Landing site with this fix: https://mongodbcom-cdn.staging.corp.mongodb.com/ko-kr/docs/tools-and-connectors/

Deployed mongodb intelliJ with this fix: https://mongodbcom-cdn.staging.corp.mongodb.com/ko-kr/docs/mongodb-intellij/

### README updates

### DIFF
--- a/src/components/Link.tsx
+++ b/src/components/Link.tsx
@@ -12,10 +12,6 @@ import { joinClassNames } from '../utils/join-class-names';
 import { validateHTMAttributes } from '../utils/validate-element-attributes';
 import { useSiteMetadata } from '../hooks/use-site-metadata';
 import { assertLeadingAndTrailingSlash } from '../utils/assert-trailing-and-leading-slash';
-// import { removeTrailingSlash } from '../utils/remove-trailing-slash';
-// import { assertLeadingSlash } from '../utils/assert-leading-slash';
-// import { removeLanguage } from './UnifiedSidenav/UnifiedSidenav';
-
 /*
  * Note: This component is not suitable for internal page navigation:
  * https://www.gatsbyjs.org/docs/gatsby-link/#recommendations-for-programmatic-in-app-navigation
@@ -207,7 +203,7 @@ const Link = ({
       // Get rid of the contenteSite in link for internal links
       // Get rid of the path contentSite in link for internal links
       const editedTo = assertLeadingAndTrailingSlash(to.replace(pathPrefix, ''));
-      console.log('editTO', editedTo, to);
+
       return (
         <GatsbyLink
           className={cx(className)}
@@ -223,15 +219,6 @@ const Link = ({
       );
     }
 
-    // if pathprefix contains language replace newurl's path prefix with the pathprefix
-    // console.log('to', to);
-    // const hasLang = langArray.some((lang) => pathPrefix?.includes(lang));
-    // if (hasLang) {
-    //   to = to.replace(removeLanguage(pathPrefix), '');
-    //   to = removeTrailingSlash(pathPrefix) + assertLeadingSlash(to);
-    // }
-    // console.log('to after', to);
-    // On the Unified SideNav but linking to a different content site
     return (
       <a className={cx(className)} href={to}>
         {children}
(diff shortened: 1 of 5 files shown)


src/components/Link.tsx
```diff
@@ -220,6 +221,13 @@ const Link = ({
       );
     }
 
+    // if pathprefix contains langauge replce newurl's path prefix with the pathprefix
+    const hasLang = langArray.some((lang) => pathPrefix?.includes(lang));
```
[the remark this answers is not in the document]
biancalaube: The use case here is that the `to` is from the toc.ts file which would be. `docs/mongosync/current/topologies/` but if the path prefix is: `ko-kr/docs/mongosync/current` then I want to remove the path prefix in the `to` and replace it with `pathPrefix` which has which language its suppose to be. So I only wanted to do those steps _if_ its suppose to be translated. But I could also just do it in every scenario and it would just replace itself

src/components/UnifiedSidenav/UnifiedSidenav.js
```diff
@@ -165,11 +175,13 @@ export function UnifiedSidenav({ slug }) {
...
   const { pathname, hash } = useLocation();
   const tempSlug = isBrowser ? removeLeadingSlash(removeTrailingSlash(window.location.pathname)) : slug;
-  slug = tempSlug?.startsWith('docs/')
-    ? tempSlug
-    : tempSlug === '/'
-    ? pathPrefix + tempSlug
-    : `${pathPrefix}/${tempSlug}/`;
+  const hasLang = langArray.some((lang) => tempSlug?.includes(lang));
+  slug =
+    tempSlug?.startsWith('docs/') || hasLang
```
caesarbell, line 180: Do we need the `|| hasLang`? 

Couldn't we just run the `removeLanguage` on slug since in that method you are already looking to see if the prefix has one of the language options included in the path?

src/components/Link.tsx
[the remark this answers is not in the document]
caesarbell: I second what @cbush mentioned. In the `removeLanguage` method you check to see if the path includes a language prefix or not and return just the slug if it doesn't. So having the additional logic to create the value for `hasLang` might not be needed.

src/utils/assert-leading-slash.js
```diff
@@ -1,8 +1,6 @@
 const assertLeadingSlash = (url) => {
-  if (url && url.match(/^\//)) {
-    return url;
-  }
-  return `/${url}`;
+  if (!url) return '/';
```
caesarbell, line 2: Just need clarity on why this needed to change.

caesarbell (commented):
Left some comments and a question.

src/utils/assert-leading-slash.js
biancalaube: I updated it since if there was a double prefix infront of the slug when using this function it wouldnt fix it. For example if I removed the language it would turn into `//docs/tools-and-connectors`

src/components/UnifiedSidenav/UnifiedSidenav.js
biancalaube: change to just use hasLang

src/utils/assert-leading-slash.js
caesarbell: @biancalaube gotcha, could we have also removed the `/` along with the language prefix to not have to modify this function? 

i.e. `/pt-br`

PR #1537, opened 2025-08-29
Dop 6157 fetch docsets

mmeigs:
### Stories/Links:

DOP-NNNN

### Current Behavior:

### Staging Links:

### README updates

### DIFF
--- a/src/utils/docsets.ts
+++ b/src/utils/docsets.ts
@@ -1,0 +1,12 @@
+import { Docset } from '../types/data';
+
+// Fetches for Docsets (made separate for usage and creating mocks for testing components)
+export const fetchDocset = async (dbName: string, project: string): Promise<Docset> => {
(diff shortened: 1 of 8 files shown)


src/utils/docsets.ts
```diff
@@ -0,0 +1,12 @@
+import { Docset } from '../types/data';
+
+// Fetches for Docsets (made separate for usage and creating mocks for testing components)
+export const fetchDocset = async (dbName: string, project: string): Promise<Docset> => {
+  const res = await fetch(`${process.env.GATSBY_NEXT_API_BASE_URL}/docsets/${project}?dbName=${dbName}`);
```
caesarbell, line 5: Super nit: Would we want to separate this fetch call into its own util that we can reuse in the different files? I feel that if we ever need to make a change, we would have to change in all these places.

caesarbell (approved):
I left a nit, but overall LGTM.

caesarbell:
Snooty has a functional CI now. Any new pushes should reflect that.

PR #1541, opened 2025-09-03
DOP-5821 handles styling issue along with Safari blocking clipboard behavior

caesarbell:
### Stories/Links:

DOP-5821

### Current Behavior:

[Query Atlas with Natural Language Using LangChain and LangGraph](https://www.mongodb.com/docs/atlas/ai-integrations/langchain/natural-language-to-mql/)

### Staging Links:

### Notes:

This handles styling issues in the copy page button (mainly found when changing the language) - a new line was created and you would get this odd stacking of the words. 

Also, on Safari, the copy to clipboard behavior was being rejected with the following message; (It works find in Chrome).

```bash
NotAllowedError: The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.
writeText
(anonymous function) — index.tsx:49
```

Which led me to believe in Safari, the user gesture is being lost by the time the call to writeText was made (due to the awaiting of a network fetch first).

Also, it was probably better for performance to prefetch the markdown.

### README updates

### DIFF
--- a/src/components/Widgets/MarkdownWidget/index.tsx
+++ b/src/components/Widgets/MarkdownWidget/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import { useLocation } from '@gatsbyjs/reach-router';
 import { SplitButton } from '@leafygreen-ui/split-button';
 import { Size } from '@leafygreen-ui/button';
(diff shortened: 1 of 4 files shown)


caesarbell:
> It seems to fail copying in Safari in the preview link

Sorry about that, I should have left a message that I didn't push to preprd yet. I wanted to wait until the release week was over and then ended up not updating. Doing that now though.

PR #1549, opened 2025-09-10
DOP-6158: Migrate `fetchDocument`, `fetchDocuments`

mmeigs:
### Stories/Links:

DOP-6158

### Current Behavior:

### Staging Links:

### README updates


plugins/utils/documents.js
```diff
@@ -0,0 +1,31 @@
+const fetchDocument = async (dbName, collectionName, query, projectionOptions, sortOptions) => {
```
seungpark, line 1: is this required? seems like only the sorted one is used for initialization

src/utils/setup/save-asset-files.js
```diff
@@ -33,7 +36,15 @@ const saveAssetFiles = async (assets, db) => {
 
   for (const [id, filenames] of assets) {
     if (filenames) {
-      const buffer = await db.getAsset(id);
+      let buffer;
+      if (db instanceof RealmDocumentDatabaseClass) {
+        const assetQuery = { _id: id };
+        const assetDataDocuments = await fetchDocumentSorted(siteMetadata.database, ASSETS_COLLECTION, assetQuery);
```
seungpark, line 42: this was previously `fetchDocuments(ASSETS_COLLECTION, assetQuery)`  but sorted does not hurt either - also [minor nit] if we leave this within realm db (to have parity with manifest db and realm db as previous implementation) and call the next api from there and assume consumers do not know a difference, but no strong preference

ASteinheiser (approved):
lgtm

src/utils/setup/save-asset-files.js
mmeigs: Okay, what's craaaazy is that it was fetchDocuments... that called fetchDocumentSorted - not joking 😨 

Hmmm - okay, I understand the logic to keeping it in there. I guess we would just not use init() or maybe it would be an empty method?

seungpark: was thinking [this might](https://github.com/mongodb/snooty/pull/1549/files#diff-5ce42b93f0104da065ed87929695d3f9275954c9fa4c0da2665e26fe35b4711aL79) call the [next API](https://github.com/mongodb/snooty/pull/1549/files#diff-cc6402534f1665b58774875e94bed507e5e7ed7ba2683b3324f146fd1667e3e1R26)

mmeigs: What's the difference betweeen [this](https://github.com/mongodb/snooty/pull/1549#:~:text=was%20thinking-,this%20might,-call%20the%20next) and [this](https://github.com/mongodb/snooty/pull/1549/files#:~:text=1%3E%3B-,export,-const%20fetchDocument%20%3D)?

PR #1554, opened 2025-09-19
DOP-5322 🆕  introduces the language callout user flow

caesarbell:
### Stories/Links:

DOP-5322

### Current Behavior:

[Log in with Atlas](https://www.mongodb.com/docs/relational-migrator/getting-started/atlas-log-in/)

### Staging Links:

### Notes:

This PR introduces the language callout flow, which is a banner that will appear on pages within the targets array of the banner directive from the [snooty.toml](https://github.com/10gen/docs-mongodb-internal/blob/9ef44c815a62a0b1e29ef54ad5af8c1213d56e6f/content/atlas/snooty.toml#L737-L738). These banners carry the locale option that the writers can use to determine which languages they want to slowly introduce users too, and when a user hits that page that match that locale, the banner will show for that page with a list of pages that are supported in this language. 

The reason for this flow is to slowly introduce users to new languages. The localization team wanted a way to slowly introduce new pages, and this language callout flow was what they decided on. 

### README updates

### DIFF
--- a/src/components/Footer.js
+++ b/src/components/Footer.js
@@ -1,9 +1,11 @@
-import React from 'react';
+import React, { useContext } from 'react';
 import { UnifiedFooter } from '@mdb/consistent-nav';
-import { getAvailableLanguages, getCurrLocale, onSelectLocale } from '../utils/locale';
+import { getCurrLocale, onSelectLocale } from '../utils/locale';
+import { LocaleContext } from '../context/locale';
 
 const Footer = () => {
-  const enabledLocales = getAvailableLanguages().map((language) => language.localeCode);
+  const { enabledLocalesData } = useContext(LocaleContext);
+  const enabledLocales = enabledLocalesData.map((language) => language.localeCode);
   return <UnifiedFooter onSelectLocale={onSelectLocale} locale={getCurrLocale()} enabledLocales={enabledLocales} />;
 };
 
--- a/src/components/Header/index.tsx
+++ b/src/components/Header/index.tsx
@@ -4,8 +4,9 @@ import { css, cx } from '@leafygreen-ui/emotion';
 import { UnifiedNav } from '@mdb/consistent-nav';
 import SiteBanner from '../Banner/SiteBanner';
 import { theme } from '../../theme/docsTheme';
-import { getAvailableLanguages, getCurrLocale, onSelectLocale } from '../../utils/locale';
+import { getCurrLocale, onSelectLocale } from '../../utils/locale';
 import { isOfflineDocsBuild } from '../../utils/is-offline-docs-build';
+import { LocaleContext } from '../../context/locale';
 import { HeaderContext } from './header-context';
 
 interface StyledHeaderProps {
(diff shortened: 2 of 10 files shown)


src/context/locale.tsx
```diff
@@ -0,0 +1,44 @@
...
+      setEnableLocalesData([...data, betaLocale]);
+    } else {
+      setEnableLocalesData(data);
+    }
+  }, [locale, betaLocale]);
+
+  return <LocaleContext.Provider value={{ enabledLocalesData }}>{children}</LocaleContext.Provider>;
+};
+
+export { LocaleContext, LocaleProvider };
```
ASteinheiser, line 44: Not blocking, but just like you made the `LocaleProvider` I am a huge fan of exporting a `useWhatever` hook, that basically just calls `useContext` and throws if there's no provider:
```
export const useLocale = () => {
    const context = useContext(LocaleContext);

    if (!context) throw new Error('useLocale must be used within a LocaleProvider');

    return context;
}
```

ASteinheiser: So then, the only exports are the `LocaleProvider` and the `useLocale`, neither require extra config/imports to work.

src/components/Header/index.tsx
```diff
@@ -76,7 +79,7 @@ const Header = ({ eol }: HeaderProps) => {
                 showLanguageSelector={true}
                 onSelectLocale={onSelectLocale}
                 // @ts-ignore
```
ASteinheiser, line 81: Q: I noticed the variable updated, do we still need these ignore comments now?

src/components/Header/index.tsx
```diff
@@ -44,7 +45,9 @@ type HeaderProps = {
 
 const Header = ({ eol }: HeaderProps) => {
   const unifiedNavProperty = 'DOCS';
-  const enabledLocales = getAvailableLanguages().map((language) => language.localeCode);
+  const { enabledLocalesData } = useContext(LocaleContext);
+  const enabledLocales = enabledLocalesData.map((language) => language.localeCode);
+  const locale = getCurrLocale();
```
ASteinheiser, line 50: Just a gut feeling, but since this is based off the page, could this value just come from the Locale context and get calculated via an effect that runs on url path change?

ASteinheiser: Same with the `enabledLocales` above. Would it make sense to encapsulate this logic in the Locale context/hook/provider

src/components/Footer.js
```diff
@@ -1,9 +1,11 @@
...
+import React, { useContext } from 'react';
 import { UnifiedFooter } from '@mdb/consistent-nav';
-import { getAvailableLanguages, getCurrLocale, onSelectLocale } from '../utils/locale';
+import { getCurrLocale, onSelectLocale } from '../utils/locale';
+import { LocaleContext } from '../context/locale';
 
 const Footer = () => {
-  const enabledLocales = getAvailableLanguages().map((language) => language.localeCode);
+  const { enabledLocalesData } = useContext(LocaleContext);
+  const enabledLocales = enabledLocalesData.map((language) => language.localeCode);
```
ASteinheiser, line 8: Great case for stuffing this logic away in the hook/provider, since it's copied here and one other place already

ASteinheiser: For example:
```
const { enabledLocales } = useLocale();
```

ASteinheiser (approved):
Left some nits and hook design comments

src/components/Header/index.tsx
caesarbell: @ASteinheiser, for the unified nav and footer, we still need the ignore comment since the locale being passed in is of type `AvailableLocaleType` and not all locales are included to match the Locale type. Primarily 'it-it', which I may need to ask around and see what that value is about.

PR #1604, opened 2026-03-12
Add LocalizedLinkProvider and upgrade consistent-nav package

mmeigs:
### Stories/Links:

DOP-6631

### Current Behavior:

https://www.mongodb.com/docs/atlas/

### Staging Links:

https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/main/cloud-docs/matt.meigs/DOP-6631-nav/index.html

### Notes:

Updated `consistent-nav` package to latest.
This required adding their new `LocalizedLinkProvider`. We will want to test this in preprd before release. This is a breaking change that they added without any guidance. I looked through their implementation and it seems simple, but we want to make sure it works with smartling and our system.

### README updates

### DIFF
--- a/src/components/Header/index.tsx
+++ b/src/components/Header/index.tsx
@@ -1,12 +1,13 @@
 import React, { useContext } from 'react';
 import styled from '@emotion/styled';
 import { css, cx } from '@leafygreen-ui/emotion';
-import { UnifiedNav } from '@mdb/consistent-nav';
+import { LocalizedLinkProvider, UnifiedNav } from '@mdb/consistent-nav';
 import SiteBanner from '../Banner/SiteBanner';
 import { theme } from '../../theme/docsTheme';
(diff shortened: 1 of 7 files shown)


src/components/Header/index.tsx
```diff
@@ -49,14 +50,19 @@ const Header = ({ eol }: HeaderProps) => {
   const locale = getCurrLocale();
   const { hasBanner } = useContext(HeaderContext);
 
+  const pageUrl = (() => {
+    if (isBrowser) return stripLocale(window.location.pathname);
+    else return '';
+  })();
```
ASteinheiser: nit: since this is used in two spots, and imports a few utils to make this work, how do you feel about creating a util for this?

ASteinheiser (approved):
left a non-blocking nit

PR #1605, opened 2026-04-02
DOP-6722: Remove /docs from origin string for LocalizedLinkProvider

mmeigs:
### Ticket

DOP-6722

### Notes:

New `LocalizedLinkProvider` is incorrectly changing paths to `/es/docs/es/docs/...`, etc.

### README updates


src/components/Footer.js
```diff
@@ -8,8 +8,10 @@ const Footer = () => {
   const { enabledLocales } = useLocale();
   const pageUrl = getPageUrl();
 
+  console.log('pageUrl', pageUrl);
```
biancalaube: NiT: Remove this log

biancalaube (approved):
LGTM 🚀  quick question though. How did you figure out this was the fix? I'm assuming it's because that `pageUrl` starts with `/docs`

tests/unit/Presentation.test.js
```diff
@@ -25,14 +25,19 @@ useStaticQuery.mockImplementation(() => ({
...
   beforeAll(() => {
     jest.spyOn(document, 'querySelector');
   });
 
   it('renders the necessary elements', async () => {
     await act(async () => {
-      mockLocation(null);
+      mockLocation(
+        null,
+        '/docs/drivers/community-supported-drivers',
```
biancalaube, line 54: NiT: should we add a test that has a url that just starts with `docs/` instead of the starting with the slash just to make sure we account for this scenario

